Issue #13, opened 2020-08-16
Hold all PRs in stable repo for prometheus* charts


scottrigby:
Existing prometheus PRs in stable are on hold for now (see linked PRs on https://github.com/prometheus-community/community/issues/28), but we don't have any automated way to hold new PRs that may open between now and when that issue is resolved (this new helm repo ready to be transferred to prometheus-community, stable prometheus charts deprecated, new helm repo with charts listed on the hubs). 

**Maintainers:** Please help with this if you see new PRs for prometheus charts open in the meantime.

Let's keep this open to track the effort, until the above is complete.

- [x] Existing prometheus PRs in stable
- [ ] New prometheus PRs in stable

scottrigby:
I did another round of this last night

scottrigby:
Myself and a few others have been doing this all along.

Now that #28 is finished, and we've been keeping up with this issue along the way, we can close this now and move onto #29

Issue #12, opened 2020-08-16
Invite current stable prometheus chart maintainers as collaborators with write access


scottrigby:
- [x] 1. We can start by adding a single GitHub [CODEOWNERS](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) file to reproduce what the prow OWNERS files contain now. This can help with enabling [required reviews on pull requests](https://docs.github.com/en/github/administering-a-repository/enabling-required-reviews-for-pull-requests)
- [x] 2. Then all collaborators can be invited with [write access](https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization), without changing existing functionality from `stable`
~- [ ] 3. cleanup: remove the prow `OWNERS` files~ **Edit:** will be done as part of #14 so the initial helm index will be clean

## 📊 Maintainers poll
Do we want to keep this level of granularity, or would it be better to give all collaborators write access to all charts in this repo? Let's discuss in comments below…

asherf:
👋

scottrigby:
Invited all current maintainers as collaborators to this repo, and also pinged everyone in Slack ✅

scottrigby:
Will close this issue for now since tasks are done ✅  but we can still comment here in the meantime if there is overwhelming concern about implementation and re-open if needed ⏳

Xtigyro:
I believe we should keep the `OWNERS` file and the granularity because different people are involved in developing the different charts though closely related. 

I do think though that we should all have owners permissions regarding the repo. This shouldn't take from the actual permission to push commits/approve PRs for the different charts.

scottrigby:
@Xtigyro after slack chat, I understand what you mean. OK so yes, keeping `CODEOWNERS` file as-is, but can not change access level until after repo transfer. Also open questions about effects of combining that access level with `CODEOWNERS` file. Added all that to a new issue, to be addressed after the transfer: https://github.com/scottrigby/prometheus-helm-charts/issues/20 Thanks!

PR #14, opened 2020-08-16
Prep initial charts indexing

### LINKED ISSUE #7: Update each chart's README to add new helm repo in installation instructions
Ideally this should happen before indexing/releasing first version of each chart, otherwise when aggregated in the hubs the README will refer `stable` instead of the new `prometheus-community` repo.

### LINKED ISSUE #8: Deprecate prometheus-operator chart before helm repo index, so that it won't be listed in the hubs
This chart will be renamed, but first must be deprecated before the prometheus-community/helm-charts repo is indexed, so that it won't be listed in the hubs. See [this prometheus-community issue](https://github.com/prometheus-community/community/issues/28#issuecomment-670406329) for reasoning and next steps.

scottrigby:
Fixes #7, Fixes #8

- [x] Cherry-pick commits from the one PR that was merged since the code was copied https://github.com/helm/charts/pull/23506 (see #13 for context)
- [x] Deprecate charts/prometheus-operator (see #8 and #9 for context)
- [x] Remove prow `OWNERS` files (see #12 for context)

Charts that need to be prepared for initial helm index
- [x] charts/prometheus
- [x] charts/prometheus-adapter
- [x] charts/prometheus-blackbox-exporter
- [x] charts/prometheus-cloudwatch-exporter
- [x] charts/prometheus-consul-exporter
- [x] charts/prometheus-couchdb-exporter
- [x] charts/prometheus-mongodb-exporter
- [x] charts/prometheus-mysql-exporter
- [x] charts/prometheus-nats-exporter
- [x] charts/prometheus-node-exporter
- [x] charts/prometheus-postgres-exporter
- [x] charts/prometheus-pushgateway
- [x] charts/prometheus-rabbitmq-exporter
- [x] charts/prometheus-redis-exporter
- [x] charts/prometheus-snmp-exporter
- [x] charts/prometheus-to-sd

**Suggestion:** per the steps in the plan at https://github.com/prometheus-community/community/issues/28#issuecomment-670230344 (which has been approved for the prometheus-community), for this PR review and comments let's discuss only what should happen in the intial step. To be clear, this PR is the first step, before "Step 2. Collaborate with prometheus devs to refactor/rename and simplify charts as appropriate." I outlined what I think we need in #7, but please let me know if I'm missing anything.

## 📊  Maintainers poll
How do you all feel about the new `charts/prometheus` README update in b324685? I Needed to update references to stable repo, but took the opportunity to reorganize, fix and simplify. Is this a good pattern for the other chart READMEs? Note major differences are:
- Added Helm 2 and Helm 3 commands, with links to Helm 3 command reference links (from there, users can select Helm 2 from the version dropdown if they wish)
- Removed the options table that's duplicated from (often out of sync with) `values.yaml` in favor of `helm inspect` (Helm 2) and `helm show values` (Helm 3)

**Results:** new template has been approved by all who responded

### DIFF
--- a/charts/prometheus-couchdb-exporter/README.md
+++ b/charts/prometheus-couchdb-exporter/README.md
@@ -12,7 +12,6 @@ This chart bootstraps a [CouchDB Exporter](https://github.com/gesellix/couchdb-p
 
 ```console
 helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
-helm repo add stable https://kubernetes-charts.storage.googleapis.com/
 helm repo update
 ```
 
--- a/charts/prometheus-couchdb-exporter/README.md
+++ b/charts/prometheus-couchdb-exporter/README.md
@@ -1,80 +1,68 @@
 # prometheus-couchdb-exporter
 
-[couchdb-prometheus-exporter](https://github.com/gesellix/couchdb-prometheus-exporter) is a Prometheus exporter for CouchDB metrics.
+A Prometheus exporter for [CouchDB](https://couchdb.apache.org/) metrics.
 
-## TL;DR;
-
-```bash
-$ helm install stable/prometheus-couchdb-exporter
-```
-
-## Introduction
-
-This chart bootstraps a [couchdb-exporter](https://github.com/gesellix/couchdb-prometheus-exporter) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+This chart bootstraps a [CouchDB Exporter](https://github.com/gesellix/couchdb-prometheus-exporter) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
 
 ## Prerequisites
 
 - Kubernetes 1.8+ with Beta APIs enabled
 
-## Installing the Chart
+## Get Repo Info
 
-To install the chart with the release name `my-release`:
-
-```bash
-$ helm install --name my-release stable/prometheus-couchdb-exporter
+```console
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+helm repo update
 ```
 
-The command deploys prometheus-couchdb-exporter on the Kubernetes cluster in the default configuration.
+_See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation._
 
-## Uninstalling the Chart
+## Install Chart
 
-To uninstall/delete the `my-release` deployment:
+```console
+# Helm 3
+$ helm install [RELEASE_NAME] prometheus-community/prometheus-couchdb-exporter
 
-```bash
-$ helm delete my-release
+# Helm 2
+$ helm install --name [RELEASE_NAME] prometheus-community/prometheus-couchdb-exporter
 ```
 
-The command removes all the Kubernetes components associated with the chart and deletes the release.
+_See [configuration](#configuration) below._
+
+_See [helm install](https://helm.sh/docs/helm/helm_install/) for command documentation._
+
+## Uninstall Chart
+
+```console
+# Helm 3
+$ helm uninstall [RELEASE_NAME]
+
+# Helm 2
+# helm delete --purge [RELEASE_NAME]
+```
+
+This removes all the Kubernetes components associated with the chart and deletes the release.
+
+_See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall/) for command documentation._
+
+## Upgrading Chart
+
+```console
+# Helm 3 or 2
+$ helm upgrade [RELEASE_NAME] [CHART] --install
+```
+
+_See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documentation._
 
 ## Configuration
 
-The following table lists the configurable parameters and their default values.
+See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing). To see all configurable options with detailed comments, visit the chart's [values.yaml](./values.yaml), or run these configuration commands:
 
-| Parameter              | Description                                         | Default                                 |
-| ---------------------- | --------------------------------------------------- | --------------------------------------- |
-| `replicaCount`         | desired number of prometheus-couchdb-exporter pods  | `1`                                     |
-| `image.repository`     | prometheus-couchdb-exporter image repository        | `gesellix/couchdb-prometheus-exporter`  |
-| `image.tag`            | prometheus-couchdb-exporter image tag               | `16`                                    |
-| `image.pullPolicy`     | image pull policy                                   | `IfNotPresent`                          |
-| `service.type`         | desired service type                                | `ClusterIP`                             |
-| `service.port`         | service external port                               | `9984`                                  |
-| `ingress.enabled`      | enable ingress controller resource                  | `false`                                 |
-| `ingress.annotations`  | annotations for the host's ingress records          | `false`                                 |
-| `ingress.path`         | path for the ingress route                          | `/`                                     |
-| `ingress.hosts`        | list of host address for ingress creation           |                                         |
-| `ingress.tls`          | utilize TLS backend in ingress                      |                                         |
-| `resources`            | cpu/memory resource requests/limits                 | {}                                      |
-| `nodeSelector`         | node labels for pod assignment                      | {}                                      |
-| `tolerations`          | tolerations for pod assignment                      | {}                                      |
-| `affinity`             | affinity settings for proxy pod assignments         | {}                                      |
-| `couchdb.uri`          | address of the couchdb                              | `http://couchdb.default.svc:5984`       |
-| `couchdb.databases`    | comma separated databases to monitor                | `_all_dbs`                              |
-| `couchdb.username`     | username for couchdb                                |                                         |
-| `couchdb.password`     | password for couchdb                                |                                         |
+```console
+# Helm 2
+$ helm inspect values prometheus-community/prometheus-couchdb-exporter
 
-
-For more information please refer to the [couchdb-prometheus-exporter]https://github.com/gesellix/couchdb-prometheus-exporter) documentation.
-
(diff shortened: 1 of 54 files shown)


Miouge1:
LGTM, I think it makes sense to drop the table of values from the README.md, it's often hard to keep it in sync with reality.

scottrigby:
So far it sounds like we may be able to compromise with `helm upgrade [RELEASE_NAME] [CHART] --install`. Welcome other opinions here as well (though not trying to bike-shed). Looking for feedback on the readme reorganization overall as a pattern for the rest of the chart updates. Thanks!

monotek:
I'm ok with using the upgrade command. Maybe the long version (--install instead of -i) of the arguments should be used to be clear.

Rest of the readme looks good to me and could be used as a template for the other charts.

scottrigby:
@monotek @steven-sheehy & gang, I put the Helm 3 commands above Helm 2. And instead of using `helm upgrade` for installation instructions, I took suggestions above for making `helm install` use an explicit release name, then added `helm upgrade` commands in a similar way to the "Upgrading chart" section so there's no confusion. I also reorganized the upgrading MAJOR version headers slightly to be more clear to end users, depending on the version they're upgrading to. This has worked out well for other community charts. The pattern is like this:

```markdown
## Upgrading Chart
(helm commands)

## To v6.0
(instructions)

## To v5.0
(instructions)

## To v4.0
(instructions)

## From `foo` legacy chart located <HERE> (if any)
(instructions)
```

Does this look reasonable? I can apply this same pattern to the other charts.

monotek (approved):

charts/prometheus/README.md
```diff
@@ -2,57 +2,82 @@
...
+To disable the dependency during installation, set `kubeStateMetrics.enabled` to `false`.
+
+_See [helm dependency](https://helm.sh/docs/helm/helm_dependency/) for command documentation._
+
+## Uninstall Chart
 
 ```console
-helm delete my-release
+# Helm 3
+$ helm uninstall [RELEASE_NAME]
```
Xtigyro, line 49: Why not just `helm delete` for both Helm versions? In Helm v3 `helm del` actually purges the chart by default, right?

charts/prometheus/README.md
```diff
@@ -2,57 +2,82 @@
...
 ## Prerequisites
 
 - Kubernetes 1.3+ with Beta APIs enabled
 
-## Installing the Chart
+## Get Repo Info
 
-To install the chart with the release name `my-release`:
+```console
+$ helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
```
Xtigyro, line 14: the `$` shouldn't be there if there isn't a real output after exec-ing the cmd.

Xtigyro: The same for all the `$` additions.

Xtigyro (changes requested):

charts/prometheus/README.md
scottrigby: For Helm 2, `--purge` is recommended. In Helm 3 `delete` is an alias for `uninstall`, but has no `--purge` flag. Therefore it can't be identical.

charts/prometheus/README.md
scottrigby: I just added for consistency between all console blocks, so it's unambiguous to end users which parts of codeblocks are commands, and which refer to output or otherwise. Does that makes sense?

Xtigyro: Yeah - but if there's no output, there should be no `$` according to the MD linter.

charts/prometheus/README.md
scottrigby: Follow-up from chat: `delete` was added as an alias to ease some CI workflows, but it's not encouraged for Helm 3 generally, which is why it's not in the docs. The Helm `uninstall` concept itself is different now with Helm 3 architectural changes than default Helm 2 `delete` , but in the end maps to Helm 2's helm `delete --purge`.

Mainly we are trying to help lead users to Helm 3, since there is only now 3 months until Helm 2 is unsupported.

To be clear, `delete` can be used with Helm 3, but not encouraged due to user confusion between the different Helm versions, when `--purge` is needed or not for a full uninstall, leading to different results depending on version. It's a lot to explain fully so IMO easiest to document here as different commands for Helm 3 and 2, then just remove Helm 2 references after Helm 2 support ends.

You already said in Slack that makes sense, but just so others who may be interested can catch up as well 😄

charts/prometheus/README.md
scottrigby: That makes sense. My feeling is, the only real reason to have multiline shell commands with no $ prefix is for multi-line copy/paste. This is how I interpret the linting rule rationale you posted above as well. I agree in this case it's a good idea to copy/paste as a block - changing! Thanks for the detailed review.

scottrigby:
I've now resolved all prometheus chart maintainer feedback, so will apply this pattern to the other charts so we can get this merged and expedite the repo transfer ⏳

Xtigyro (approved):

charts/prometheus/README.md
```diff
@@ -433,7 +169,7 @@ serverFiles:
 ```
 
 ```console
-helm install stable/prometheus --name my-release -f values.yaml -f service1-alert.yaml -f service2-alert.yaml
+helm install -f values.yaml -f service1-alert.yaml -f service2-alert.yaml ...etc
```
Xtigyro, line 172: Can we include please only working examples - just noticed this one?

Xtigyro (changes requested):

charts/prometheus/README.md
scottrigby: Fixed in adb4e21

Xtigyro (approved):

scottrigby:
Tests were failing for `prometheus-couchdb-exporter` (example [lint-test](https://github.com/scottrigby/prometheus-helm-charts/pull/14/checks?check_run_id=1005193296) GH Actions job). Had to fix k8s 1.16 deprecated PodSecurityPolicy in the extensions/v1beta1 API version. See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ I'm guessing this chart hasn't been modified in a while 😛

scottrigby:
**Update:** All initial chart readme changes have been pushed. Waiting on tests. Please give this a once over review. Once this is approved, we'll merge and move ahead 😅

charts/prometheus-couchdb-exporter/README.md
```diff
@@ -1,80 +1,68 @@
...
-$ helm install stable/prometheus-couchdb-exporter
+## Prerequisites
+
+- Kubernetes 1.8+ with Beta APIs enabled
+
+## Get Repo Info
+
+```console
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo add stable https://kubernetes-charts.storage.googleapis.com/
```
gkarthiks, line 15: @scottrigby do you still want to have the stable here?

zanhsieh (approved):

okgolove (approved):

charts/prometheus-couchdb-exporter/README.md
scottrigby: Good catch. No we only want this on the prometheus chart, since only it and the deprecated still older named `prometheus-operator` chart (see #9) have dependencies in the stable repo.

scottrigby:
PS re-running gh action job. This appears to be an API blip:
>  ✖︎ prometheus-consul-exporter => (version: "0.1.6", path: "charts/prometheus-consul-exporter") > Error validating maintainer 'timm088': 502 Bad Gateway

Just tested locally and it validates fine:
```console
$ docker run --rm -it -v $(pwd):/app:cached quay.io/helmpack/chart-testing:v3.0.0
$ / #
$ / # cd /app
$ /app # ct lint --charts /app/charts/prometheus-consul-exporter/
...
1 chart(s) linted, 0 chart(s) failed
------------------------------------------------------------------------------------------------------------------------
 ✔︎ prometheus-consul-exporter => (version: "0.1.7", path: "/app/charts/prometheus-consul-exporter/")
------------------------------------------------------------------------------------------------------------------------
All charts linted successfully
```

However, now getting similar errors:
>  ✖︎ prometheus => (version: "11.12.1", path: "charts/prometheus") > Error validating maintainer 'gianrubio': 429 too many requests

According to GitHub everything is fine https://www.githubstatus.com/ 🤔 

We may have to just try to run the job again 🤦 ⏳

scottrigby:
😑 https://github.com/scottrigby/prometheus-helm-charts/runs/1007003791

I've honestly no idea which GH user `GITHUB_TOKEN` ties to for [GH API rate limits](https://developer.github.com/v3/rate_limit/). Don't seem to see this in [the docs for this token](https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret) in actions. It doesn't seem to be my user though:
```
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4992
X-RateLimit-Reset: 1597920256
```

Fine, I'll go back to sleep and re-run later 😴 

**PS:** If anyone wants to join us at the [Deep Dive Into Helm](https://kccnceu20.sched.com/event/Zx4K/deep-dive-into-helm-scott-rigby-codeacademy-paul-czarkowski-pivotal) session today, it starts in a little over 2 hours 🕣 See some of you then!

scottrigby:
Woot! tests have passed! Let's get to the next steps. Not all maintainers of these charts have participated yet, but they may or may not be available. The goal of this PR is to change at least the charts readmes so the initial helm repo indexing will contain only updated info, and relevant to the new repo (we had to fix things along the way only because some of these charts have not been touched for some time). In future PRs, they will only be merged once at least one maintainer of the affected charts approve the PR. For that reason I recommend future PRs try to touch only one chart at a time like we do in the stable and incubator git source repo (github.com/helm/charts).

scottrigby:
**Update:** Maintainers please check https://github.com/scottrigby/prometheus-helm-charts/issues/11 for another poll to see if there are any remaining blockers for the repo transfer to `prometheus-community` :shipit: 

There are also testing steps in that issue, if anyone wants to try it out now 🚀

gkarthiks (commented):
approved

PR #32, opened 2020-08-27
fix errors reported by superlinter

torstenwalter:
The goal of this is to resolve all errors reported by superlinter or to disable the linters which report errors.
That way we would not block contributions. Disabled linters can be enabled one by one in separate PRs if desired together with fixing issues reported by them.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,5 +16,10 @@ jobs:
         uses: github/super-linter@v3
         env:
           VALIDATE_ALL_CODEBASE: false
+          VALIDATE_BASH: false
+          VALIDATE_PYTHON: false
+          VALIDATE_PYTHON_FLAKE8: false
+          VALIDATE_PYTHON_BLACK: false
+          VALIDATE_YAML: false
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,3 +1,3 @@
-## Community Code of Conduct
+# Community Code of Conduct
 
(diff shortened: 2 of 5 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
```
scottrigby, line 1: I'll merge master into the kube-prometheus PR so this file doesn't appear the diff. Already committed in: https://github.com/scottrigby/prometheus-helm-charts/pull/27

.github/workflows/linter.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v2
+      - name: Lint Code Base
+        uses: github/super-linter@v3
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          VALIDATE_BASH: false
+          VALIDATE_PYTHON: false
+          VALIDATE_PYTHON_FLAKE8: false
+          VALIDATE_PYTHON_BLACK: false
+          VALIDATE_YAML: false
```
scottrigby, line 23: Isn't `VALIDATE_YAML` the only one we need to disallow (because gotemplate in helm "yaml" files will never validate)? Just curious why we need to bother disallowing the others

torstenwalter: Unfortunately no. There is a bash script and some python code where the linter complains as well.

.github/stale.yml
scottrigby: Done

.github/workflows/linter.yml
scottrigby: That's fine. This seems worth it. If in future it becomes a pain, someone can always make a GH Action for DavidAnson/markdownlint only.

PR #1, opened 2020-08-09
[prometheus-operator] Rename to kube-prometheus-stack

scottrigby:
Fixes #9 

## Background

See https://github.com/prometheus-community/community/issues/28#issuecomment-670406329

## Description

This is an attempt to rename the chart. There are a _lot_ of changes needed, so it's taken some time to review.

I have also dramatically simplified the README per new charts template decision in #14. (tl;dr, `helm show values` gives all the info a user needs, and makes that the older convention of duplicating these value configurations in the README unnecessary). Without deep experience using this chart, I can't know if the existing upgrading sections are still relevant, so I organized and left them in case they are useful to end users.

## PR status

- [x] All feedback has been addressed so far
- [x] Chart tested both for new installs, and upgrade path from `stable/prometheus-operator`.
    - See https://github.com/scottrigby/prometheus-helm-charts/pull/1#issuecomment-678551180 and https://github.com/scottrigby/prometheus-helm-charts/pull/1#discussion_r474766376 for reasoning for manual testing steps (tl;dr, one of this chart's dependencies has also moved here, so to test installing before we transfer this repo, you need to manually edit `requirements.yaml` and change `https://prometheus-community.github.io/helm-charts` to `https://scottrigby.github.io/prometheus-helm-charts`. The URL will be correct after repo transfer)
- [x] @brancz please address https://github.com/scottrigby/prometheus-helm-charts/pull/1#issuecomment-678617365 by @Xtigyro (`prometheus` chart maintainer) about the new name chosen, and make a final decision.
    - This needs a final say one way or the other. Some alternatives have been suggested, while most others agree `kube-prometheus` is the best name. Be sure to "load more" comments to see the discussion through to https://github.com/scottrigby/prometheus-helm-charts/pull/1#issuecomment-678969511. If you decide to use a different name I will update the PR accordingly
- [x] Rename the chart `kube-prometheus-stack` per https://github.com/prometheus-community/helm-charts/pull/1#issuecomment-687072186

Once steps above are complete, we can merge this PR. Merging will release the new chart package and add to the helm repo index.

**Note:** merging this PR has not been identified as a blocker to transferring the repo to `prometheus-community`. See https://github.com/scottrigby/prometheus-helm-charts/issues/11

PR #33, opened 2020-09-02
disabled failing linters and fixed markdown issues (#32)

torstenwalter:
- fixes markdown issues reported by markdownlint
- disabled yamllint as helm templates are never valid
- disabled the other linters as there is a problem with a shell script and some python code
  once that is fixed we could enable them again

Note: After cherry-picking this I removed changes for prometheus-operator  as thos are better done in #1 (otherwise I would need to bump the version because of a minikube README change)

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,5 +16,10 @@ jobs:
         uses: github/super-linter@v3
         env:
           VALIDATE_ALL_CODEBASE: false
+          VALIDATE_BASH: false
+          VALIDATE_PYTHON: false
+          VALIDATE_PYTHON_FLAKE8: false
+          VALIDATE_PYTHON_BLACK: false
+          VALIDATE_YAML: false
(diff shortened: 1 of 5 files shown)


.github/workflows/linter.yml
```diff
@@ -16,5 +16,10 @@ jobs:
         uses: github/super-linter@v3
         env:
           VALIDATE_ALL_CODEBASE: false
+          VALIDATE_BASH: false
+          VALIDATE_PYTHON: false
+          VALIDATE_PYTHON_FLAKE8: false
+          VALIDATE_PYTHON_BLACK: false
+          VALIDATE_YAML: false
```
gkarthiks, line 23: just curious on why we are not validating the YAML

gkarthiks: gotcha!!! because the helm files are not a proper yaml files.

scottrigby (approved):

PR #39, opened 2020-09-05
added README.md to have an index.html generated too

monotek:
Signed-off-by: André Bauer <monotek23@gmail.com>

added README.md to have an index.html generated for http://prometheus-community.github.io/ too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ You can then run `helm search repo prometheus-community` to see the charts.
 
 ## Contributing
 
-We'd love to have you contribute! Please refer to our [contribution guidelines](CONTRIBUTING.md) for details.
+The source code of all Prometheus community Helm charts can be found on Github: <https://github.com/prometheus-community/helm-charts/>
 
-The Helm chart source code can be found on Github in: <https://github.com/prometheus-community/helm-charts/>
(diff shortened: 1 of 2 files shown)


scottrigby:
I agree a readme for gh-pages would be nice for end users.

Somewhere – in another Helm repo with someone (maybe @torstenwalter ?) we discussed having a GitHub action that automatically syncs the `main` branch README to the `gh-pages` branch when updated. I think this is a good idea for this repo, and would be pretty easy to do. The few changes/improvements you made here could also be worked into the main README, to make this easier to maintain in the long run, right?

scottrigby:
I added a GH Actions workflow file to do this in #41 

@monotek once that is merged, would you like to instead make a PR (or update this one) with suggested improvements only to the `README.md` file in `main` – so that when merged, it will then auto-sync to `gh-pages`?

monotek:
Good idea to automate it this way. I'll do the pr tomorrow :)

sbussard:
What happened to http://prometheus-community.github.io/ ?

monotek:
Never existed. 
You're looking for: https://prometheus-community.github.io/helm-charts/

PR #69, opened 2020-09-08
[prometheus-blackbox-exporter] Add additional metrics relabels blackbox and update to 0.17.0

xbglowx:

Allow passing in additional metrics relabels for blackbox-exporter.

One useful example would be to add an owner label that would help with routing alerts via Alertmanager

N/A

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.4.0
+version: 4.4.1
 appVersion: 0.17.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.4.1
+version: 4.5.0
 appVersion: 0.17.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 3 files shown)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.4.0
-appVersion: 0.16.0
+version: 4.4.1
```
desaintmartin, line 4: Could you bump minor version instead of patch version? This is a new feature according to semver.

charts/prometheus-blackbox-exporter/templates/servicemonitor.yaml
```diff
@@ -30,6 +30,10 @@ spec:
         replacement: {{ .url }}
       - targetLabel: target
         replacement: {{ .name }}
+        {{- range $targetLabel, $replacement := .additionalMetricsRelabels }}
```
desaintmartin, line 33: Should we have defaults as well?

desaintmartin (changes requested):
thanks for the PR!

charts/prometheus-blackbox-exporter/Chart.yaml
xbglowx: Sure thing.

charts/prometheus-blackbox-exporter/templates/servicemonitor.yaml
xbglowx: That should be set with https://github.com/prometheus-community/helm-charts/pull/69/files#diff-711ccd42b755d858f62f44246a271e46R132 ?

xbglowx: Oh, wait, I see what you mean.

desaintmartin (approved):
Thanks!

PR #79, opened 2020-09-10
[kube-prometheus-stack] add secret-field-selector

Sayrus:

Following prometheus-operator/prometheus-operator#3355, add support for this new option. This should be released in v0.41.0.
This option is used to filter out secrets that are usually watched by the operator but not useful. This includes filtering out Helm V3 secrets which can cause significant CPU and Memory usage or to select only Prometheus Secret if you want to tag them all with a type (or any Field Selector actually).

Old PR: helm/charts#23330

None

This should not introduce any breaking change and should work (not enable the option but not crash the operator) before v0.41.0. I did not version-gate the option for this reason.

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-operator/kube-prometheus
   - https://github.com/prometheus-operator/prometheus-operator
-version: 9.3.4
+version: 9.3.5
 appVersion: 0.38.1
 tillerVersion: ">=2.12.0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


Xtigyro:
@Sayrus Thank you very much for the contribution! This is looking good! 💯 

Could you please instead of bumping the `patch` version number - bump the `minor` version number because this is adding new functionality to the chart?

Sayrus:
@Xtigyro  Done !

Can add this to CONTRIBUTING.md to avoid others doing this mistake?

Xtigyro:
> 
> 
> @Xtigyro Done !
> 
> Can add this to CONTRIBUTING.md to avoid others doing this mistake?

@Sayrus Yes - this seems like a very reasonable idea. Please do in a separate PR.

Also - it appears you haven't signed your last commit.

Xtigyro (approved):
LGTM

PR #80, opened 2020-09-10
[prometheus-node-exporter] Add rootfs mount by default

palm93:
Signed-off-by: Dan Jones <dan@palm93.com>

@gianrubio 
@vsliouniaev 

Adds rootfs mount by default, resolving an issue where root filesystem metrics are incorrectly reported if the container runtime storage is not on the host root filesystem.

issue from old repo: https://github.com/helm/charts/issues/22905

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.0.1
 description: A Helm chart for prometheus node-exporter
 name: prometheus-node-exporter
-version: 1.11.2
+version: 1.11.3
 home: https://github.com/prometheus/node_exporter/
 sources:
 - https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -38,6 +38,7 @@ spec:
           args:
             - --path.procfs=/host/proc
             - --path.sysfs=/host/sys
+            - --path.rootfs=/host/root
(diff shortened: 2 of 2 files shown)


monotek:
Not stale.

@gianrubio @vsliouniaev 
Please review.

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.0.1
 description: A Helm chart for prometheus node-exporter
 name: prometheus-node-exporter
-version: 1.11.2
+version: 1.11.3
```
monotek, line 5: New feature should raise minor version.

monotek (changes requested):

charts/prometheus-node-exporter/Chart.yaml
palm93: Thanks, updated. :)

monotek (commented):
I'm not sure if this should be optional.

@gianrubio @vsliouniaev 
Thoughts?

vsliouniaev (approved):
Looks good to me, were running the same configuration

xxbbxb:
Hi everybody, I'm having an issue with this patch: docker is unable to mount `/` with error:
```
failed to start container "node-exporter": Error response from daemon: path / is mounted on / but it is not a shared or slave mount
```
We need either remove mountPropagation option or make root shared among namespaces `mount --make-shared /`
corresponding kubernetes issue https://github.com/kubernetes/kubernetes/issues/61058#issuecomment-372516782

monotek:
Please create a new issue using the issue template so we have all info about your environment.

PR #89, opened 2020-09-11
[kube-prometheus-stack] Update README.md

justinwalz:
### What this PR does / why we need it:

Looks like the generate command was renamed here https://github.com/prometheus-operator/kube-prometheus/commit/0f6cd6d0a8881ad7e80f2817c597dab8345f8790#diff-b67911656ef5d18c4ae36cb6741b7965

#### Checklist
I don't think we need to bump version number here. 

- [x] Chart Version bumped 


scottrigby:
The reason the version needs to bump with any change in the chart directory is because helm chart version packages must be immutable. CI checks for that reason. This sounds like a PATCH bump.

justinwalz:
Oh makes sense. Pushed an updated commit with a patch version update.

bismarck (approved):
lgtm

PR #115, opened 2020-09-19
[prometheus-adapter] added psp for prometheus adapter

ahmednasir91:
Signed-off-by: Ahmed Nasir <ahmednasir91@gmail.com>

Added Pod Security Policy for prometheus adapter and also fixed the condition for role binding creation of custom metrics.

The PR was created here but since repo moved, so again creating here.
https://github.com/helm/charts/pull/23536

#### Checklist
- [X] Chart Version bumped


monotek (changes requested):
New feature should raise minor version.
So it should be 2.6.0.

ahmednasir91:
@monotek updated version to 2.6.0

monotek (dismissed):

charts/prometheus-adapter/templates/custom-metrics-apiserver-psp.yaml
```diff
@@ -0,0 +1,71 @@
...
+  runAsGroup:
+    rule: MustRunAs
+    ranges:
+    - max: 0
+      min: 0
+  runAsUser:
+    rule: MustRunAs
+    ranges:
+    - max: 10001
+      min: 10001
```
steven-sheehy, line 27: Would be better to specify as min 1024 and max 65535. There's no explicit reason it has to be 10001 except that we arbitrarily hardcoded the chart to that. We'll probably want to make the security policies configurable on the chart at some point.

charts/prometheus-adapter/templates/custom-metrics-apiserver-psp.yaml
```diff
@@ -0,0 +1,71 @@
...
+spec:
+  {{- if .Values.hostNetwork.enabled }}
+  hostNetwork: true
+  {{- end }}
+  fsGroup:
+    rule: RunAsAny
+  runAsGroup:
+    rule: MustRunAs
+    ranges:
+    - max: 0
```
steven-sheehy, line 21: Would prefer RunAsAny since we probably want to switch this to non root group soon.

steven-sheehy (commented):
Some minor recommendations

charts/prometheus-adapter/templates/custom-metrics-apiserver-psp.yaml
ahmednasir91: Sounds good

charts/prometheus-adapter/templates/custom-metrics-apiserver-psp.yaml
ahmednasir91: Ok later we can change to RunAsNonRoot

steven-sheehy (approved):

ahmednasir91:
ah there was an issue in this one related to port range, fixed it in https://github.com/prometheus-community/helm-charts/pull/135

PR #125, opened 2020-09-22
[prometheus-blackbox-exporter] Allow custom labels for pods. Add service labels to values.yaml.

metroshica:
Allow custom labels for pods. Also adds label option to service entry in values.yaml that was missed in commit [d696fd4c388fdece1026491248be0b45412c163a](https://github.com/prometheus-community/helm-charts/commit/d696fd4c388fdece1026491248be0b45412c163a)

I created a top-level `pods` entry in `values.yaml` to mimic other top-level resource values such as `service`, `ingress`, etc. However, there is currently a top-level entry named podAnnotations that I feel should fit under it. I however didn't change that key to be under `pods` as it would be a breaking change.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.5.2
+version: 4.5.3
 appVersion: 0.17.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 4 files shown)


desaintmartin:
Thanks!

Indeed, it makes sense, maybe 1/ add comment near podAnnotations stating this 2/ bump major version and break 3/ do not use pod.

I would need other opinions on this one!

charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.5.2
+version: 4.5.3
```
monotek, line 4: new feature should raise minor version to 4.6.0.

monotek (changes requested):

charts/prometheus-blackbox-exporter/Chart.yaml
metroshica: Bumped version to 4.6.0.

monotek (approved):

PR #135, opened 2020-09-23
[prometheus-adapter] bugfix psp port range

ahmednasir91:
Fixes the bug in prometheus adapter psp where the port range defined was incorrect max > min.

```
 Error: cannot patch "prometheus-adapter" with kind PodSecurityPolicy: PodSecurityPolicy.extensions "prometheus-adapter" is invalid: spec.runAsUser.ranges[0].min: Invalid value: 65535: min cannot be greater than max
``` 

#### Checklist
- [X] Chart Version bumped


steven-sheehy (dismissed):
Will need to re-bump

ahmednasir91:
@steven-sheehy bumped

steven-sheehy (approved):

PR #139, opened 2020-09-24
[prometheus-blackbox-exporter] Feat use external secret for configura…

### LINKED ISSUE #138: [prometheus-blackbox-exporter] config.file values is not correct when providing secret to use
**Describe the bug**

We are using SealedSecret to manage our secrets on our clusters and also the secret containing the blackbox-exporter configuration because it contain basic auth credentials.

The creation of the secret is managed outside the chart so I set :
```
secretConfig: true
config: false
```

the secret is mounted in the pod but the `--config.file` match `/etc/blackbox_exporter/config.yml` instead of the mounted secret `/config/blackbox.yaml`

**Which chart**: prometheus-blackbox-exporter
**Which version of the chart**: 4.5.1

create a secret outside the chart, and set :
```
secretConfig: true
config: false
```

MCyprien:
Add a parameter to allow creation of configuration outside the chart as secret. (Using SealedSecret for example)

  - fixes #138 

#### Checklist

- [] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [] Chart Version bumped
- [] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.6.0
+version: 4.6.1
```
desaintmartin, line 4: Could you bump minor version instead of patch? This is new feature according to semver.

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -46,6 +46,9 @@ tolerations: []
 affinity: {}
 
 secretConfig: false
+# if the configuration is managed as secret outside the chart, using SealedSecret for example,
+# provide the name of the secret here
+configExistingSecretName: ""
```
desaintmartin, line 51: Maybe state that this is ignored if secretConfig is set ? (and document what is secretConfig, as those two could be misunderstood)

desaintmartin (changes requested):
Thanks for the PR!

charts/prometheus-blackbox-exporter/values.yaml
MCyprien: I  changed the comment and the order of variable to group secretConfig and config as they work together, is it better now ?

desaintmartin: Did you push? I don't see the changes.

MCyprien: my bad, forgot to push...

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -45,6 +45,10 @@ nodeSelector: {}
 tolerations: []
 affinity: {}
 
+# if the configuration is managed as secret outside the chart, using SealedSecret for example,
+# provide the name of the secret here. If secretConfig is set to true, configExistingSecretName will be ignored
+# in favor of the config value.
+configExistingSecretName: ""
 secretConfig: false
```
desaintmartin, line 52:
```suggestion
# Store the configuration as a `Secret` instead of a `ConfigMap`, useful in case it contains sensitive data
secretConfig: false
```

desaintmartin (changes requested):
Just a small suggestion to improve understanding of those two fields.

desaintmartin:
@MCyprien would you like to continue with this PR?

monotek:
@desaintmartin 
I've update the chart version and used githubs "commit suggestion" feature to add the note you've added to the values file, making you the commit author.
Unfortunately dco check does not pass now :(
Would you mind tu pull the branch and fix/sign your commit?

desaintmartin:
You are right, so that we can merge this! Done.

desaintmartin (approved):

monotek (approved):

PR #149, opened 2020-09-28
[kube-prometheus-stack] Add details to access dashboards

fai555:
Added the necessary commands and credentials to access the prometheus, grafana and alert-manager dashboards

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -35,6 +35,25 @@ _See [configuration](#configuration) below._
 
 _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documentation._
 
+## Access Dashboards
+**Access Prometheus Dashboard**
+```console
+$ kubectl  port-forward svc/my-release-kube-prom-prometheus 9090:9090
+```
+Then you can access the dashboard on http://localhost:9090
+
+**Access Grafana Dashboard**
+```console
+$ kubectl  port-forward svc/my-release-grafana 3000:80
+```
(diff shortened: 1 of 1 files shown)


gkarthiks:
Hi @fai555, thank you for the PR. As these are generic commands to do port-forwarding and not a production grade way of running/viewing dashboards, we will pass on this PR. 

Hope to see you contribute more with Prometheus Community.

/cc @scottrigby

fai555:
@gkarthiks I understand. What happens is that right after going through the installation step naturally you want to test out if everything is working. The current doc doesn't really help to do that.

You have to go through previous project's repos to get the info. So I made the PR.

scottrigby:
@fai555 hello 👋 documentation help is very appreciated! But I agree in this case the goal moving forward should be to further simplify documentation, adding only the required info for installing, upgrading, configuring, and uninstalling charts. Where additional info is absolutely necessary to note, we should link out to existing external docs if possible.

There's also a certain amount of general k8s concepts and knowledge users must be expected to possess - or be able to find elsewhere - in order to manage helm charts (like listing services and port forwarding). I think this has to remain outside the scope of each chart's README, because where would it end?

But in the spirit of sharing, here's a fun note about port forwarding a lot of people don't know: binding with port "0" will automatically assign an unused port from the dynamic range (49152-65535 l. See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml). I periodically get the opportunity to recommend this for local dev, like now 🙂 Similarly I don't think this should go in any chart README, and it's not even Helm specific. But it might be useful to note _somewhere_ easy to find for local k8s development.

PR #166, opened 2020-09-30
[prometheus] Proposition to not scrape not running pods

mtparet:
Prometheus try to scrape not running pods which cause reporting failures to scrape these ones. 

If it seems legit, I will review the PR to follow the contribution guidelines.

Best,

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 11.16.0
+version: 11.16.1
 appVersion: 2.21.0
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -1521,6 +1521,9 @@ serverFiles:
           - source_labels: [__meta_kubernetes_pod_name]
             action: replace
             target_label: kubernetes_pod_name
+          - source_labels: [__meta_kubernetes_pod_phase]
+            regex: Pending|Succeeded|Failed
+            action: drop
 
       # Example Scrape config for pods which should be scraped slower. An useful example
       # would be stackriver-exporter which querys an API on every scrape of the pod
@@ -1560,6 +1563,9 @@ serverFiles:
           - source_labels: [__meta_kubernetes_pod_name]
             action: replace
             target_label: kubernetes_pod_name
+          - source_labels: [__meta_kubernetes_pod_phase]
+            regex: Pending|Succeeded|Failed
+            action: drop
 
 # adds additional scrape configs to prometheus.yml
 # must be a string so you have to add a | after extraScrapeConfigs:


zanhsieh:
@mtparet 
Could you bum the Chart version to latest please?

naseemkullah:
It might make sense to first get it (and any prometheus.yml changes) merged in the [promtheus-kubernetes.yml](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml) from the app repo, and then this chart can base the out of box config from there.

@prometheus-community/helm-charts-admin thoughts?

Xtigyro:
> 
> 
> It might make sense to first get it (and any prometheus.yml changes) merged in the [promtheus-kubernetes.yml](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml) from the app repo, and then this chart can base the out of box config from there.
> 
> @prometheus-community/helm-charts-admin thoughts?

That might be a good idea but can the two files be absolutely the same at some point? I don't know whether the Prom devs would want that.

Xtigyro (changes requested):
@mtparet The chart version needs bumping. Also - could you please sign your commits?

brancz:
I don't think the files being identical is the point, it's just that the upstream config should reflect configuration that applies to everyone, which this case falls into.

Xtigyro:
@brancz Agreed.

kfox1111:
if it avoids scraping a non running pod, will prometheus still know to report that endpoint as up=0?

monotek (changes requested):
This should be optional and default should be disabled, to not break current behaviour.

Minor Chart version needs to be raised. 

Commits needs to be signed: https://github.com/prometheus-community/helm-charts/pull/166/checks?check_run_id=1188733204

mtparet:
> @mtparet The chart version needs bumping. Also - could you please sign your commits?

@Xtigyro done :)


> Minor Chart version needs to be raised.
Commits needs to be signed: #166 (checks)

@Xtigyro done :)


> if it avoids scraping a non running pod, will prometheus still know to report that endpoint as up=0?

This metrics does not come from scrapping pods but from kube state metrics.

> This should be optional and default should be disabled, to not break current behaviour.

If the pods are not running, we already didn't collect any metrics when trying to scrape these one. So I do not think this is a breaking change. I'm happy to understand/hear the contrary :)

Xtigyro:
@naseemkullah @monotek Are we happy to merge this now? I think it should be ready.

zanhsieh (approved):

naseemkullah:
I still think config changes should go through [promtheus-kubernetes.yml](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml) 🤷‍♀️ but don't want to block so do as you please.

Xtigyro:
@naseemkullah I agree with ya. What can we do to achieve that?

naseemkullah:
@mtparet would it be possible to make a PR to https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml ?

@Xtigyro we could add a note in readme that `prometheus.yml` changes should be made in that repo then they can make the PR here or maintainers periodically sync from there to here.

monotek (approved):

monotek:
Imho the sync issue should not prevent us to merge this.

Xtigyro:
Agreed - let's merge this and create a PR in the Prom repo and try to be in sync with it.

Xtigyro (approved):

mtparet:
Thanks for merging!

@naseemkullah I will do that:)

mtparet:
I just took a look at https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml there are already lot of difference with the current state of the file here.

I'm not sure if the purpose of the example is to be the exactly config file to be applied in production and if Prometheus maintainers want to maintain it in their repo.

Xtigyro:
@mtparet They don't need to be identical but the upstream config should reflect configuration that applies to everyone - so you can add just your code there from this PR.

PR #194, opened 2020-10-09
[prometheus-blackbox-exporter] Added support of sidecars and extraVolumes

pburgisser:
Signed-off-by: Philippe Bürgisser <philippe.burgisser@camptocamp.com>

- Adding support of multiple container in the same pod (side car, like oAuth-proxy)
- Adding support of multiple secret (necessary for oauth)
- Adding support of annotation on service, service account
- Adding support of authenticated Service Monitor

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/deployment.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/deployment.yaml
@@ -90,7 +90,7 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           ports:
-            - containerPort: {{ .Values.port }}
+            - containerPort: {{ .Values.service.port }}
               name: http
           livenessProbe:
             {{- toYaml .Values.livenessProbe | trim | nindent 12 }}
--- a/charts/prometheus-blackbox-exporter/templates/deployment.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/deployment.yaml
@@ -58,6 +58,9 @@ spec:
       priorityClassName: "{{ .Values.priorityClassName }}"
       {{- end }}
       containers:
+        {{- if .Values.extraContainers }}
+{{ toYaml .Values.extraContainers | indent 8 }}
+        {{- end }}      
         - name: blackbox-exporter
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
@@ -87,7 +90,7 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           ports:
-            - containerPort: {{ .Values.service.port }}
+            - containerPort: {{ .Values.port }}
               name: http
           livenessProbe:
             {{- toYaml .Values.livenessProbe | trim | nindent 12 }}
@@ -116,6 +119,9 @@ spec:
         {{- toYaml .Values.dnsConfig | nindent 8 }}
 {{- end }}
       volumes:
+    {{- if .Values.extraVolumes }}      
+{{ toYaml .Values.extraVolumes | indent 8 }}      
+    {{- end }}      
         - name: config
 {{- if .Values.secretConfig }}
           secret:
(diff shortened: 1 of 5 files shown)


charts/prometheus-blackbox-exporter/templates/deployment.yaml
```diff
@@ -87,7 +90,7 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           ports:
-            - containerPort: {{ .Values.service.port }}
+            - containerPort: {{ .Values.port }}
```
monotek, line 93: Why the change?
Now port is configurable twice?

charts/prometheus-blackbox-exporter/templates/servicemonitor.yaml
```diff
@@ -16,7 +16,15 @@ metadata:
 spec:
   endpoints:
   - port: http
-    scheme: http
+    {{- if .Values.serviceMonitor.scheme }}
+    scheme: {{ .Values.serviceMonitor.scheme }}
```
monotek, line 20: this breaks current behaviour as the value would change.
imho this would make this a breaking change.

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/deployment.yaml
pburgisser: If you add a oAuthProxy in front, you need to differentiate the service port and the container port, so there is not direct access to the blackbox.
In Prometheus-pushgateway, the port if hard coded [1], maybe we can hard code it too ?

[1] https://github.com/prometheus-community/helm-charts/blob/a72459c5a1a5672c0bcd812a5a078ec4517693d2/charts/prometheus-pushgateway/templates/deployment.yaml#L40

charts/prometheus-blackbox-exporter/templates/servicemonitor.yaml
pburgisser: Good catch !
Changed the default values to http so I can keep changing the scheme in case I'm doing authenticated prometheus (with oAuth proxy on OCP)

charts/prometheus-blackbox-exporter/templates/deployment.yaml
monotek: Why don't you add a targetPort key to the service and make it configurable by a variable like `targetPort: {{ .Values.service.targetPort }}`?
With this you could set the proxy to be accessed by the service instead of the alertmanager container.

pburgisser: OK, I did as you suggested. I hope that works for you now.

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -96,7 +126,7 @@ service:
   labels: {}
   type: ClusterIP
   port: 9115
-
+  targetPort: 9115
```
monotek, line 129: Please add blank line below again

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -39,6 +68,7 @@ runAsUser: 1000
 readOnlyRootFilesystem: true
 runAsNonRoot: true
 
+
```
monotek, line 71: Please remove double blank line here.

monotek (changes requested):

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -13,6 +13,35 @@ podDisruptionBudget: {}
...
 
+extraVolumes: []
+  # - name: secret-blackbox-oauth-htpasswd
+  #   secret:
+  #     defaultMode: 420
+  #     secretName: blackbox-oauth-htpasswd
+  # - name: storage-volume
+  #   persistentVolumeClaim:
+  #     claimName: example
+extraContainers: []
```
monotek, line 24: Add blank line above please

pburgisser: There you go

monotek (approved):

PR #146, opened 2020-09-27
Add stable package history from before charts adoption

scottrigby:
See https://github.com/scottrigby/helm-adopt-package-history

Ran with:

```console
$ helm-adopt-package-history \
    --old-repo=stable=https://kubernetes-charts.storage.googleapis.com \
    --new-repo=prometheus-community=https://prometheus-community.github.io/helm-charts \
    --local-dir=$USER/code/github.com/prometheus-community/helm-charts \
    --exclude-charts=kube-prometheus-stack
```

There are a massive number of historic packages for all the stable/prometheus-* chart versions prior to moving to this repo (839). The total size is 188MB. Fine for GitHub, but an undertaking to review. Please check out the script above.

Also, there are far more line changes in `index.yaml` picked up by git than there should be. I reviewed carefully in an IDE (because I had to do some manual resolutions for package versions that were both listed as deprecated in `stable` and also the initial version in this repo). I wondered if some of those changes could be due to an initial mix of line endings, but I ran `dos2unix` and it's clear. This could be just due to similar, repeating lines when splicing in package history.

Small note: stable/prometheus-pushgateway 1.4.3 has been removed. It was added in stable after deprecating the chart, but should not have been, because it now conflicts with a release in prometheus-community/helm-charts. I'm choosing the community version since the stable PR was just a one-letter change. Keeping a re-named `prometheus-pushgateway-1.4.3-from-stable.tgz` for posterity. See helm/charts#23813

PR #200, opened 2020-10-12
[kube-prometheus-stack] Specify old chart name on update instructions

pedroslopez:
The instructions mention needing to update to specific versions, but these are invalid for the new name (prometheus-community/kube-prometheus-stack). I had to use the old name to perform these updates on an old installation.

#### Checklist

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 10.1.0
+version: 10.1.1
 appVersion: 0.42.1
 tillerVersion: ">=2.12.0"
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


monotek (changes requested):
Please raise bugfix version of chart

scottrigby:
I don't understand why this is the case, since we've now ported chart package version history to this repo as well. The package versions are immutable, so they should work exactly the same from this source as from stable. Can someone explain what the problem is? Did you perhaps try this before the package history pull request was merged and running `helm repo update` again?

pedroslopez:
I did this 4 days ago and had just added the repo to helm. I’d say getting the old versions on there would be ideal, but according to https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack it doesn’t seem like they’re available.

scottrigby:
@pedroslopez The older versions are in this chart repo. I know only because I added them (see https://github.com/prometheus-community/helm-charts/pull/146).

The confusion is probably due to the renaming. For example, `kube-prometheus-stack` versions only go back to `9.3.4` (see https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack). You can not see a similar entry for `prometheus-operator` in this chart repo, because the last release was marked `deprecated`, and deprecated charts do not appear in the hub.

However, you may use `helm search` to find them. Can you try the command in this issue comment, and see if your results are different? https://github.com/prometheus-community/helm-charts/pull/146#issuecomment-699677681

This proves the versions are present in this repo. They are also installable from this repo, just as they are from stable.

PR #220, opened 2020-10-16
[prometheus-cloudwatch-exporter] Allow Helm variables in config

rsotnychenko:
Signed-off-by: Rostyslav Sotnychenko <me@sota.sh>

One might want to collect the exact same set of metrics in different environments, with a difference only in tag filter value (`metrics.*.aws_tag_select.tag_selections.someLabel=XXXX`).
Given that definition of what metrics to collect takes most of the configuration file, it's handy to have an ability to change a single variable instead of overwriting the whole config from scratch. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-cloudwatch-exporter/templates/configmap.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/configmap.yaml
@@ -9,4 +9,4 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   config.yml: |
-{{ printf .Values.config | indent 4 }}
+{{ tpl .Values.config . | indent 4 }}
--- a/charts/prometheus-cloudwatch-exporter/templates/configmap.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/configmap.yaml
@@ -9,4 +9,4 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   config.yml: |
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
Maybe a hint in the values.yaml that the variable is templated would be nice.

rsotnychenko:
@monotek readme for this chart does not contain any information regarding chart's variables, but a link to `values.yaml`. So I added the note directly into `values.yaml`

charts/prometheus-cloudwatch-exporter/templates/configmap.yaml
```diff
@@ -9,4 +9,4 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   config.yml: |
-{{ printf .Values.config | indent 4 }}
+{{ tpl .Values.config . | indent 4 }}
```
torstenwalter, line 12:
```suggestion
    {{ tpl .Values.config . | nindent 4 }}
```

rsotnychenko: @torstenwalter that is done.

torstenwalter (approved):

torstenwalter:
@monotek For me this looks good now. You requested changes earlier on. I think those are already implemented. Could you approve and merge?

monotek (approved):

PR #224, opened 2020-10-16
[kube-prometheus-stack] Allow use of Prometheus Operator master image tag

rothsa:
Signed-off-by: Sally Lehman <slehman@ripple.com>

The latest master build of the prometheus-operator chart cannot currently be used with kube-prometheus-stack because the deployment.yaml restricts customization only to release numbers.

This modification allows for releases, as well as the latest [master build of prometheus operator](https://quay.io/repository/prometheus-operator/prometheus-operator?tab=tags
) to be used with the chart.

cc'ing chart maintainers mentioned in Chart.yaml, as recommended in the Pull request guidelines:
@vsliouniaev @bismarck @gianrubio @gkarthiks @scottrigby @Xtigyro 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -39,7 +39,8 @@ spec:
           {{- end }}
           imagePullPolicy: "{{ .Values.prometheusOperator.image.pullPolicy }}"
           args:
-            {{- if semverCompare "< v0.39.0" .Values.prometheusOperator.image.tag }}
+            {{- if eq .Values.prometheusOperator.image.tag "master" -}}
+            {{- else if (semverCompare "< v0.39.0" .Values.prometheusOperator.image.tag) -}}
             - --manage-crds={{ .Values.prometheusOperator.manageCrds }}
             {{- end }}
             {{- if .Values.prometheusOperator.kubeletService.enabled }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -39,7 +39,7 @@ spec:
           {{- end }}
           imagePullPolicy: "{{ .Values.prometheusOperator.image.pullPolicy }}"
           args:
-            {{- if eq .Values.prometheusOperator.image.tag "master" -}}
+            {{- if regexMatch "master.*" .Values.prometheusOperator.image.tag -}}
             {{- else if (semverCompare "< v0.39.0" .Values.prometheusOperator.image.tag) -}}
             - --manage-crds={{ .Values.prometheusOperator.manageCrds }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -39,7 +39,8 @@ spec:
           {{- end }}
           imagePullPolicy: "{{ .Values.prometheusOperator.image.pullPolicy }}"
           args:
-            {{- if semverCompare "< v0.39.0" .Values.prometheusOperator.image.tag }}
+            {{- if eq .Values.prometheusOperator.image.tag "master" -}}
```
scottrigby, line 42: @rothsa thanks I wasn't aware there was a `master` tag.

But just so I understand the issue, what happens when you configure `prometheusOperator.image.tag` to `master`? I haven't tested it but guessing `semverCompare` thinks `master` is `<` `v0.39.0`? Assuming yes (or whatever the reason), would you also add a comment above this line explaining why there's an empty `if` condition? I can easily imagine this being cleaned up in future if there's no why.

Just a thought - since I see there are multiple `master*` tags (`master-arm64` etc), it may be better to use [regexMatch ](http://masterminds.github.io/sprig/strings.html#regexmatch-mustregexmatch) check instead of checking only `master`. WDYT?

scottrigby (changes requested):

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
metroshica: @scottrigby Correct, it throws an error when trying to compare `master` to a semver. Here's what we were seeing when trying to set the `prometheusOperator.image.tag` to `master`:

`Error: UPGRADE FAILED: template: kube-prometheus-stack/templates/prometheus-operator/deployment.yaml:42:19: executing "kube-prometheus-stack/templates/prometheus-operator/deployment.yaml" at <semverCompare "< v0.39.0" .Values.prometheusOperator.image.tag>: error calling semverCompare: Invalid Semantic Version`

I'll let @rothsa answer the regex question but shouldn't be too hard to implement a `master*` regex.

scottrigby: Sounds fair 🙂

metroshica: Looks like `{{- if regexMatch "master.*" .Values.prometheusOperator.image.tag -}}` does the trick. I'll touch base with @rothsa this morning.

rothsa: Yes, this looks great. Metroshica is correct about the error. Thanks for the suggestion and the changes.

rothsa:
This is ready for review again. :)

metroshica:
@scottrigby Anything else needed here to get this merged?

bismarck (approved):
lgtm

bismarck:
@scottrigby Any idea why the linter is failing with `Error: no repository definition for https://kubernetes-charts.storage.googleapis.com/. Please add the missing repos via 'helm repo add'`?

bismarck:
@rothsa Could you update your fork one more time? Looks like we're getting hit by #260

rothsa:
@bismarck thanks, done.

scottrigby:
@rothsa  nice work 👏

PR #240, opened 2020-10-21
[alertmanager] Add support change default config

pcheliniy:
Default image prom/alertmanager use configfile called `alertmanager.yml`, but in the chart we use configfile with name `config.yml`.
Configuration from `.Values.config` is rendered correctly but it doesn't matter because we don't use this configfile.
Also, we don't have an option how to say alertmanager to use our configfile.
This PR allows us to:
* use this configuration file (fix)
* configure the name of the configuration file.
* add some extra args if it needed.

We have added `configFileName` as param just in case when someone uses a custom image and default alertmanager.yml contains some kind of reference configurations or examples.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -28,15 +28,14 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or .Values.extraArgs .Values.configFileName }}
           args: 
-            {{- range $key, $value := .Values.extraArgs }}
-            - --{{ $key }}={{ $value }}
-            {{- end }}
+            --storage.path=/alertmanager
             {{- if .Values.configFileName }}
             - --config.file=/etc/alertmanager/{{ .Values.configFileName }}
             {{- end }}
-          {{- end }}
+            {{- range $key, $value := .Values.extraArgs }}
+            - --{{ $key }}={{ $value }}
+            {{- end }}
           ports:
             - name: http
               containerPort: 9093
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -28,6 +28,15 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if or .Values.extraArgs .Values.configFileName }}
+          args: 
+            {{- range $key, $value := .Values.extraArgs }}
+            - --{{ $key }}={{ $value }}
+            {{- end }}
+            {{- if .Values.configFileName }}
+            - --config.file=/etc/alertmanager/{{ .Values.configFileName }}
+            {{- end }}
+          {{- end }}
           ports:
             - name: http
               containerPort: 9093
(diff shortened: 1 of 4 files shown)


charts/alertmanager/templates/statefulset.yaml
```diff
@@ -28,6 +28,14 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args: 
+            - --storage.path=/alertmanager
+            {{- if .Values.configFileName }}
```
naseemkullah, line 33: Instead of add a configFilename value, if the purpose is to omit the config file, can we insead check `{{- if .Values.conig }}` ? i.e. if we set `config: null` in the values file, it will not pass it in.

naseemkullah (changes requested):
Thanks @pcheliniy! re: omitting the config file, we could probably use `config: null` and evaluate `.Values.config`

charts/alertmanager/values.yaml
```diff
@@ -81,6 +83,7 @@ persistence:
     - ReadWriteOnce
   size: 50Mi
 
+configFileName: config.yml
```
naseemkullah, line 86: With the other fixes you've applied and evaluation of .Values.config for the if statement we could probably omit this option, WDYT?

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -28,6 +28,14 @@ spec:
...
+            - --config.file=/etc/alertmanager/{{ .Values.configFileName }}
+            {{- end }}
+            {{- range $key, $value := .Values.extraArgs }}
+            - --{{ $key }}={{ $value }}
+            {{- end }}
```
naseemkullah, line 38: 👏

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -28,6 +28,14 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args: 
+            - --storage.path=/alertmanager
+            {{- if .Values.configFileName }}
+            - --config.file=/etc/alertmanager/{{ .Values.configFileName }}
```
naseemkullah, line 34:
```suggestion
            - --config.file=/etc/alertmanager/{{ .Values.configFileName }}
```

👏 
As per other comments I think we can leave the filename hardcoded.

charts/alertmanager/templates/statefulset.yaml
pcheliniy: I'm not sure that I fully understand.
Do you mean that we have to check `{{- if .Values.config }}` and if it empty skip --config.filename ?
I'm not sure that it is possible because it will cause alertmanager to start finding config in local dir (default config path) and it will fail. To avoid it we will have to add some default behavior as some new helper or hardcode default value inplace.

I guess we should check {{- if .Values.config }} and if it is empty do not install ConfigMap, Volume, and skip corresponding VolumeMount. In my opinion, it is more predictable behavior.

charts/alertmanager/values.yaml
pcheliniy: The reason why we decided to do this described in PR notes. If this is not an important case I agree with you.

charts/alertmanager/templates/statefulset.yaml
pcheliniy: ok

charts/alertmanager/templates/statefulset.yaml
naseemkullah: ah yeah that makes sense. Maybe we skip the evaluation and always use the config file instead, that will be simpler

naseemkullah (approved):

PR #243, opened 2020-10-21
[prometheus] Make emptyDir settings consistent across components

unicell:

Currently only prometheus server takes `emptyDir.sizeLimit` for
`deployment` but not for `statefulset`. And alertmanager has no notion
about `emptyDir.sizeLimit` at all.

This change fixes that by following existing pattern of prometheus
server deployment spec, to allow `sizeLimit` to be set for both `server`
and `alertmanager`, regardless it is a deployment spec or statefulset.

On the pushgateway side, existing comment string in `values.yaml`
suggests if `persistentVolume.enabled` set to false, `emptyDir` will be
used. But that's actually not the case. Since the existing pushgateway
deployment spec by default has persistent volume turned off and doesn't
specify emptyDir at all (neither in volume nor  volumeMount), this
change fixes the inconsistency by updating the comment string in
values.yaml

Described above.

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 11.16.3
+version: 11.16.4
 appVersion: 2.21.0
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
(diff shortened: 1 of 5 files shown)


Xtigyro (changes requested):
Hey @unicell - thanks for the contribution, it is appreciated!

Could you please sign your commits?

unicell:
Sure, I GPG signed but forget the signoff signature. Let me update.

unicell:
@Xtigyro Could you double-check? Just updated the PR.

Xtigyro (approved):

Xtigyro:
@unicell LGTM! ⭐

unicell:
Thanks!!

Xtigyro:
> Thanks!!

And thank you as well!

PR #279, opened 2020-10-28
fix typos in kube-prometheus-stack values.yaml

derekperkins:


monotek (changes requested):
Please raise chart version.

PR #280, opened 2020-10-29
[kube-prometheus-stack] Update prometheus-operator to v0.43.0

bismarck:
As part of updating prometheus operator to v0.43.0
- Updated CRDs
- Updated RBAC
- Updated docs
- Removed tlsProxy and enabled native TLS support
- Replaced deprecated fields in `Alertmanager` and `Prometheus` objects
- Conditionally removed `--config-reloader-image` when using >=v0.43.0
  - fixes #

#### Checklist

- [x] Chart Version bumped


charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -32,7 +32,7 @@ spec:
 {{ toYaml .Values.prometheus.prometheusSpec.apiserverConfig | indent 4}}
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.image }}
-  baseImage: {{ .Values.prometheus.prometheusSpec.image.repository }}
+  image: {{ .Values.prometheus.prometheusSpec.image.repository }}
```
vsliouniaev, line 35: Unlike `baseImage`, this needs the `tag` appended. (Version is still required as part of the spec)

```
image: {{ .Values.prometheus.prometheusSpec.image.repository }}:{{ .Values.prometheus.prometheusSpec.image.tag }}
```

charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
 spec:
 {{- if .Values.alertmanager.alertmanagerSpec.image }}
-  baseImage: {{ .Values.alertmanager.alertmanagerSpec.image.repository }}
+  image: {{ .Values.alertmanager.alertmanagerSpec.image.repository }}
```
vsliouniaev, line 12: Same here, needs a tag at the end.

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
```diff
@@ -19,16 +19,17 @@ rules:
   - monitoring.coreos.com
```
vsliouniaev, line 19: Might be good to clean this up in line with https://github.com/prometheus-operator/prometheus-operator/blob/master/example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml

vsliouniaev (changes requested):
Should we drop `--config-reloader-image` flag from the operator deployment?

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
bismarck: I sync'd this up.

charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
bismarck: Thanks for catching this. Fixed!

charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
bismarck: Fixed!

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
bismarck: I sync'd this up.

bismarck:
@vsliouniaev The `--config-reloader-image` is going to be around until at least v0.45.0. I'm not sure if it makes sense to keep it around for people upgrading?

vsliouniaev:
Just tried running this install with helm3 and got
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "alertmanagerconfigs.monitoring.coreos.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "prom-op"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "monitoring"
```

I think that we should remove CRDs from this chart, since an upgrade using helm 2 doesn't work and using helm 3 to upgrade won't add `alertmanagerconfigs`. Someone expecting to use helm to manage CRDs will be in for a surprise sooner or later and would have to set up automation to handle upgrades

bismarck:
I think we need the CRDs in the crds directory at least for CI?  I removed the `crd-install` hooks from the CRDs since we can [no longer install them](https://github.com/helm/helm/issues/6783) with Helm 2. I also removed the `createCustomResource` and `cleanupCustomResource` and `manageCrds` options from the chart.

vsliouniaev (dismissed):
LGTM!

vsliouniaev (approved):

PR #285, opened 2020-10-30
[kube-prometheus-stack] kube-state-metrics serviceMonitor labelSeletor update

janlud:
`app.kubernetes.io/instance` is overriden by ArgoCD on the service and because of that this `serviceMonitor` fails to detect the service

Signed-off-by: Jan Ludvik <ludvik.honza@gmail.com>

ArgoCD overrides `app.kubernetes.io/instance` label with it's own value (`<project_name>-<release_name>`). If the service monitor using this label to look for a service, it will fail as the label is not what the chart will expect.
This will cause issues for all users of ArgoCD who didn't change the default ArgoCD label to something else.

I wanted to consolidate the labelSelector as other `serviceMonitor` have with app and release but the kube-state-metrics is an external chart.

I am using additional custom serviceMonitor as a workaround now but would like to get rid of that.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 10.3.2
+version: 10.3.3
 appVersion: 0.42.1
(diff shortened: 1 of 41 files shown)


monotek (changes requested):
As this is argocd specific removing the label should be optional.

janlud:
Accidenally made some mess in git, will reopen correctly. My first GH PR - apologies! :)

PR #301, opened 2020-11-04
[kube-prometheus-stack] Disable scraping kubelet resource metrics endpoint

bismarck:
Based on the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/kubelet-resource-metrics-endpoint.md) for the Kubelet Resource Metrics Endpoint, I don't believe we should be scraping this endpoint. In the past, it's caused [issues](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/458) with some of the metrics reported by cAdvisor. Also, the kube-prometheus project (which this chart is based on) [does not scrape](https://github.com/prometheus-operator/kube-prometheus/blob/master/manifests/prometheus-serviceMonitorKubelet.yaml) this endpoint.

Also, bumping the prometheus to v2.22.1 because it has fixes for a [retention size bug](https://github.com/prometheus/prometheus/pull/8139).
  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 11.0.1
+version: 11.0.2
 appVersion: 0.43.1
 tillerVersion: ">=2.12.0"
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


gkarthiks:
I guess adding a reasoning comment on the values.yml file will help the community to know why it is disabled by default. Totally open for suggestions. :)

vsliouniaev (approved):
Might want to rename the PR / fix the name so that the commit includes the prometheus bugfix in it

PR #307, opened 2020-11-05
[kube-prometheus-stack] Migrate to chart v2

mrueg:
Initial attempt to move to chart v2, since helm v2 is running out of support.

#### Checklist

- [x] Chart Version bumped


vsliouniaev:
We're currently stuck on Helm 2 and are relying on this chart. Could we delay doing this upgrade until there is need to break this compatability?

mrueg:
> We're currently stuck on Helm 2 and are relying on this chart. Could we delay doing this upgrade until there is need to break this compatability?

Grafana's chart has migrated over to helm3 already, I'm not sure if delaying means that we will be stuck on the current grafana?

r3kzi:
> We're currently stuck on Helm 2 and are relying on this chart. Could we delay doing this upgrade until there is need to break this compatability?

After November 13, 2020 no further Helm v2 releases (even for security patches) will be published. So IMHO we should not support Helm v2 features after this date.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -33,3 +33,17 @@ annotations:
...
+  repository: https://charts.helm.sh/stable
+  condition: kubeStateMetrics.enabled
+- name: prometheus-node-exporter
+  version: "1.12.*"
+  repository: https://prometheus-community.github.io/helm-charts
+  condition: nodeExporter.enabled
+- name: grafana
+  version: "5.8.*"
+  repository: https://grafana.github.io/helm-charts
+  condition: grafana.enabled
```
torstenwalter, line 49: I like it that you keep dependencies at the same version, while making the switch to helm 3.

torstenwalter:
I can only agree to what @r3kzi states. As helm 2 will no longer be supported in a couple of days I see no reason of holding back this change.

It's also the correct chart to start with as there should not be any issue with a helm 3 chart depening on a helm 2 chart.

torstenwalter:
@mrueg can you resolve conflicts?

mrueg:
@torstenwalter Thanks, I fixed the conflicts.

torstenwalter (approved):

torstenwalter:
I approved this as technically his changes look correct. For this to be merged a maintainer of the chart needs to approve and merge it.

charts/kube-prometheus-stack/Chart.yaml
mrueg: I plan to do a follow-up PR bumping grafana to a later version.

vsliouniaev (approved):

charts/kube-prometheus-stack/README.md
```diff
@@ -163,26 +161,6 @@ You can read more information on how to add firewall rules for the GKE control p
 
 Alternatively, you can disable the hooks by setting `prometheusOperator.admissionWebhooks.enabled=false`.
 
-### Helm fails to create CRDs
```
vsliouniaev, line 166: [Helm 3 will only install CRDs and not perform any updates on them afterwards, any changes to CRDs still require someone to maintain them outside of helm](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you). Do you think we should move this section into the upgrade instructions?

vsliouniaev (changes requested):

charts/kube-prometheus-stack/Chart.yaml
scottrigby: Only thing is, updating to the newer version of Grafana is the only argument I can see for moving this chart to v2 now.

charts/kube-prometheus-stack/README.md
scottrigby: Yes we should, if there were past versions of the CRDs, in the upgrading section for the chart version where the new CRD versions were introduced, we should note that they need to be manually upgraded to that newer version 👍

r3kzi: @mrueg can you accept the requested changes so we can proceed with merging the Pull Request?

vsliouniaev (approved):

vsliouniaev:
I'd be happy if we can update the Readme in a follow-up PR with a general statement about CRD behaviour.

charts/kube-prometheus-stack/README.md
mrueg: @r3kzi there's nothing I can accept from a Github UI, I added a paragraph to the update piece linking the doc that @vsliouniaev  mentioned.

mrueg:
@scottrigby @vsliouniaev @r3kzi would you agree that a CRD change should result in a major version bump of the chart as a signal? Maybe that's something we can include as well?

vsliouniaev:
I think that's a really good way to do it

vsliouniaev (approved):

PR #323, opened 2020-11-07
[kube-prometheus-stack] Allow setting persistence settings for grafana

wilsonge:
Allows persisting grafana storage to a PV - meaning that no dashboard setups get lost when pods get deleted.

#### Notes to reviewers
Relevent options in grafana helm chart https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L230-L244 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 11.0.4
+version: 11.0.5
 appVersion: 0.43.1
 tillerVersion: ">=2.12.0"
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


bismarck (changes requested):
Not sure why this is needed in the kube-prometheus-stack chart since you can already override this?  Dashboards should already persist across restarts since they're stored as configmaps.

Vickey-Wu:
I have the same request, i need to persistent my grafana dashbaord, when i delete and deploy it again i lost my dashboard

scottrigby:
@bismarck is right. A user may override any values of a dependent chart from the parent, it's not required to explicitly specify those in the parent chart's values file.

Vickey-Wu:
> @bismarck is right. A user may override any values of a dependent chart from the parent, it's not required to explicitly specify those in the parent chart's values file.

Hi, could you please show me how to override values to persistent grafana data even when i delete and deploy it again. Thank you!

wilsonge:
You need to set enabled to true ;)

Vickey-Wu:
> You need to set enabled to true ;)

Sorry, I didn't notice that...

PR #330, opened 2020-11-09
Allow rules be enabled independent of exporters

unguiculus:
Some of the rules are only enabled if the corresponding exporters are
enabled as well. This is not ideal. The exporters might already be
installed.

In our case, we use the chart to install the operator and to also
provision further Prometheus instances via operator. The exporters are
only installed with the release that installs the operator. When
additional Prometheus instances are installed they might also want to
get these rules but can't without installing the exporters again, which
doesn't make sense.

Bottom line: Let the user decide what rules they want. Don't base it on
whether or not exporters are installed as well.

Signed-off-by: Reinhard Nägele <unguiculus@gmail.com>


bismarck:
@unguiculus I'm okay with this change. However, you need to make this change [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/hack/sync_prometheus_rules.py#L52-L84) instead of the rule files.

unguiculus:
Thanks for pointing that out. I didn't think about that and will update the PR.

bismarck (approved):
lgtm

bismarck:
@torstenwalter @scottrigby I think I'm unable to merge this because of the .gitignore change?

torstenwalter (approved):
Approved the `.gitignore`

Issue #340, opened 2020-11-11
[prometheus] Double merge of global configs with no clear override method


hoffmanntravis:
**Describe the bug**
I don't see a good way to set external_labels.  This is specifically for Thanos configuration.  The most likely route to success was passing ServerFiles.prometheus.yml, but these lines of code prevents that approach from working:
https://github.com/prometheus-community/helm-charts/blob/d79879d5001144d847c914da869c887e88c3960f/charts/prometheus/templates/server/cm.yaml#L14-L16

Note that this issue is very similar to https://github.com/prometheus-community/helm-charts/issues/230, but I am specifically trying to override global values (external_labels subfield).  I also tried passing those values into extraScrapeConfigs as suggested in 230, although naturally that didn't work since that doesn't really make sense in context.  But since it was doing a suffix append to the prometheus.yml file content I gave it a shot.

**Version of Helm and Kubernetes**:
Helm Version:
I believe v3.  There are several abstractions between me and helm so I'm not 100% sure, but I can get this info if it's important.  However, this is mostly an issue with the chart and I think the other info here should be adequate.

Kubernetes Version:

```console
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.11", GitCommit:"3a3612132641768edd7f7e73d07772225817f630", GitTreeState:"clean", BuildDate:"2020-09-02T06:46:33Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**:
https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus

**Which version of the chart**:
prometheus-11.15.0

**What happened**:
I got a double merged / concatenated global config similar to issue 230, which I don't believe is valid:
```
global:
  evaluation_interval: 1m
  scrape_interval: 1m
  scrape_timeout: 10s
global:
  external_labels: 
    environment: dev
    region: cae
scrape_configs:

... <rest of config>
```

**What you expected to happen**:
[...]

PR #342, opened 2020-11-11
[prometheus] added support for external_labels to server.global section of prometheus chart.

hoffmanntravis:
Adds support for external_labels to prometheus chart.  This is especially useful for thanos as it requires these external labels.  Additional detail in issue below.
Addresses this issue:
https://github.com/prometheus-community/helm-charts/issues/340
Note that this worked in my testing, but there may be a better approach.  This is very much a _proposed_ change

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 11.16.8
+version: 11.16.9
 appVersion: 2.21.0
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -643,6 +643,9 @@ server:
     ## How frequently to evaluate rules
     ##
     evaluation_interval: 1m
+    ##Adds external_labels as a map per: https://prometheus.io/docs/prometheus/latest/configuration/configuration/
+    ##
+    external_labels: {}
   ## https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
   ##
   remoteWrite: []
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -643,7 +643,7 @@ server:
     ## How frequently to evaluate rules
     ##
     evaluation_interval: 1m
-    ##Adds external_labels as a map per: https://prometheus.io/docs/prometheus/latest/configuration/configuration/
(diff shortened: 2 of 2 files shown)


hoffmanntravis:
This is working in my limited example that I tested and deploy. The values showed up in the config as expected and Thanos picked them up. However, I'll relent that there may be a better way of doing this overall or something I have missed, so only approve accordingly. Also, this is my first PR to this repo so if I missed the mark anywhere my apologies.

Prometheus log snippet:
level=info ts=2020-11-11T06:36:34.734523631Z caller=reloader.go:222 component=reloader msg="started watching config file and recursively rule dirs for changes" cfg= out= dirs=
level=info ts=2020-11-11T06:36:35.745115859Z caller=sidecar.go:165 msg="successfully loaded prometheus external labels" **external_labels="{environment=\"dev\", region=\"cae\"}**"
level=info ts=2020-11-11T06:36:35.74516196Z caller=intrumentation.go:48 msg="changing probe status" status=ready

charts/prometheus/values.yaml
```diff
@@ -643,6 +643,9 @@ server:
     ## How frequently to evaluate rules
     ##
     evaluation_interval: 1m
+    ## Adds external_labels as a map per: https://prometheus.io/docs/prometheus/latest/configuration/configuration/
+    ##
+    external_labels: {}
```
monotek, line 648: This should be camelcase

charts/prometheus/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 11.16.8
+version: 11.16.9
```
monotek, line 3: New feature should raise minor version.

monotek (changes requested):
I alos don't see where the "external_labels" var is used in the chart?

hoffmanntravis:
@monotek 

> I alos don't see where the "external_labels" var is used in the chart?

I was thinking the line below, but after running some more tests I think I was incorrect on the premise of this PR.  I thought I needed to add the value explicitly for the helm chart to pick it up.  However, unless I am missing something in the test deploys I just ran, I think it's pulling in the value through the line below despite it not being explicitly set in values.yml.  If you can indeed override values without specifically setting them in the values.yml I believe this PR is not needed and can be abandoned.

https://github.com/prometheus-community/helm-charts/blob/4346bfad3b8d4a77ccb9cb276406bca8a846d7e5/charts/prometheus/templates/server/cm.yaml#L16

hoffmanntravis:
Closing as I don't think this is needed per comments above and linked issue

PR #348, opened 2020-11-12
[alertmanager] Add optional config reloader

arramos84:
This PR adds the option to enable a sidecar config-reloader. It's needed to allow for dynamics 
reloads of alertmanager configurations

No issue, just a small feature that can be enabled

Tested with and without configmapReloader enabled. Verified slightest change to configmap
triggers a reload and that the status page for the alertmanager config was properly updated.

#### Checklist

- [x] Chart Version bumped


charts/alertmanager/Chart.yaml
```diff
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.1.4
+version: 0.1.5
```
monotek, line 9: minor version should be raised if new feature is added.

monotek (changes requested):
Would be nice you could add an ci file testing the enabled value.

arramos84:
> Would be nice you could add an ci file testing the enabled value.

@monotek , you mean a github action workflow for alertmanager? A helm-test , or something via chart-test? Please advise/guide and will do. Thanks for taking a look!

arramos84:
> > Would be nice you could add an ci file testing the enabled value.
> 
> @monotek , you mean a github action workflow for alertmanager? A helm-test , or something via chart-test? Please advise/guide and will do. Thanks for taking a look!

@monotek @naseemkullah please advise, thanks!

naseemkullah:
Thanks @arramos84, I'm guessing @monotek  meant something in here: https://github.com/prometheus-community/helm-charts/tree/main/charts/alertmanager/templates/tests anything there should run at CI time

arramos84:
> Thanks @arramos84, I'm guessing @monotek meant something in here: https://github.com/prometheus-community/helm-charts/tree/main/charts/alertmanager/templates/tests anything there should run at CI time

@naseemkullah @monotek , wouldn't this require a change to the workflow? In order to write a test that takes into account the flag being on or off, then I'd have to write a ci test in `alertmanager/templates/tests/`, then I'd need to add a workflow and/or change `helm-charts/ct.yaml` in order for the installation to change the values that the chart is installed with? Please advise as I believe there are 0 examples in this repo of tests for flags. Unless I'm mistaken, please point me to one. Thanks again!

monotek:
I was talking about the optional usage of a chart-testing feature, which is used to test the charts

> Charts may have multiple custom values files matching the glob pattern
> '*-values.yaml' in a directory named 'ci' in the root of the chart's
> directory. The chart is installed and tested for each of these files.
> If no custom values file is present, the chart is installed and
> tested with defaults.

See: https://github.com/helm/chart-testing/blob/b3899d2ce9e5536bcbccadb64bd96636e5b605bc/ct/cmd/install.go#L45

arramos84:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Any further update will cause the issue/pull request to no longer be considered stale. Thank you for your contributions.

Putting out the testing today.

monotek (approved):

PR #375, opened 2020-11-18
[prometheus-adapter] Add missing dnsPolicy for hostNetwork

TamasSzigeti:

In order to install prometheus adapter on EKS that is using custom CNI (e.g. Calico), it needs to use hostNetwork. This setting is available but we also need to set the correct dnsPolicy, same way [it is done here for example](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L127).

  - fixes https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/268

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 2.7.2
+version: 2.8.0
 appVersion: v0.7.0
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/DirectXMan12/k8s-prometheus-adapter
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
I would prefer to have the dns policy outside of the "{{- if .Values.hostNetwork.enabled }}" and make it configurable.
Default value for dnsPolicy should be "Default".

TamasSzigeti:
> I would prefer to have the dns policy outside of the "{{- if .Values.hostNetwork.enabled }}" and make it configurable.
> Default value for dnsPolicy should be "Default".

Thank you, done.

monotek (approved):

steven-sheehy:
LGTM, can you fix the conflicts?

TamasSzigeti:
@steven-sheehy Done

steven-sheehy (approved):
Thanks

msvticket:
The default dnsPolicy in EKS is not Default, it is ClusterFirst. This change makes the DNS lookup fail for DNS-lookup of in-cluster service: lookup prometheus on XX.XX.XX.XX:53: no such host. To correctly. I just created a PR with a fix for this.

PR #390, opened 2020-11-18
[prometheus-postgres-exporter] Migrate to chart v2

zanhsieh:
Signed-off-by: zanhsieh <zanhsieh@gmail.com>

Initial attempt to move to chart v2, since helm v2 has been out of support from Nov. 15th, 2020.

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-postgres-exporter]`)


monotek (changes requested):
maybe we should add the chart type too?

see: https://helm.sh/docs/topics/charts/#chart-types

monotek (approved):

zanhsieh:
@scottrigby 
Could you help me to review this PR please?

scottrigby (changes requested):
Would you please add an "Upgrading" section, with a subheader for "2.0.0", and make crystal clear that this version requires Helm 3? I see you added to prerequisites but this would also help make clear when the change happened for users who may need to take longer upgrading from Helm 2 to 3. Thanks!

Issue #142, opened 2020-09-24
[kube-prometheus-stack] helm upgrade --install runs with errors after first time


yosefy:
**Describe the bug**

**Version of Helm and Kubernetes**:

Helm Version:

```console
$ helm version
version.BuildInfo{Version:"v3.3.1", GitCommit:"249e5215cde0c3fa72e27eb7a30e8d55c9696144", GitTreeState:"dirty", GoVersion:"go1.15"}
```

Kubernetes Version:

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.12-gke.2", GitCommit:"fb7add51f767aae42655d39972210dc1c5dbd4b3", GitTreeState:"clean", BuildDate:"2020-06-01T22:20:10Z", GoVersion:"go1.12.17b4", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**:
 prometheus-kube-stack
**Which version of the chart**:

**What happened**:
running
helm --debug --stderrthreshold 1  upgrade --install prometheus prometheus-community/kube-prometheus-stack -f values.yaml

first time runs ok but when running second time, with or without changes, we get this type of errors about rules

client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-alertmanager"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-kube-controller-manager"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-kube-etcd"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-kube-proxy"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-kube-scheduler"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-operator"
client.go:436: [debug] Looks like there are no changes for Service "prometheus-kube-prometheus-prometheus"
client.go:436: [debug] Looks like there are no changes for DaemonSet "prometheus-prometheus-node-exporter"
[...]

PR #400, opened 2020-11-20
[kube-prometheus-stack] Add support for specifying webhook port for GKE workaround

gw0:
Allows the admission webhook to work on GKE clusters without configuring firewall rules and probably also fixes https://github.com/prometheus-community/helm-charts/issues/142 if you configure `prometheusOperator.tls.internalPort=10250`.

The `prometheus-operator` webhook is using the port `8443`. In GKE private clusters, there is a firewall between Kubernetes masters and nodes and only ports `443` and `10250` are allowed by default. This is the reason of the issue #142.

A known workaround (used by [some Helm charts](https://github.com/jetstack/cert-manager/blob/d4c9a7beb4ab7c734c84f2f3a46b9613997e6172/deploy/charts/cert-manager/values.yaml#L269-L274)) is to allow specifying the webhook port and set it to `10250` which is the only non-privileged port allowed by the GKE firewall by default. This PR implements support for setting this.


monotek (changes requested):
please raise chart minor version

gw0:
Rebased and increased minor version.

vsliouniaev:
@gw0 does this not require some customisation of the service and the admission hook configurations as well?

gw0:
It should not require any other customization, because:

- Deployment exposes the `containerPort` under `name: https`(this PR)
- Service specifies under `name: https` the port with `targetPort: https`
- and both Webhooks refer to the Service (not to a port) by `namespace` and `name`

gw0:
Oh, @vsliouniaev, you are right... Because the Webhooks refer to the Service which uses `port: 443`, the Webhooks fail:

```
helm.go:75: [debug] Internal error occurred: failed calling webhook "prometheusrulemutate.monitoring.coreos.com": Post https://tmp-monitoring3-kube-prome-operator.default.svc:443/admission-prometheusrules/mutate?timeout=10s: no endpoints available for service "tmp-monitoring3-kube-prome-operator"
```

This means that this PR does not solve anything. I will update it to change the Service port instead.

gw0:
I reproduced the issue https://github.com/prometheus-community/helm-charts/issues/142#issuecomment-737063085 and confirmed that this PR when setting the webhook port to 10250 solves it on GKE private clusters. The error in my previous comment is unrelated.

I rebased the PR and set the webhook port by default to 10250 in order to work out-of-the-box in GKE private clusters and avoid adding firewall rules.

tricky42:
I would be interested in this fix. Anyway I can support here to get it updated (there have been many updates to the chart, but no conflict apart from the `Chart.yaml` as the `version` and `appVersion` have been changed in the meantime and anyway the update to the `version` was missing. 

@gw0 will you be able to do the updates, otherwise I would maybe create an update PR myself...?

gw0:
@tricky42 Rebased and increased minor version.

tricky42:
Thx @gw0. I think now everything should be good to go @monotek, or?

monotek (approved):

vsliouniaev (dismissed):
Could you please bump the chart version?

gw0:
@vsliouniaev Bumped chart version.

gkarthiks (approved):
lgtm

PR #403, opened 2020-11-20
CI: Check if CODEOWNERS reflect chart maintainers

torstenwalter:

Script generates CODEOWNERS based on chart maintainers and checks if generated file shows a diff.

  - fixes #

#### Checklist

### DIFF
--- a/.github/workflows/check-codeowners.yaml
+++ b/.github/workflows/check-codeowners.yaml
@@ -1,0 +1,17 @@
+name: Check CODEOWNERS
+
+on: pull_request
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: install yq
+        run: |
+          sudo snap install yq
+      - name: generate CODEOWNERS
+        run: |
+          ./scripts/check-codeowners.sh > .github/CODEOWNERS
+      - name: check CODEOWNERS for modifications
+        run: |
+          git diff --exit-code
--- a/.github/workflows/check-codeowners.yaml
+++ b/.github/workflows/check-codeowners.yaml
@@ -6,10 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: install yq
-        run: |
-          sudo snap install yq
       - name: generate CODEOWNERS
+        uses: mikefarah/yq@3.4.1
         run: |
           ./scripts/check-codeowners.sh > .github/CODEOWNERS
       - name: check CODEOWNERS for modifications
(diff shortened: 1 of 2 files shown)


torstenwalter:
Should be merged after #275

.github/workflows/check-codeowners.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+on: pull_request
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: install yq
+        run: |
+          sudo snap install yq
```
monotek, line 11: not sure how snap handles majo apllication updates inside same os version.
so i would suggest not to use the snap but to install yq via binary download, with a fixed version.

maybe we should also go for a fixed os version instead "ubuntu-latest".

monotek (changes requested):

monotek:
Not stale. #275 is merged so I guess ready to proceed.

.github/workflows/check-codeowners.yaml
scottrigby: Unless we need pinned versions I'd say we keep it simple. 

- For GitHub actions it's safe to use `ubuntu-latest` unless a specific version for an application is required (ref: https://github.com/actions/virtual-environments)
- snap is the preferred method for installing yq on Ubuntu (https://github.com/mikefarah/yq#ubuntu-and-other-linux-distros-supporting-snap-packages). I notice it took about 20s, which isn't too bad for this use (we could wget/curl it, but then we'd incur the maintenance cost of keeping it up to date when the version doesn't matter for our use case). I haven't tested the official GitHub action, but this open issue and comment by the yq maintainer doesn't make me feel too motivated to recommend it yet: https://github.com/mikefarah/yq/issues/418

scottrigby (approved):
Just needs a `main` merge/rebase to pass the check, otherwise LGTM

torstenwalter:
rebased changes against `main`.

.github/workflows/check-codeowners.yaml
torstenwalter: It's now using the github action to download yq, which under the hood uses snap to install it ;-)

scottrigby: Haha cool

monotek: Ok, let's keep the snap.

I already had problems using Ubuntu latest as they just moved from Ubuntu 18.04 to 20.04. problems been Python related. So I guess this could be a problem again when the image switches to Ubuntu 22.04.

monotek (commented):
See comment above

torstenwalter:
> See comment above

@monotek What should be changed?

monotek (approved):

scottrigby (approved):
👌

bismarck:
@torstenwalter I got this failure after this change was merged: https://github.com/prometheus-community/helm-charts/actions/runs/461677874

torstenwalter:
@bismarck Thanks for the ping. I am looking into it.

torstenwalter:
@bismarck should be fixed now.

PR #406, opened 2020-11-23
[prometheus-cloudwatch-exporter] Allow user-defined matchLabels for deployment and servicemonitor in prometheus-cloudwatch-exporter

paweldudzinski:

The changes will allow the end user to define custom matchLabels for deployment and serviceMonitor. This is crucial for having metrics in Prometheus Operator.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
@@ -13,14 +13,14 @@ spec:
     matchLabels:
       app: {{ template "prometheus-cloudwatch-exporter.name" . }}
       release: {{ .Release.Name }}
-{{- if .Values.deployment.matchLabels }}
-{{ toYaml .Values.deployment.matchLabels | indent 6 }}
-{{- end }}
   template:
     metadata:
       labels:
         app: {{ template "prometheus-cloudwatch-exporter.name" . }}
         release: {{ .Release.Name }}
+        {{- if .Values.pod.labels }}
+{{ toYaml .Values.pod.labels | indent 8 }}
+        {{- end }}
       annotations:
         {{ if .Values.aws.role}}iam.amazonaws.com/role: {{ .Values.aws.role }}{{ end }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
--- a/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
@@ -13,6 +13,9 @@ spec:
     matchLabels:
       app: {{ template "prometheus-cloudwatch-exporter.name" . }}
       release: {{ .Release.Name }}
+{{- if .Values.deployment.matchLabels }}
+{{ toYaml .Values.deployment.matchLabels | indent 6 }}
+{{- end }}
   template:
     metadata:
       labels:
--- a/charts/prometheus-cloudwatch-exporter/values.yaml
+++ b/charts/prometheus-cloudwatch-exporter/values.yaml
@@ -34,6 +34,9 @@ service:
   annotations: {}
   labels: {}
 
+pod:
+  labels: {}
+
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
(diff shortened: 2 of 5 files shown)


charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
```diff
@@ -13,6 +13,9 @@ spec:
     matchLabels:
       app: {{ template "prometheus-cloudwatch-exporter.name" . }}
       release: {{ .Release.Name }}
+{{- if .Values.deployment.matchLabels }}
+{{ toYaml .Values.deployment.matchLabels | indent 6 }}
```
adambabik, line 17: You can align indentation of `if` and `toYaml` where it belongs if you use `nindent`.

```suggestion
      {{ toYaml .Values.deployment.matchLabels | nindent 6 }}
```

The same comment about indentation applies to other changes in this PR.

torstenwalter:
@paweldudzinski Can you elaborate why you need to specify these. I don't understand the need for it.

torstenwalter:
Isn't a `ServiceMonitor` used to tell prometheus operator which metric to scrape?
https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md#related-resources

paweldudzinski:
> Isn't a `ServiceMonitor` used to tell prometheus operator which metric to scrape?
> https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md#related-resources

@torstenwalter The ServiceMonitor has a label selector to select Services and their underlying Endpoint objects. The labels have to match. Some setups might have special labels that are required. For example, in my specific case the existing setup of predefined labels is not enough to fully match the resources.

torstenwalter:
I don;t see why you need to adjust selectors for the deployent and the service objects. It's an immutable fiels in Kubernetes. Having it configurable via values.yaml might lead to the impression that people can change it at any point in time. That's not the case!

paweldudzinski:
Let me describe our case. There are several way to configure Prometheus Operator so it scrapes metrics. Our PO bases on label called `owner`, other configuration may base on different labels. In order for it to work we need to setup such label in deployment and then match those labels. Therefore the need of extra labels. Just like in a service: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-cloudwatch-exporter/templates/service.yaml#L12 it can be applied to any other resource.

torstenwalter:
For that it should be sufficient to introduce additional labels for the pod like:
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-blackbox-exporter/templates/deployment.yaml#L26-L28

There is no need to use them as `matchLabels` in deployment or `selector` in service.

paweldudzinski:
ok, I missed that possibility to customize the pod labels - thanks for that. That changes a lot. 
Yet, for my case I'd still need to do the same thing (add labels) for deployment.
PR is updated just to this use case. @torstenwalter do you think I'd be possible?

charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     chart: {{ template "prometheus-cloudwatch-exporter.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+{{- if .Values.deployment.labels }}
+  {{ toYaml .Values.deployment.labels | indent 4 }}
+{{- end }}
```
torstenwalter, line 12: I you move these down to `template.metadata.labels` and name them `pod.labels` then it's fine for me.

charts/prometheus-cloudwatch-exporter/values.yaml
```diff
@@ -34,6 +34,9 @@ service:
   annotations: {}
   labels: {}
 
+deployment:
+  labels: {}
```
torstenwalter, line 38: I suggest to either use:
```suggestion
deployment:
  podLabels: {}
```
or
```suggestion
pod:
  labels: {}
```

torstenwalter (changes requested):

paweldudzinski:
@torstenwalter all done (same style as prometheus-blackbox). thanks for guidance.

torstenwalter (approved):
LGTM

torstenwalter:
@paweldudzinski Thank you for the contribution and the discussion about it.

PR #411, opened 2020-11-25
[prometheus] Potentially change default liveness and readiness to abide by GKE limitations

### LINKED ISSUE #410: [prometheus] Default liveness & Readiness probes have scrape timeouts longer than scrape period
**Describe the bug**

The default liveness and readiness probes have scrape timeouts longer than the scrape period. This causes the GCP Ingress Controller to fail to bring up a HTTP Load Balancer with the error: "Error during sync: error running backend syncing routine: error ensuring health check: googleapi: Error 400: Invalid value for field 'resource.timeoutSec': '30'. TimeoutSec should be less than checkIntervalSec., invalid"

Helm Version:

```console
version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}
```
Kubernetes Version:

```console
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.12-gke.1504", GitCommit:"17061f5bd4ee34f72c9281d49f94b4f3ac31ac25", GitTreeState:"clean", BuildDate:"2020-10-19T17:00:22Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**:
The deprecated prometheus chart

An old version of the deprecated prometheus chart. But I can see the root cause in the chart on master

**What happened**:
I deployed the helm chart, and my Ingress which pointed at the helm chart. Rather than the ingress controller setting up the appropriate resources, it kicked out the error: "Error during sync: error running backend syncing routine: error ensuring health check: googleapi: Error 400: Invalid value for field 'resource.timeoutSec': '30'. TimeoutSec should be less than checkIntervalSec., invalid"

Using v1.17.12-gke.1504 with an ingress try to expose the Prometheus helm chart with the service type of NodePort.

values.yaml
```console
server.service.type: NodePort
```

**Workaround**

To work-around, configure a readinessProbeTimeout and livenessProbeTimeout of less than the probe period.

```console
  server:
    readinessProbePeriodSeconds: 15
    readinessProbeTimeout: 10
    livenessProbePeriodSeconds: 15
    livenessProbeTimeout: 10
```

strideynet:
Changes the default configuration so that the liveness and readiness probes have timeouts shorter than the period. This addresses issue #410 which means that the chart does not currently play nicely with GKE HTTP Ingresses when configured in NodePort mode.

It does continue to make sense to me that the timeout for these things should be less than the period (Indeed, this is how Prometheus itself treats scraping). I am not sure that 4 seconds is the correct timeout for us to be using here, but 30 seconds does seem absurd when both the health and liveness endpoints have been designed to respond relatively quickly. I thought opening this PR would at least give us a chance to discuss this, or decide to just improve the documentation around deploying onto GKE rather than changing these default values. 

Fixes #410 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -897,7 +897,7 @@ server:
   readinessProbeSuccessThreshold: 1
   livenessProbeInitialDelay: 30
   livenessProbePeriodSeconds: 15
-  livenessProbeTimeout: 4
+  livenessProbeTimeout: 10
   livenessProbeFailureThreshold: 3
   livenessProbeSuccessThreshold: 1
 
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -892,12 +892,12 @@ server:
   ##
   readinessProbeInitialDelay: 30
   readinessProbePeriodSeconds: 5
-  readinessProbeTimeout: 30
+  readinessProbeTimeout: 4
   readinessProbeFailureThreshold: 3
   readinessProbeSuccessThreshold: 1
   livenessProbeInitialDelay: 30
   livenessProbePeriodSeconds: 15
-  livenessProbeTimeout: 30
+  livenessProbeTimeout: 4
   livenessProbeFailureThreshold: 3
   livenessProbeSuccessThreshold: 1
 
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: prometheus
-version: 12.0.0
+version: 12.0.1
 appVersion: 2.21.0
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/


Xtigyro (changes requested):
Hey @strideynet 👋 Thanks for the contribution - it's appreciated!

Could you please:
- bump the chart version in `Chart.yalm`;
- check whether when GKE is used - anything from the built-in objects can be used to catch that - https://helm.sh/docs/chart_template_guide/builtin_objects/?

strideynet:
> Hey @strideynet 👋 Thanks for the contribution - it's appreciated!
> 
> Could you please:
> 
> * bump the chart version in `Chart.yalm`;
> * check whether when GKE is used - anything from the built-in objects can be used to catch that - https://helm.sh/docs/chart_template_guide/builtin_objects/?

I'm not really sure that this is *specifically* a GKE issue, it's possible there are other systems out there that may be expecting the period of the check to be longer than the timeout. Surely we should establish if it would be sensible to just change these default values for everyone? If not, then I will proceed with making it GKE only.

Xtigyro:
@strideynet Yeah - that will do as well.

charts/prometheus/values.yaml
```diff
@@ -892,12 +892,12 @@ server:
...
   readinessProbeInitialDelay: 30
   readinessProbePeriodSeconds: 5
-  readinessProbeTimeout: 30
+  readinessProbeTimeout: 4
   readinessProbeFailureThreshold: 3
   readinessProbeSuccessThreshold: 1
   livenessProbeInitialDelay: 30
   livenessProbePeriodSeconds: 15
-  livenessProbeTimeout: 30
+  livenessProbeTimeout: 4
```
monotek, line 900: as livenessProbePeriodSeconds is 15 here i would got to 10 seconds for the livenessProbeTimeout.

monotek (changes requested):
please raise chart version

charts/prometheus/values.yaml
strideynet: @monotek One thing I am considering is that surely the timeout threshold should remain the same for both probes? Given that the internal Prometheus implementation for both is very similar and both respond very quickly, it seems that 10 seconds is almost certainly overkill

monotek: The liveness probe would restart the pod when its failing so imho is reasonable to give it some more time.

strideynet: That seems like a reasonable justification. I've incorporated your suggest change.

zanhsieh (approved):

Xtigyro (approved):

monotek (approved):

PR #413, opened 2020-11-25
[kube-prometheus-stack] Configure possibility of providing additional rules to prometheus cluster role

dieterdemeyer:

This change will fix the Unhealthy Target Group `kubernetes-nodes` which will scrape node metrics from the `/proxy` endpoint.

#### Checklist
- [x] Chart Version bumped


monotek (approved):

bismarck (changes requested):
@dieterdemeyer There's no `kubernetes-nodes` job in the kube-prometheus-stack chart. I see one in the [prometheus chart](https://github.com/prometheus-community/helm-charts/blob/4f4574545ccd5549bd67ef495b8540ad40e3310c/charts/prometheus/values.yaml#L1281-L1315)?

dieterdemeyer:
@bismarck ,

Yes, you are correct.

We actually use custom `values.yaml` file which specifies a number of additional jobs and `kubernetes-nodes` is one of them.
We currently use the `prometheus-operator` helm chart and the additional proxy endpoint is listed in the cluster role defined there but is missing in the `kube-prometheus-stack` helm chart.

vsliouniaev:
I'm against changing the default RBAC scope to support a custom configuration. The chart allows a different role to be used that is not provisioned by the chart, so that is probably your best option without any modifications.

If you do want to extend functionality here to support additions to the RBAC scope I think that would be a reasonable extension.

dieterdemeyer:
@vsliouniaev , I've updated the PR. Is this more in line of what you meant as an extension ?

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.2.3
+version: 12.2.4
```
monotek, line 21: this is a new feature now and should raise minor version.

monotek (changes requested):

charts/kube-prometheus-stack/Chart.yaml
dieterdemeyer: Version has been bumped to 12.3.1.

monotek (changes requested):
version should be 12.5.0

dieterdemeyer:
@monotek , version has been bumped.

monotek (approved):

charts/kube-prometheus-stack/templates/prometheus/clusterrole.yaml
```diff
@@ -24,4 +24,7 @@ rules:
   verbs: ["get", "list", "watch"]
 - nonResourceURLs: ["/metrics", "/metrics/cadvisor"]
   verbs: ["get"]
+{{- if and .Values.prometheus.enabled .Values.prometheus.additionalRulesForClusterRole }}
```
bismarck, line 27: `.Values.prometheus.enabled` is checked at the top of the file, so it isn't needed here.

bismarck (changes requested):

charts/kube-prometheus-stack/templates/prometheus/clusterrole.yaml
dieterdemeyer: Unnecessary check has been removed.

bismarck (approved):
LGTM

PR #440, opened 2020-12-02
Prometheus - GH-417 Allow merging server sidecarContainers

### LINKED ISSUE #417: Allow Helm "list-merge" for sidecar containers
A limitation with Helm is that it cannot merge lists (arrays) when using more than one values file (https://github.com/helm/helm/issues/3486); One workaround is having an object (dictionary) instead of the list, and iterating over the object. This allows to add/remove/modify internal properties of the objects when knowing their name.

In particular, we have a definition for a `sidecarContainers` that adds the `thanos-sidecar`; But we want to set different `resources.requests` for each of our different environments, without copying the whole sidecar container definition to each and every environment values file.

The existing use of `sidecarContainers` in the Deployment and StatefulSet of Prometheus server looks like this:
https://github.com/prometheus-community/helm-charts/blob/9abd1cfc61b9a3828c9d6909bf1fc7f755f0536d/charts/prometheus/templates/server/sts.yaml#L154-L156
https://github.com/prometheus-community/helm-charts/blob/9abd1cfc61b9a3828c9d6909bf1fc7f755f0536d/charts/prometheus/templates/server/deploy.yaml#L149-L151

It would be much more useful if changed to look like this:

```yaml
{{- if .Values.server.sidecarContainers }}
  {{- range $name, $spec :=  .Values.server.sidecarContainers }}
    - name: {{ $name }}
      {{- toYaml $spec | nindent 10 }}
  {{- end }}
{{- end }}
```

The above change will allow naming and defining a `sidecarContainer` in a single `values.yaml` file, then having an override for certain container properties (such as image tag, resources.requests/limits, etc...) in small per-environment files.

A shared `values-shared.yaml` can include:
```yaml
server:
  sidecarContainers:
    thanos-sidecar:
      image: quay.io/thanos/thanos:v0.16.0
      resources:
        requests:
          memory: 64Mi
```

And another `values-staging-env.yaml` can include only the override, like so:
```yaml
server:
  sidecarContainers:
    thanos-sidecar:
      resources:
        requests:
          memory: 128Mi
```

And all would work when using `helm install -f values-shared.yaml -f values-staging-env.yaml`

kesor:
Explained in https://github.com/prometheus-community/helm-charts/issues/417

- fixes #417 

#### Checklist
- [x] Bump chart version


zanhsieh:
@kesor 
Could you bump the Chart version to latest please? Thank you.

zanhsieh (approved):

monotek (changes requested):
i guess this breaks it for all users using the sidcear option now.
so a bugfix version bump is imho not enough.

naseemkullah:
Would post rendering work to add the sidecars? https://helm.sh/docs/topics/advanced/#post-rendering

kesor:
> i guess this breaks it for all users using the sidcear option now.
> so a bugfix version bump is imho not enough.

Yes. Would you like me to change the version bump?

Also, how bad is it really? It breaks the chart in such a way that anyone who is using sidecar containers will not be able to use the chart with their previous values file and will have to fix it. So it is not possible to deploy a new version of the chart with a broken configuration, it has to be fixed by the person doing it.

monotek:
Yes and this would need a major version bump.

monotek (approved):

PR #445, opened 2020-12-03
Changing default container port in prometheus-blackbox-exporter

rzjfr:
Imagine we want to change the service port from 9115 to another value.
Because the `containerPort` in `deployment.yaml` is referring to service port, if we want to change the service port in `Values.yaml`, app will still listen in 9115 and container in the newly specified port and the probes will fail. That's why I think container port meant to listen to the `targetPort` in the `Values.yaml`.

#### Comment to the maintainers
Though I cannot see the reason why someone would like to do that, but the same issue will arise if someone changes the targetPort in `Values.yaml`. We should either change app listening port to be set to the targetPort [(here)](https://github.com/prometheus/blackbox_exporter/blob/57b2925620b3ea2d3bb20899e16a415c2d5dd95b/main.go#L55) or just remove the the line from `Values.yaml`. I didn't change that because it is rare scenario and also it also depends on how maintainers wanted to proceed.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 4.10.1
+version: 4.10.2
 appVersion: 0.18.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
--- a/charts/prometheus-blackbox-exporter/templates/deployment.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/deployment.yaml
@@ -90,7 +90,7 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           ports:
-            - containerPort: {{ .Values.service.targetPort }}
+            - containerPort: {{ .Values.containerPort }}
               name: http
           livenessProbe:
             {{- toYaml .Values.livenessProbe | trim | nindent 12 }}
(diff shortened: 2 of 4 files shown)


desaintmartin:
Thanks ! Indeed, there is a bug here.

1/ In fact, `targetPort: http` would still be valid from a Kube point of view but would break your solution, so either we forbid this, or we introduce a new variable
2/ Changing this will not change the configuration, still listening to the default port

so, just an idea, maybe we could:
- Change the default `service.targetPort` to http so that we don't have any problem changing actual service port number
- Optionally, later, change the container port number according to the configuration (but there would introduce maybe too much intelligence into this poor helm template)

What do you think?

rzjfr:
Thank you for reviewing my PR.

The point about targetPort in `service.yaml` is something I missed and should be corrected for the current solution to work.

Also thanks for the ideas, to summarize we can do this:

1. Hardcode targetPort to http in `service.yaml`, and I'd say remove `service.targetPort` and put it under new name of `containerPort` since it's no longer related to service and targetPort is hard coded anyway. Then extraArgs should be set alongside the new variable `containerPort` if someone wanted to change the listening port of the app.

For the second point you've made, I agree, we are adding too little value compared to the number of codes and the complexity we increase by doing so. But if we want to do that we can:

2. Add a logic in `deployment.yaml` to set an extraArgs based on a new vaialbe like listeningPort ("--web.listen-address=" + listeningPort) if it is set. Downside is if the person ignores the listeningPort and tries changing using extraArgs with "--web.listen-address" then there will she will face an error.
or

3. Add a helper function to extract the container port from "--web.listen-address" if was set in the extraArgs or set it to default otherwise. In both cases "service.targetPort" should go anyway. Downside is that new logic (more than a couple of lines) should be maintained as well.


I personally vote for item 1.

desaintmartin:
I vote for 1 as well! This would make it a breaking change.

monotek (approved):

PR #454, opened 2020-12-04
[alertmanager] Allow passing notification template

lekspyl:

This PR adds ability to specify [notification template](https://prometheus.io/docs/alerting/latest/notifications/#notification-template-reference) that will be used by Alertmanager.

None known.

#### Checklist

- [x] Chart Version bumped


charts/alertmanager/values.yaml
```diff
@@ -101,3 +104,5 @@ config:
     group_interval: 5m
     receiver: default-receiver
     repeat_interval: 3h
+
+template: |-
```
monotek, line 108: imho it would be better to do this like:

```
templates:
  - name: "/etc/alertmanager/alertmanager.tmpl"
    content: ""
```

Otherwise its unclear, that these options belong together.

monotek (changes requested):

charts/alertmanager/values.yaml
naseemkullah: would `name: alertmanager.tmpl` suffice? it would go in same dir as other config i imagine (/etc/alertmanager)

monotek: yes :)

lekspyl: I'm afraid I don't get your point. We are dumping `config` as a whole to the `alertmanager.yml` file, so altering `templates` config key structure will not be accepted by Alertmanager as it [expects](https://prometheus.io/docs/alerting/latest/configuration/#filepath) to see simple list of files, not list of maps:

Screenshot 2020-12-08 at 17 52 32

charts/alertmanager/templates/configmap.yaml
```diff
@@ -8,4 +8,8 @@ metadata:
 data:
   alertmanager.yml: |
     {{- toYaml .Values.config | default "{}" | nindent 4 }}
-{{- end }}
\ No newline at end of file
+  {{- if .Values.config.templates }}
+  alertmanager.tmpl: |-
+    {{- .Values.template | nindent 4 }}
+  {{- end }}
+{{- end }}
```
naseemkullah, line 15:
```suggestion
  {{- range $key, $value := .Values.templates }}
  {{ $key }}: |-
    {{- .$value | nindent 4 }}
  {{- end }}
{{- end }}
```

naseemkullah (changes requested):
Thanks for pointing that out @lekspyl 

I've added some suggested code changes (they might be slightly off but the idea is there).

let's allow an abitrary amount of additional templated, by leverage the wildcard feature in alertmanager, and the range method in go templating. WDYT?

charts/alertmanager/values.yaml
```diff
@@ -101,3 +104,6 @@ config:
     group_interval: 5m
     receiver: default-receiver
     repeat_interval: 3h
+
+# templates:
+#   alertmanager.tmpl: |-
```
naseemkullah, line 109:
```suggestion
templates: {}
#   alertmanager.tmpl: |-
```

optional nit, should have suggested it in the first place sorry

naseemkullah (dismissed):
lgtm, left an optional nit

naseemkullah:
@lekspyl did you get a chance to test the templating functionality as-is locally?

naseemkullah (approved):

lekspyl:
@naseemkullah yes, it's renderable both via undefined `templates` and with it. Thanks for the help!

naseemkullah:
> @naseemkullah yes, it's renderable both via undefined `templates` and with it. Thanks for the help!

Great, thanks for the great feature! Seems like there might be some unsigned commits according to failed DCO check, after that is sorted out and @monotek takes a look we should be getting that in!

lekspyl:
Aw snap. I'll better reopen a new PR.

naseemkullah:
> Aw snap. I'll better reopen a new PR.

sure, np

lekspyl:
Closed in favor of https://github.com/prometheus-community/helm-charts/pull/484

PR #457, opened 2020-12-07
[kube-prometheus-stack] Align Grafana and Prometheus scrape intervals

### LINKED ISSUE #394: [kube-prometheus-stack] Scrape interval misconfigured in Grafana's default data source
**Describe the bug**

The [`grafana-datasource` ConfigMap](https://github.com/prometheus-community/helm-charts/blob/b643f58714a063e5a0ddf6832cd00ed29c0f4fab/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml) doesn't define any scrape interval for Prometheus, which means the [default value of `15s`](https://grafana.com/docs/grafana/latest/datasources/prometheus/) is applied.

However, Prometheus' global scrape interval is `30s`:

```console
$ kubectl -n monitoring exec sts/prometheus-prometheus-stack-kube-prom-prometheus -c prometheus -- cat /etc/prometheus/config_out/prometheus.env.yaml
global:
  evaluation_interval: 30s
  scrape_interval: 30s
...
```

This makes the results of `rate()` functions used together with `[$__interval]` imprecise, or sometimes just plain wrong.

**Proposal**

I believe it would be sensible to default this value to `30s` if the user didn't explicitly define a scrape interval in the [chart's values](https://github.com/prometheus-community/helm-charts/blob/52dfc37b648001202edd9179effaec6aa188d887/charts/kube-prometheus-stack/values.yaml#L1659-L1670).

Here's how this can be achieved: https://github.com/grafana/grafana/issues/12656#issuecomment-435307645

**Other references**

See this blog post for a reference about why defining the scrape interval correctly is important: https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/

_Helm and Kubernetes versions are irrelevant here._

antoineco:

Full description in the issue referenced below

*tl;dr:* it is important to ensure the scrape interval defined in the Grafana data source matches the one configured by the Prometheus operator. Right now, default values do not match ([`15s`](https://grafana.com/docs/grafana/latest/datasources/prometheus/) vs. [`30s`](https://github.com/prometheus-operator/prometheus-operator/blob/release-0.44/pkg/prometheus/promcfg.go#L180-L183)).

Fixes #394

I'm bumping the _minor_ part of the chart's semver because the change does have an effect on the current default.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.5.0
+version: 12.6.0
 appVersion: 0.44.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
+++ b/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
@@ -22,6 +22,8 @@ data:
       url: http://{{ template "kube-prometheus-stack.fullname" . }}-prometheus:{{ .Values.prometheus.service.port }}/{{ trimPrefix "/" .Values.prometheus.prometheusSpec.routePrefix }}
       access: proxy
       isDefault: true
+      jsonData:
+        timeInterval: {{ .Values.prometheus.prometheusSpec.scrapeInterval | default "30s" }}
 {{- if .Values.grafana.sidecar.datasources.createPrometheusReplicasDatasources }}
 {{- range until (int .Values.prometheus.prometheusSpec.replicas) }}
     - name: Prometheus-{{ . }}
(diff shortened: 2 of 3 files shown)


antoineco:
Just to prove the change works as intended:

image

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.5.0
+version: 12.6.0
```
gkarthiks, line 21: lgtm, but would it be an increment to minor or patch?

antoineco: I wasn't sure 100% myself. As mentioned in the PR description, I though _minor_ was appropriate here, what do you think?

> I'm bumping the _minor_ part of the chart's semver because the change does have an effect on the current's default.

bismarck (approved):
LGTM. This was a good catch.

PR #490, opened 2020-12-14
[kube-prometheus-stack] Use https for kube-scheduler and kube-controller-manager metrics

aladmit:
Since k8s 1.18.x uses https ports and disable http port for Kube-scheduler, Kube-controller manager. So we need to update service monitor settings to work with new ports.

prometheus-operator already did that: https://github.com/prometheus-operator/kube-prometheus/pull/621/files

Without these changes alerts KubeControllerManagerDown and kubeSchedulerDown are firing.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -863,8 +863,8 @@ kubeControllerManager:
   ## If using kubeControllerManager.endpoints only the port and targetPort are used
   ##
   service:
-    port: 10252
-    targetPort: 10252
+    port: 10257
+    targetPort: 10257
     # selector:
     #   component: kube-controller-manager
 
@@ -876,10 +876,10 @@ kubeControllerManager:
     ## Enable scraping kube-controller-manager over https.
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
     ##
-    https: false
+    https: true
 
     # Skip TLS certificate validation when scraping
-    insecureSkipVerify: null
+    insecureSkipVerify: true
 
     # Name of the server to use when validating TLS certificate
     serverName: null
@@ -1056,8 +1056,8 @@ kubeScheduler:
   ## If using kubeScheduler.endpoints only the port and targetPort are used
   ##
   service:
-    port: 10251
-    targetPort: 10251
+    port: 10259
+    targetPort: 10259
     # selector:
     #   component: kube-scheduler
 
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -876,10 +876,10 @@ kubeControllerManager:
     ## Enable scraping kube-controller-manager over https.
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
     ##
-    https: false
+    https: true
 
     # Skip TLS certificate validation when scraping
-    insecureSkipVerify: null
+    insecureSkipVerify: true
```
monotek, line 882: not sure if this should be true by default?
is there no option to add the root ca cert so the request will work without disabling verification complelty?

aladmit: Unfortunately, I didn't find a way to do that.

Root ca already in prometheus container and we can find it at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt And the chart already use it for https connections: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/exporters/kube-scheduler/servicemonitor.yaml#L28

If I set `insecureSkipVerify: false`, I get this error:
```
Get "https://10.234.3.101:10259/metrics": x509: certificate is valid for 127.0.0.1, not 10.234.3.101
```

10.23.3.101 -- a master node with Kube-scheduler.

If I set `tlsConfig.serverName` to 127.0.0.1 to fix this, I get another one:
```
Get "https://10.234.3.101:10259/metrics": x509: certificate signed by unknown authority
```

So, as I understand, kube-scheduler uses it's own certificate which doesn't signed with root cluster ca.

aladmit: May be that's the reason why insecure enabled in kube-prometheus chart too: https://github.com/prometheus-operator/kube-prometheus/blob/master/manifests/prometheus-serviceMonitorKubeScheduler.yaml#L15

aladmit: Oh, I guess they are just don't know too :)) https://github.com/prometheus-operator/kube-prometheus/issues/469

monotek: Not sur if it works. 
Just remembered that i've used it like "curl --cacert ca.crt..." to speak to the kube api on port 6443 without cert error.

vsliouniaev:
Overall makes sense but if the default compatibility with other k8s versions is getting broken we should increase the major version and describe the migration steps.

PR #505, opened 2020-12-18
[kube-prometheus-stack] update and document grafana default admin creds

### LINKED ISSUE #506: [kube-prometheus-stack] generate a random grafana admin password by default

By default, all grafana instances generated by the kube-prometheus-stack chart will use `prom-operator` as the grafana admin password. If users expose the grafana UI publically, the interface can easily be entered, exploited, hacked etc.

It should generate a random password by default. The grafana helm chart already supports this, we just have to update the default password to be an empty string in the values.yml

**Additional context**

I have proposed the changes in #505

adamkgray:
notes:
- explicitly expose the grafana adminUser value
- generate a psuedo-random grafana admin password by default

Signed-off-by: adamkgray <adamkgray@pm.me>

  - fixes #506 

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -580,7 +580,15 @@ grafana:
   ##
   defaultDashboardsEnabled: true
 
-  adminPassword: prom-operator
+  ## Set the default admin username
+  #
+  adminUser: admin
+
+  ## Set the default admin password
+  ## An empty string generates a psuedo-random password
+  ## Otherwise, this value is used
+  #
+  adminPassword: ""
 
   ingress:
     ## If true, Grafana Ingress will be created


charts/kube-prometheus-stack/values.yaml
```diff
@@ -580,7 +580,15 @@ grafana:
...
-  adminPassword: prom-operator
+  ## Set the default admin username
+  #
+  adminUser: admin
+
+  ## Set the default admin password
+  ## An empty string generates a psuedo-random password
+  ## Otherwise, this value is used
+  #
+  adminPassword: ""
```
monotek, line 591: imho thats a breaking change, even if nobody should use a static password here.
not sure if it should be a bugfix or major breaking change?

please also raise chart version.

monotek (changes requested):
.

scottrigby:
Please see this comment https://github.com/prometheus-community/helm-charts/issues/506#issuecomment-748295501

monotek (changes requested):
Please rebase master / update version.

vsliouniaev:
I don't think it's a good idea to change the defaults like this. They are currently inheriting defaults from the upstream grafana chart, so if it doesn't make sense to do there I don't think it makes sense to do here

PR #524, opened 2020-12-28
[kube-prometheus-stack] Allow default data source name change

### LINKED ISSUE #523: [kube-prometheus-stack] Allow default datasource name change
Not necessarily a problem, but purely a feature request.

I would like to be able to override the default Prometheus data source name. If this chart is being used in a federated manner and there are multiple Prometheus sources, being able to change the name of the default data source to conform with the other, named data sources would be helpful.

None

aaronkjones:
Allows the default Prometheus data source name to be overridden from "Prometheus". If this chart is being used in a federated manner and there are multiple Prometheus sources, being able to change the name of the default data source to conform with the other, named data sources would be helpful.

  - fixes #523 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.9.1
+version: 12.9.0
 appVersion: 0.44.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.8.1
+version: 12.9.1
 appVersion: 0.44.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.8.1
+version: 12.9.1
```
monotek, line 21: I guess this should be 12.9.0?

charts/kube-prometheus-stack/values.yaml
```diff
@@ -626,7 +626,9 @@ grafana:
       multicluster: false
     datasources:
       enabled: true
-      defaultDatasourceEnabled: true
```
monotek, line 629: I don't think it's a good to change the variable name as this would mean the feature could break for users wich copied the whole values file.

To prevent a therefore major version bump of the chart i would keep it the way it is.

monotek (changes requested):

monotek (changes requested):
See comments above...

charts/kube-prometheus-stack/values.yaml
aaronkjones: Do you have another proposal?
`defaultDatasourceName: "prometheus-sample"`?

If this is a decent request, then perhaps it should follow the format used throughout the values file, even if it were to require a major version change.

monotek: This is already a bit mixed up in the chart so i just would avoid to change things, which could break stuff.
In the end, i'm not a nmaintainer of the chart so they have to approve or not anyway.

monotek (changes requested):

charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
```diff
@@ -16,8 +16,8 @@ data:
   datasource.yaml: |-
     apiVersion: 1
     datasources:
-{{- if .Values.grafana.sidecar.datasources.defaultDatasourceEnabled }}
-    - name: Prometheus
+{{- if .Values.grafana.sidecar.datasources.defaultDatasource.enabled }}
```
vsliouniaev, line 19: This makes sense but bump the major version and describe the changes in the readme

vsliouniaev (changes requested):

Issue #241, opened 2020-10-21
[prometheus] misconfiguration in an IPv6-only environment


inistor:
**Describe the bug**
In an IPv6 only environment (where $(POD_IP) is an IPv6 address), some components are incorrectly configured:
```
MacBook-Pro:prometheus user$ grep "(POD_IP)" -R  *
templates/alertmanager/deploy.yaml:            - --cluster.advertise-address=$(POD_IP):6783
templates/alertmanager/sts.yaml:            - --cluster.advertise-address=$(POD_IP):6783
```

This results in the combination of address port being unparsable - in IPv6, the correct syntax is `[IPv6 Address]:port`.

**Version of Helm and Kubernetes**:

Helm Version:

```console
$ helm version
version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}
```

Kubernetes Version:

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:23:04Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**:
prometheus

**Which version of the chart**:
11.16.3

**What happened**:
In an IPv6 only environment (where $(POD_IP) is an IPv6 address), some components are incorrectly configured.

**What you expected to happen**:
The chart should work in an IPv6 only environment.

**How to reproduce it** (as minimally and precisely as possible):
Use an IPv6 only Kubernetes cluster.
Deploy the prometheus 11.16.3 chart.

localconfig.yaml (only put values which differ from the defaults)

```
alertmanager:
  persistentVolume:
    size: 1Gi
    storageClass: "ion-sc"
server:
  persistentVolume:
    size: 1Gi
    storageClass: "ion-sc"
```

```
helm install prom-instance prometheus-community/prometheus  -f localconfig.yaml
```

[...]

PR #528, opened 2020-12-30
Allow alertmanager to be deployed in IPv6 only environments 

inistor:
This PR is to address #241. In this stage, it only aims to allow the deployment to proceed properly.

It may be needed to also modify the `cluster.listen-address` parameter (it's now set to 0.0.0.0).

I've simply added brackets to the $(POD_IP) parameter; the alertmanager code seems to properly handle this, so using brackets around the IP address is supported by alertmanager for the "cluster.advertise-address" command line option for both ipv4 and ipv6 so this should not affect existing IPv4 functionality.
See: https://github.com/prometheus/alertmanager/blob/ce108378d4c8d580804452dcc3a31222067793b2/cluster/cluster.go#L135 - this uses net.SplitHostPort() which handles, as said above, brackets around the host.

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: prometheus
-version: 13.0.2
+version: 13.0.3
 appVersion: 2.22.1
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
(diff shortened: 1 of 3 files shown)


zanhsieh (approved):

Frankkkkk:
Couldn't we, at the same time, change the listen address to `[::]` ? It will also work on v4-only clusters.
Cheers

monotek (approved):

PR #535, opened 2021-01-04
[kube-prometheus-stack] Fix an errant dash in prometheus-operator's args

blampe:

When templating with

`version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"dirty", GoVersion:"go1.15.4"}`

a stray dash is inserted which causes the following error: 

```
Service \"prom-kubelet- --logtostderr=true\" is invalid: metadata.name: Invalid value: \"prom-kubelet- --logtostderr=true\": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start w │
│ ith an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')"
```

This is because `-}}` [removes all trailing whitespace](https://golang.org/pkg/text/template/#hdr-Text_and_spaces), so the args get lumped together.

This PR resolves the error. Tested and verified locally.

Before:
```
args:
  - --kubelet-service=kube-system/prom-kubelet- --logtostderr=true
```

After:
```
args:
  - --kubelet-service=kube-system/prom-kubelet
  - --logtostderr=true
```

cc @bismarck @scottrigby

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 12.9.2
+version: 12.9.3
 appVersion: 0.44.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


bismarck (approved):
lgtm

PR #565, opened 2021-01-12
[prometheus-cloudwatch-exporter] - Added option to enable sts regional endpoints

lewisstrong1991:

This PR fixes the issue that when using STS to assume a role the global endpoint is used by default. If this is not accessible and a regional endpoint is configured the launched pods/containers will get in a crash loop. By setting the ENV variable `AWS_STS_REGIONAL_ENDPOINTS=regional` this tells the SDK to use the regional endpoint instead of the global endpoint.

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-cloudwatch-exporter/Chart.yaml
+++ b/charts/prometheus-cloudwatch-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 description: A Helm chart for prometheus cloudwatch-exporter
 name: prometheus-cloudwatch-exporter
-version: 0.12.1
+version: 0.12.2
 home: https://github.com/prometheus/cloudwatch_exporter
 sources:
 - https://github.com/prometheus/cloudwatch_exporter
--- a/charts/prometheus-cloudwatch-exporter/Chart.yaml
+++ b/charts/prometheus-cloudwatch-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 description: A Helm chart for prometheus cloudwatch-exporter
 name: prometheus-cloudwatch-exporter
-version: 0.12.2
+version: 0.13.0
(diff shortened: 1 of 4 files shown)


lewisstrong1991:
@asherf @gianrubio @torstenwalter - Any chance someone can please take a look at this?

charts/prometheus-cloudwatch-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 description: A Helm chart for prometheus cloudwatch-exporter
 name: prometheus-cloudwatch-exporter
-version: 0.12.1
+version: 0.12.2
```
torstenwalter, line 5: Can you change it to 0.13.0. As it's a new feature it's worth incrementing the minor version.

torstenwalter (changes requested):
LGTM - just a small change. Just ping we afterwards and I'll merge it.

charts/prometheus-cloudwatch-exporter/Chart.yaml
lewisstrong1991: No worries! Bumped and also fixed the version in the `CHANGELOG.md` as realised i'd mistyped that.

lewisstrong1991:
@torstenwalter - Fixed and all tests passed

torstenwalter:
Don't we have a duplicate `env` section with this in case the endpoint and aws secret are defined?

lewisstrong1991:
They should be mutually exclusive. STS would only be used when specifying an AWS Role and not used where AWS credentials are used. Could look to code this in an IF-ELSE statement?

torstenwalter:
Would make sense if they are exclusive.

lewisstrong1991:
@torstenwalter - Made a change and tested this locally. Will only be applied when a AWS role is set in the values.

torstenwalter (approved):

torstenwalter:
@lewisstrong1991 thanks for implementing it.

PR #584, opened 2021-01-14
[WIP][kube-stack-prometheus] Use kube-prometheus main branch to sync from

mrueg:
I believe this was frozen to release-0.1 by mistake and we should use master (or match a release of kube-prometheus for it)

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/hack/sync_prometheus_rules.py
+++ b/charts/kube-prometheus-stack/hack/sync_prometheus_rules.py
@@ -35,7 +35,7 @@ charts = [
         'min_kubernetes': '1.14.0-0'
     },
     {
-        'source': 'https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/release-0.1/manifests/prometheus-rules.yaml',
+        'source': 'https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/master/manifests/prometheus-rules.yaml',
         'destination': '../templates/prometheus/rules',
         'min_kubernetes': '1.10.0-0',
         'max_kubernetes': '1.14.0-0'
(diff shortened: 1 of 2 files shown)


mrueg:
Ah nevermind, after reading twice I think this is only for dashboards that have k8s 1.14 or earlier. Which makes me think, should we remove those since prometheus-operator only supports 1.16 or later?

bismarck:
@mrueg Regarding the < 1.14 dashboards and rules, we should probably remove them.

charts/kube-prometheus-stack/hack/sync_prometheus_rules.py
```diff
@@ -35,7 +35,7 @@ def new_representer(dumper, data):
         'min_kubernetes': '1.14.0-0'
     },
     {
-        'source': 'https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/release-0.1/manifests/prometheus-rules.yaml',
+        'source': 'https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/master/manifests/prometheus-rules.yaml',
```
fsschmitt, line 38: @mrueg @bismarck please be aware that prometheus-operator has removed this file from its repo:
https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/master/manifests/prometheus-rules.yaml

As they've moved in this PR:
https://github.com/prometheus-operator/kube-prometheus/pull/873/files?file-filters%5B%5D=.libsonnet&file-filters%5B%5D=.yaml&hide-deleted-files=true#diff-1d254ebb0389617e83d5b134983121c1ffbee7f93e68fd80cf4f1d22d93fbc43

PR #588, opened 2021-01-15
[kube-prometheus-stack] Allow override of prometheus and alertmanager default base images

QuentinBisson:
… 

This PR adds overridability of the prometheus and alertmanager base image which we are in need of as we are retagging images so they can be deployed in China (customer registry) and having to specify the image to use in every CR should be unnecessary if we can define the default image to use

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 13.2.1
+version: 13.2.2
 appVersion: 0.45.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 13.2.2
+version: 13.3.0
 appVersion: 0.45.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -59,6 +59,12 @@ spec:
             - --namespaces={{ $ns | join "," }}
             {{- end }}
             - --localhost=127.0.0.1
+            {{- if .Values.prometheusOperator.prometheusDefaultBaseImage }}
+            - --prometheus-default-base-image={{ .Values.prometheusOperator.prometheusDefaultBaseImage }}
+            {{- end }}
+            {{- if .Values.prometheusOperator.alertmanagerDefaultBaseImage }}
+            - --alertmanager-default-base-image={{ .Values.prometheusOperator.alertmanagerDefaultBaseImage }}
+            {{- end }}
             {{- if .Values.prometheusOperator.prometheusConfigReloaderImage.sha }}
             - --prometheus-config-reloader={{ .Values.prometheusOperator.prometheusConfigReloaderImage.repository }}:{{ .Values.prometheusOperator.prometheusConfigReloaderImage.tag }}@sha256:{{ .Values.prometheusOperator.prometheusConfigReloaderImage.sha }}
             {{- else }}
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1449,13 +1449,6 @@ prometheusOperator:
   ##
   # alertmanagerDefaultBaseImage: quay.io/prometheus/alertmanager
 
-  ## Configmap-reload image to use for reloading configmaps
-  ##
-  configmapReloadImage:
-    repository: docker.io/jimmidyson/configmap-reload
-    tag: v0.4.0
-    sha: ""
-
   ## Prometheus-config-reloader image to use for config and rule reloading
   ##
   prometheusConfigReloaderImage:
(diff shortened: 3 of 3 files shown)


TheoBrigitte (approved):

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -62,6 +62,20 @@ spec:
...
             {{- end }}
             - --localhost=127.0.0.1
+            {{- if .Values.prometheusOperator.prometheusDefaultBaseImage }}
+            {{- if .Values.prometheusOperator.prometheusDefaultBaseImage.sha }}
+            - --prometheus-default-base-image={{ .Values.prometheusOperator.prometheusDefaultBaseImage.repository }}:{{ .Values.prometheusOperator.prometheusDefaultBaseImage.tag }}@sha256:{{ .Values.prometheusOperator.prometheusDefaultBaseImage.sha }}
+            {{- else }}
+            - --prometheus-default-base-image={{ .Values.prometheusOperator.prometheusDefaultBaseImage.repository }}:{{ .Values.prometheusOperator.prometheusDefaultBaseImage.tag }}
+            {{- end }}
+            {{- end }}
+            {{- if .Values.prometheusOperator.alertmanagerDefaultBaseImage }}
```
monotek, line 72: Why don't you use "if and"?

monotek (changes requested):

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
QuentinBisson: I am not sure how you would use if and as there is a nested if in the first one to set the sha if specified

monotek: for example:

```
{{- if and .Values.prometheusOperator.prometheusDefaultBaseImage .Values.prometheusOperator.prometheusDefaultBaseImage.sha }}
```

```
{{- if and .Values.prometheusOperator.alertmanagerDefaultBaseImage .Values.prometheusOperator.alertmanagerDefaultBaseImage.sha }}
```

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -62,6 +62,20 @@ spec:
...
+            {{- if .Values.prometheusOperator.alertmanagerDefaultBaseImage.sha }}
+            - --alertmanager-default-base-image={{ .Values.prometheusOperator.alertmanagerDefaultBaseImage.repository }}:{{ .Values.prometheusOperator.alertmanagerDefaultBaseImage.tag }}@sha256:{{ .Values.prometheusOperator.alertmanagerDefaultBaseImage.sha }}
```
monotek, line 74: Why do you force usage of the sha?
Could be optional.

monotek (changes requested):

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
QuentinBisson: I guess I could use the following instead of using nesting but this is duplicating the condition needlessly in my opinion:
```
{{- if and .Values.prometheusOperator.prometheusDefaultBaseImage
 .Values.prometheusOperator.prometheusDefaultBaseImage.sha }}
{{- else if .Values.prometheusOperator.prometheusDefaultBaseImage }}
{{- end }}
```

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
QuentinBisson: I am not forcing the sha here. I am applying it if  `.Values.prometheusOperator.alertmanagerDefaultBaseImage.sha` is set, otherwise, `--alertmanager-default-base-image={{ .Values.prometheusOperator.alertmanagerDefaultBaseImage.repository }}:{{ .Values.prometheusOperator.alertmanagerDefaultBaseImage.tag }}` will be applied

I took this part from the prometheusConfigReloaderImage part below actually

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
QuentinBisson: @monotek I misunderstood how alertmanager-default-base-image was supposed to be configured as we should not set the tag and sha. I updated my PR to configure it as it should be

QuentinBisson:
@monotek I misunderstood how prometheus/alertmanager-default-base-image was supposed to be configured as we should not set the tag and sha. I updated my PR to configure it correctly

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1435,6 +1435,21 @@ prometheusOperator:
...
 
+  ## Prometheus image to use for prometheuses managed by the operator
+  ##
+  # prometheusDefaultBaseImage: quay.io/prometheus/prometheus
+
+  ## Alertmanager image to use for alertmanagers managed by the operator
+  ##
+  # alertmanagerDefaultBaseImage: quay.io/prometheus/alertmanager
+
+  ## Configmap-reload image to use for reloading configmaps
```
mrueg, line 1446: This should probably not be readded here.

QuentinBisson: Are you asking me to remove it from the values.yaml file or place it elsewhere in the file?

mrueg: The configmap-reload image got removed with the upgrade to prometheus-operator v0.45.0, so there's no need to add it back.

QuentinBisson: Thanks, I missed that during the rebase 🤦🏻

monotek (changes requested):
Please sign your commits to get dco check to pass.

QuentinBisson:
@monotek Thanks, I signed them now 🤦🏻

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 13.2.1
+version: 13.2.2
```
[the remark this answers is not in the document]
QuentinBisson: I bumped to 13.3.0

monotek (approved):

gkarthiks (approved):

PR #609, opened 2021-01-22
[alertmanager] allow possibility to use alertmanager in HA

### LINKED ISSUE #598: [alertmanager] add cluster peer option
NO

I would like to have possibility to deploy this chart in HA mode. I guess it should not be so hard, just add array in values.yaml, what will create args in docker entry point. Description of cmd options is here https://github.com/prometheus/alertmanager/blob/master/README.md#high-availability

Another chart what will deploy [prometheus + alertmanager](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L60) using extraArgs and providing "cluster.peer": FQDN. This however allows to use only 2 deployments (cannot use cluster.peer more than once). Also this deploys also prometheus, which is not required. I would like to have HA between k8s cluster, not just replicas in one cluster.

**Additional context**
Thank you in advance

palko444:
Signed-off-by: Pavol Stanka <stanka@ehs.sk>

This allows alertmanager to be used in HA mode, either in one cluster or between clusters.

fixes #598 

@monotek , @naseemkullah 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -31,29 +31,11 @@ spec:
     - port: 9094
       targetPort: 9094
       protocol: TCP
-      name: mesh-tcp
-    {{- end }}
-  selector:
-    {{- include "alertmanager.selectorLabels" . | nindent 4 }}
----
-{{ if or (gt .Values.replicaCount 1.0) (.Values.additionalPeers) }}
-apiVersion: v1
-kind: Service
-metadata:
-  name: {{ include "alertmanager.fullname" . }}-mesh
-  labels:
-    {{- include "alertmanager.labels" . | nindent 4 }}
-spec:
-  clusterIP: None
-  ports:
-    - port: 9094
-      targetPort: 9094
-      protocol: TCP
-      name: mesh-tcp
+      name: cluster-tcp
     - port: 9094
       targetPort: 9094
       protocol: UDP
-      name: mesh-udp
+      name: cluster-udp
+    {{- end }}
   selector:
     {{- include "alertmanager.selectorLabels" . | nindent 4 }}
-{{ end}}
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -27,5 +27,33 @@ spec:
       targetPort: http
       protocol: TCP
       name: http
+    {{- if or (gt .Values.replicaCount 1.0) (.Values.additionalPeers) }}
+    - port: 9094
+      targetPort: 9094
+      protocol: TCP
+      name: mesh-tcp
+    {{- end }}
   selector:
     {{- include "alertmanager.selectorLabels" . | nindent 4 }}
+---
+{{ if or (gt .Values.replicaCount 1.0) (.Values.additionalPeers) }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "alertmanager.fullname" . }}-mesh
+  labels:
+    {{- include "alertmanager.labels" . | nindent 4 }}
+spec:
+  clusterIP: None
+  ports:
+    - port: 9094
+      targetPort: 9094
+      protocol: TCP
+      name: mesh-tcp
+    - port: 9094
+      targetPort: 9094
+      protocol: UDP
+      name: mesh-udp
+  selector:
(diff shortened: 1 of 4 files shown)


monotek (changes requested):
Adding a new feature should raise minor version

palko444:
ok. I will create new pull request, with minor version increased

charts/alertmanager/templates/services.yaml
```diff
@@ -27,5 +27,33 @@ spec:
...
+      name: mesh-tcp
+    {{- end }}
+  selector:
+    {{- include "alertmanager.selectorLabels" . | nindent 4 }}
+---
+{{ if or (gt .Values.replicaCount 1.0) (.Values.additionalPeers) }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "alertmanager.fullname" . }}-mesh
```
naseemkullah, line 43: nit: it it makes more sense, replace `mesh` with `cluster`? Seems to be the word used in alertmanager docs for this feature.

Is there a reason why additional port is added to headless, but for non-headless a new service is created entirely? Could we instead just add 9094 to the main service like we do with headless?

naseemkullah (changes requested):

charts/alertmanager/templates/services.yaml
palko444: Hello @naseemkullah,
"mesh" name is used in this [chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus). I have version 11.6.0, installed in my cluster.

"Mesh service" is created for cluster to cluster communication, TCP & UDP needs to be allowed for gossip to work. As described [here](https://github.com/prometheus/alertmanager#high-availability). TCP port forward in nginx-ingress needs to point to this service. It's not required to allow also WEB UI port to be open outside of cluster directly.

If you still want to replace cluster with mesh, im ok with it, pls let me know.

Thank you in advance
Br Palo

naseemkullah: Hi Pablo, what I mean is can we instead of create a new service called -mesh can we just conditionally enable 9094 on the existing service? Like done in the headless service. As for the name I do have a slight preference for cluster but it's to your discretion and a non blocker

palko444: Hello @naseemkullah,
I will change from mesh to cluster.
Idea about new service, was to use "mesh" one for  communication purely for alertmanager gossip and "headless" for inside cluster communication.
Mesh has only 9094 TCP/UDP(NO 9093), headless has 9093 and 9094 both TCP.
This means, even if you forward port to "mesh/cluster" service, you cannot access WEB UI on port 9093.
If you think, this should be done in one service, I'm not against. But same method is used also in prometheus chart.

Thank you in advance
Br Palo

naseemkullah: But in the ingress config we can specify to forward to 9093 not 9094 while the service can still be configured with both ports?

Thanks,
BR

palko444: Hello @naseemkullah ,
this is true, however its easier to miss port number (9093/9094) than whole service. But again, if you think it better i will do only port conditionally and not create new service.
Br Palo

naseemkullah: I think it makes sense to reuse the same service in this case, please go ahead with that change and this should be good to go! Thanks.

palko444: Hello @naseemkullah 
Changes are done as proposed, let me please know in case something else is needed.
Thank you in advance
Br Palo

naseemkullah (approved):
Lgtm, thanks @palko444

naseemkullah (changes requested):
Actually we still need the conditional port 9094 in the no. Headless service I think?

naseemkullah:
Non headless

palko444:
hello @naseemkullah,
i guess non headless service is just for WEB UI.
cluster.peer in statefull_set is pointed to **pod.headless_service**
Or am I missing something else?
Br Palo

naseemkullah:
I was not sure about that but thanks for confirming

naseemkullah (dismissed):

monotek (approved):

PR #626, opened 2021-01-30
[kube-prometheus-stack] Allow override of job names for k3s compatibility

cablespaghetti:
For kube-scheduler and kube-controller-manager, make it possible to have the dashboards/rules enabled but ServiceMonitor disabled. Then make it possible to override the job name used in rules/dashboards.

This is useful for k3s where the "apiserver" job pulls in metrics for scheduler, controller-manager, as well as the API server. With this change it is possible to add the following values and have working dashboards/rules for these components (rather than just having to disable these features of the chart).

```
kubeScheduler:
  jobName: apiserver
  serviceMonitor:
    enabled: false
kubeControllerManager:
  jobName: apiserver
  serviceMonitor:
    enabled: false
```

Unfortunately it's not yet possible to scrape the new embedded etcd implementation until a k3s release built with this change: https://github.com/k3s-io/k3s/pull/2750

N/A

I have verified that under normal (non-k3s) circumstances there are no changes to the templated YAML. I've removed a couple of if statements in `rules-1.14/kubernetes-system-scheduler.yaml` and `rules-1.14/kubernetes-system-controller-manager.yaml` which weren't having any effect due to the big if statement at the top of these files.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -875,6 +875,8 @@ kubeControllerManager:
     #   component: kube-controller-manager
 
   serviceMonitor:
+    enabled: true
+
     ## Scrape interval. If not set, the Prometheus default scrape interval is used.
     ##
     interval: ""
@@ -906,6 +908,9 @@ kubeControllerManager:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+
+  ## Override job name used in rules and dashboards. Useful if ServiceMonitor disabled above, such as in k3s where kube-controller-manager metrics come from the "apiserver" job.
+  jobName: kube-controller-manager
 
 ## Component scraping coreDns. Use either this or kubeDns
 ##
@@ -1101,7 +1106,7 @@ kubeScheduler:
     #   replacement: $1
     #   action: replace
 
-  ## Override job name used in rules and dashboards. Useful if ServiceMonitor disabled above, such as in k3s where scheduler metrics come from the "apiserver" job.
+  ## Override job name used in rules and dashboards. Useful if ServiceMonitor disabled above, such as in k3s where kube-scheduler metrics come from the "apiserver" job.
   jobName: kube-scheduler
 
 ## Component scraping kube proxy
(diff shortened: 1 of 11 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -875,6 +875,8 @@ kubeControllerManager:
     #   component: kube-controller-manager
 
   serviceMonitor:
+    enabled: true
```
hectorj2f, line 878: why would it be disabled ?

cablespaghetti: For this use case you don't need the scrape job for controller-manager or scheduler because with k3s there is a single metrics endpoint for all three control plane services, and the apiserver scrape job/ServiceMonitor picks up the metrics for all of them.

cablespaghetti:
I know I and other k3s folks are keen to get this merged. Happy to explain my logic further if there are any questions and rebase if you're happy.

Thanks

charts/kube-prometheus-stack/values.yaml
hectorj2f: Why cannot it be configured in k3s instead of modifying a more generic and independent tool ?

hectorj2f: Should we add any condition/checks to this chart based on the opinionated decisions of every Kubernetes installer ? I feel it'd be a wrong decision.

cablespaghetti: I definitely see where you're coming from. The chart is already quite complex, let's not increase that complexity unnecessarily.

I'll see if there is some other approach we can take either in configuration of the Helm chart as it stands or a PR against k3s. However I did attempt this before making this PR, and I really don't want to maintain a fork.

bismarck (changes requested):
Please see the [note](https://github.com/prometheus-community/helm-charts/blob/35aa00bb4c66cbb543d7832dc14b49bc471cd6b4/charts/kube-prometheus-stack/templates/grafana/dashboards-1.14/controller-manager.yaml#L1-L5) at the top of the dashboard and rule files.

PR #660, opened 2021-02-09
[prometheus-cloudwatch-exporter] Set namespace

justinwalz:

Set namespace on objects where it is not set, to `.Release.namespace`

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-cloudwatch-exporter/Chart.yaml
+++ b/charts/prometheus-cloudwatch-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.10.0"
 description: A Helm chart for prometheus cloudwatch-exporter
 name: prometheus-cloudwatch-exporter
-version: 0.14.0
+version: 0.14.1
 home: https://github.com/prometheus/cloudwatch_exporter
 sources:
 - https://github.com/prometheus/cloudwatch_exporter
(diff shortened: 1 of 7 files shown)


monotek (commented):
Why should this be needed?
Helm should add the namespace without adding it to the chart explicitly.

justinwalz:
@monotek We use tanka to wrap our helm deployments, https://tanka.dev/helm, which doesn't seem to set it.

torstenwalter:
@monotek It also helps people who are using helm only to render the manifests and other tools for the deployment e.g. Argo CD.
Having the namespace there helps if one Argo CD application is targeting multiple namespaces. 

With `kubectl apply -n single-namespace` you can only specify a single namespace ;-)

justinwalz:
Yup that's pretty much what we have - one tanka setup targeting many namespaces. In the meantime, for our setup, I've overwritten each object to merge in metadata.namespace with jsonnet, however I believe this PR is still helpful for others.

monotek:
Ok, using helm template for some cd tool is a valid point.

But maybe one could also use postrendering with kustomize to achieve this? 

See: 
https://helm.sh/docs/topics/advanced/


See:

justinwalz:
Yeah I agree, one could use postrendering. This is why we use `tanka`, as an alternative to helm postrendering. We apply our own changes to the resources helm generates (additional labels, changing deployment parameters, etc). In this case, I've added `metadata.namespace` as well. 

Note, I pushed a new commit to rebase `main` branch to fix merge conflicts.

torstenwalter:
@monotek I would not be against having this in the chart. It does not really harm to explicitly render release namespace.
For the Jenkins chart I did something like this a while ago, but I maybe biased here therefore I would leave it up to you to judge.

monotek:
I'm ok with it too :)

monotek (approved):

torstenwalter (approved):

PR #679, opened 2021-02-16
[prometheus-postgres-exporter]: Add annotations to postgres-exporter ServiceAccount

hhtpcd:

Allow setting annotations on the ServiceAccount used by the
postgres-exporter Deployment. If we are running a sidecar container such
as the Google Cloud SQL Proxy we can annotate the service account to
enable Workload Identity. The approach is the same in EKS clusters
also.

I notice in the `values.yaml` there is already a reference to running the
cloud SQL proxy as a sidecar container.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 1.9.1
+version: 1.10.0
 home: https://github.com/wrouesnel/postgres_exporter
 sources:
 - https://github.com/wrouesnel/postgres_exporter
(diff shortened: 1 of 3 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.8.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 1.9.1
+version: 1.9.2
```

monotek (changes requested):

charts/prometheus-postgres-exporter/Chart.yaml
hhtpcd: Thanks, addressed. Bumped to `1.10.0`

monotek (approved):

hhtpcd:
@gianrubio @zanhsieh Do let me know if there is anything missing from the pull request :)

zanhsieh (approved):

PR #691, opened 2021-02-20
[prometheus] fix imagePullSecrets indentation

Patricol:
The `imagePullSecrets` list is not indented enough; and applying charts with `imagePullSecrets` set fails with errors like `Error: YAML parse error on prometheus/templates/alertmanager/deploy.yaml: error converting YAML to JSON: yaml: line 74: did not find expected key`.

It seems hard to imagine this bug went unnoticed for >2 years, so I'm guessing newer versions of helm are more strict about indentation.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: prometheus
-version: 13.3.2
+version: 13.3.3
 appVersion: 2.24.0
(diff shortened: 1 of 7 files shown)


Patricol:
I expected the behavior to only be present when `len(imagePullSecrets)>1`; but it seems to affect `len(imagePullSecrets)==1` too in my repro attempts.

monotek (approved):

PR #700, opened 2021-02-22
[prometheus] Add deploymentLabels to prometheus server chart

### LINKED ISSUE #699: [prometheus] deployment labels are not configurable
We're currently deploying using the prometheus helm chart, but have a need to configure deployment labels. This is not currently configurable

I'd like to be able to configure deployment labels the same way deployment annotations are configurable.

We've considered downloading a local copy of the chart, but figured this could be a common enough usecase to propose an update.

codeboten:
This allows users to configure labels via the `server.deploymentLabels` configuration option

  - fixes #699

#### Checklist

- [x] Chart Version bumped

Signed-off-by: Alex Boten <aboten@lightstep.com>

### DIFF
--- a/charts/prometheus/templates/server/deploy.yaml
+++ b/charts/prometheus/templates/server/deploy.yaml
@@ -9,6 +9,9 @@ metadata:
 {{- end }}
   labels:
     {{- include "prometheus.server.labels" . | nindent 4 }}
+{{- if .Values.server.deploymentLabels }}
+    {{ toYaml .Values.server.deploymentLabels | nindent 4 }}
+{{- end }}
   name: {{ template "prometheus.server.fullname" . }}
 {{ include "prometheus.namespace" . | indent 2 }}
 spec:
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: prometheus
-version: 13.3.3
+version: 13.4.0
 appVersion: 2.24.0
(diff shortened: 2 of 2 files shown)


charts/prometheus/templates/server/deploy.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
 {{- end }}
   labels:
     {{- include "prometheus.server.labels" . | nindent 4 }}
+{{- if .Values.server.deploymentLabels }}
```
monotek, line 12: Imho this should be named extra labels and added to all resources.

monotek (changes requested):

monotek:
Maybe we could even extend the metalabels helper, so it includes some additional labels from the values file?

see: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/_helpers.tpl#L24

codeboten:
> Maybe we could even extend the metalabels helper, so it includes some additional labels from the values file?
> 
> see: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/_helpers.tpl#L24

Happy to go either ways, could use some guidance using the metalabels helper if that's the preferred approach.

charts/prometheus/templates/server/deploy.yaml
jmacd: Yes, for example the node_exporter daemonset can't be labeled either.

jmacd (approved):

PR #724, opened 2021-03-03
[prometheus-rabbitmq-exporter] Polish up Prometheus rules

machine424:
I don't know if we should keep the old rules, if we should only bump the patch version...

The existing rules are vague.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-rabbitmq-exporter/Chart.yaml
+++ b/charts/prometheus-rabbitmq-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Rabbitmq metrics exporter for prometheus
 name: prometheus-rabbitmq-exporter
-version: 0.7.0
+version: 1.0.0
 appVersion: v0.29.0
 home: https://github.com/kbudde/rabbitmq_exporter
 sources:
--- a/charts/prometheus-rabbitmq-exporter/templates/prometheusrule.yaml
+++ b/charts/prometheus-rabbitmq-exporter/templates/prometheusrule.yaml
@@ -12,36 +12,58 @@ metadata:
   {{- end }}
 spec:
   groups:
-  - name: rabbitmq
-    rules:
-    - alert: RabbitmqDown
-      expr: |
-        absent(rabbitmq_up == 1)
-      for: 1m
-      labels:
-        severity: critical
-    - alert: RabbitmqNotRunning
-      expr: |
-        rabbitmq_running{self="1"} != 1
-      for: 1m
-      labels:
-        severity: critical
-    - alert: RabbitmqMessages
-      expr: |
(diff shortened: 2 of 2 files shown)


charts/prometheus-rabbitmq-exporter/templates/prometheusrule.yaml
```diff
@@ -12,36 +12,55 @@ metadata:
...
-        severity: critical
-    - alert: RabbitmqPartition
-      expr: |
-        rabbitmq_partitions{self="1"} != 0
-      for: 1m
-      labels:
-        severity: critical
+    - name: {{ template "prometheus-rabbitmq-exporter.fullname" . }}
+      rules:
+        - alert: RMQNodeDown
```
desaintmartin, line 17: Should we stick with the previous naming, like `RabbitmqNodeDown`, which is more explicit ? So that it is closer than what we can see elsewhere, like https://awesome-prometheus-alerts.grep.to/rules.html#rabbitmq ?

charts/prometheus-rabbitmq-exporter/templates/prometheusrule.yaml
```diff
@@ -12,36 +12,55 @@ metadata:
...
-        severity: critical
-    - alert: RabbitmqNotRunning
-      expr: |
-        rabbitmq_running{self="1"} != 1
-      for: 1m
-      labels:
-        severity: critical
-    - alert: RabbitmqMessages
-      expr: |
-        rabbitmq_queue_messages > 250
```
desaintmartin, line 31: Indeed, this one made little sense if you have a huge environment...

charts/prometheus-rabbitmq-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Rabbitmq metrics exporter for prometheus
 name: prometheus-rabbitmq-exporter
-version: 0.6.0
+version: 0.7.0
```
desaintmartin, line 4: Probably a breaking change, since it changes/deletes rules. Better be safe!

desaintmartin (commented):
thanks for the contribution!

charts/prometheus-rabbitmq-exporter/templates/prometheusrule.yaml
machine424: Done, `Rabbitmq` everywhere

charts/prometheus-rabbitmq-exporter/Chart.yaml
machine424: yep

desaintmartin (approved):

PR #733, opened 2021-03-05
[prometheus] Allow to enable hostNetwork

randrusiak:

In the case of using custom CNI such us calico on EKS, prometheus-server needs to use hostNetwork. The same setting is available in kube-prometheus-stack chart.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -946,6 +946,14 @@ server:
     #   cpu: 500m
     #   memory: 512Mi
 
+  # Required for use in managed kubernetes clusters (such as AWS EKS) with custom CNI (such as calico),
+  # because control-plane managed by AWS cannot communicate with pods' IP CIDR and admission webhooks are not working
+  ##
+  hostNetwork: true
+
+  # When hostNetwork is enabled, you probably want to set this to ClusterFirstWithHostNet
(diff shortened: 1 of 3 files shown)


charts/prometheus/values.yaml
```diff
@@ -946,6 +946,14 @@ server:
     #   cpu: 500m
     #   memory: 512Mi
 
+  # Required for use in managed kubernetes clusters (such as AWS EKS) with custom CNI (such as calico),
+  # because control-plane managed by AWS cannot communicate with pods' IP CIDR and admission webhooks are not working
+  ##
+  hostNetwork: true
```
monotek, line 952: This should not be enabled by default.

charts/prometheus/values.yaml
```diff
@@ -946,6 +946,14 @@ server:
...
+
+  # When hostNetwork is enabled, you probably want to set this to ClusterFirstWithHostNet
+# dnsPolicy: ClusterFirstWithHostNet
```
monotek, line 955: Please near and set the default value "ClusterFirst".

monotek (changes requested):

randrusiak (commented):
Of course, the host network shouldn't be enabled by default I forgot to disable this option after local tests.

monotek (approved):

PR #741, opened 2021-03-08
[prometheus-pingdom-exporter] Add secret override for pingdom exporter

dashford:

This PR adds the ability to specify an existing `Secret` in place of the generated secret when the chart is installed. This is useful in our case because we manage our secrets externally so already have a secure mechanism in place to generate this.

N/A

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-pingdom-exporter/Chart.yaml
+++ b/charts/prometheus-pingdom-exporter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-pingdom-exporter
-version: 2.3.2
+version: 2.4.2
 appVersion: 20190610-1
 home: https://github.com/giantswarm/prometheus-pingdom-exporter
 description: A Helm chart for Prometheus Pingdom Exporter
--- a/charts/prometheus-pingdom-exporter/Chart.yaml
+++ b/charts/prometheus-pingdom-exporter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-pingdom-exporter
-version: 2.4.2
+version: 2.4.0
 appVersion: 20190610-1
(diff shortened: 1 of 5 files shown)


charts/prometheus-pingdom-exporter/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-pingdom-exporter
-version: 2.3.2
+version: 2.4.2
```
monotek, line 3: This should be 2.4.0 or not?

charts/prometheus-pingdom-exporter/values.yaml
```diff
@@ -56,6 +56,9 @@ pod:
     # key: "true"
     # example: "false"
 
+existingSecret:
+  name: ""
+
```
monotek, line 61: Imho key should be configurable too

monotek (changes requested):

charts/prometheus-pingdom-exporter/values.yaml
dashford: Thanks for the feedback, I'm not sure I understand the change I can make here if you could elaborate a little bit? The current secret has four keys:

```
user
password
appId
accountEmail
```

Are you suggesting something where those four keys could be configurable in the values e.g.

```
existingSecret:
  name: ""
  userKey: "user"
  passwordKey: "password"
  appIdKey: "appId"
  accountEmailKey: "accountEmail"
```

monotek: Depending where your secrets are coming from you might not be able to control what's the key in the secret is.

So it might be good to have it configurable.

monotek (approved):

PR #750, opened 2021-03-10
[kube-prometheus-stack] set cert-manager generated certificate duration to include minutes and seconds

muhlba91:
@vsliouniaev , @bismarck , @gianrubio , @gkarthiks , @scottrigby , @Xtigyro 

cert-manager updates the certificate specification and sets the duration including minutes and seconds. this leads to a representation of e.g. `8760h0m0s` instead of `8760h` which is generally fine if you only apply and use the chart.
however, when using tools like argo-cd which synchronise and auto-heal out-of-sync manifests, this is detected as a drift and will get synchronised over and over again.
setting minutes and seconds in the specification resolves this problem.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 14.0.1
+version: 14.0.2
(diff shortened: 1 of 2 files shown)


jsoref:
@bismarck: what would it take to make progress on this?

I'd be happy to make a PR that allowed suppressing this field entirely. None of the other certs we have specify a duration, and the value used here is not in line with best practices which are evolving towards 90 day certificates or even shorter.

bismarck (approved):
lgtm

PR #767, opened 2021-03-15
[Alertmanager] update: Enable annotations for service tpl

ffafi-polarsquad:
Signed-off-by: ffafi <firas.fafi@polarsquad.com>
@monotek @naseemkullah 

I'm providing here minors changes for Service Annotations. I have noticed that it's not configurable via the Alertmanager Helm chart. 
The annotations might be useful for some custom clusters, in kube-stack-prometheus chart it's already enabled but here it's missing.

There are also cases when the alertmanager is linked to different Prometheus instances via external endpoint and this might involve exposure of service via LoadBalancer mode and custom annotations.

Example of usage for being able to expose service privately via LB mode or define domain using external-dns ... 

```
service:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: mydomain.com
    service.beta.kubernetes.io/azure-load-balancer-internal: true
  type: LoadBalancer
```

### DIFF
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.6.0
+version: 0.7.0
 appVersion: v0.21.0
 maintainers:
   - name: monotek
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -4,6 +4,10 @@ metadata:
   name: {{ include "alertmanager.fullname" . }}
   labels:
     {{- include "alertmanager.labels" . | nindent 4 }}
+{{- if .Values.service.annotations }}
+  annotations:
+    {{ toYaml .Values.service.annotations | indent 4 }}
+{{- end }}
 spec:
   type: {{ .Values.service.type }}
   ports:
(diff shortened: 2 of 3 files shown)


charts/alertmanager/Chart.yaml
```diff
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.6.0
+version: 0.6.1
```
monotek, line 9: New feature should raise minor version.

monotek (changes requested):

charts/alertmanager/templates/services.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: {{ include "alertmanager.fullname" . }}
   labels:
     {{- include "alertmanager.labels" . | nindent 4 }}
+{{- if .Values.service.annotations }}
+  annotations:
+{{ toYaml .Values.service.annotations | indent 4 }}
```
naseemkullah, line 9:
```suggestion
    {{- toYaml .Values.service.annotations | nindent 4 }}
```

nit: reads better imho

charts/alertmanager/Chart.yaml
ffafi-polarsquad: Sorry somehow I forgot to commit it. I will fix it now thanks!

monotek (approved):

monotek (approved):

PR #770, opened 2021-03-17
[kube-prometheus-stack] Exclude default rules from namespace enforcement

amuraru:

When enforcedNamespaceLabel is enabled the default rules installed by the
helm chart also also changed dynamically by the operator at runtime.
This breaks cluster wide recorded rules and alert rules.

This patch ensures that default prometheus rules are automatically
added to `prometheusRulesExcludedFromEnforce`

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -268,9 +268,6 @@ spec:
 {{- end }}
   portName: {{ .Values.prometheus.prometheusSpec.portName }}
 {{- end }}
-{{- if .Values.prometheus.prometheusSpec.enforcedNamespaceLabel }}
-  enforcedNamespaceLabel: {{ .Values.prometheus.prometheusSpec.enforcedNamespaceLabel }}
-{{- end }}
 {{- if .Values.prometheus.prometheusSpec.volumes }}
   volumes:
 {{ toYaml .Values.prometheus.prometheusSpec.volumes | indent 4 }}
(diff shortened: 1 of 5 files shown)


amuraru:
also see https://github.com/prometheus-operator/prometheus-operator/issues/3919

Xtigyro:
Hey @amuraru 👋 
Thank you for the contribution - it's appreciated! 🥇 

I'll be having a look at this PR today and hopefully will be merged by Monday.

charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -292,10 +289,19 @@ spec:
 {{- if .Values.prometheus.prometheusSpec.ignoreNamespaceSelectors }}
   ignoreNamespaceSelectors: {{ .Values.prometheus.prometheusSpec.ignoreNamespaceSelectors }}
 {{- end }}
-{{- if .Values.prometheus.prometheusSpec.prometheusRulesExcludedFromEnforce }}
+{{- if .Values.prometheus.prometheusSpec.enforcedNamespaceLabel }}
```
amuraru, line 292: I moved this here and nest with the `prometheusDefaultRulesExcludedFromEnforce` and `prometheusRulesExcludedFromEnforce` as the latters are not applicable unless `enforcedNamespaceLabel` is set

Xtigyro (changes requested):
@amuraru Could we please think of a way for automating the maintenance of that explicit list? With this PR if you add a new group of rules - you'll need to add it to one more place.

amuraru:
> Could we please think of a way for automating the maintenance of that explicit list? With this PR if you add a new group of rules - you'll need to add it to one more place.

thanks for your suggestion @Xtigyro - please check the new version, the patch now generates automatically the rules names dynamically in sync_prometheus_rules.py and stores them as a yaml file where they are dynamically loaded from.
lmk

Xtigyro:
@amuraru Very nice! And how about naming `files` - `aux`?

amuraru:
What about `generated` and try to move generated rules file there as well?

amuraru:
@Xtigyro please check again - I changed the generated yaml file to a helm named template.
I think it looks more compact now

Xtigyro (approved):

Xtigyro:
@amuraru Sweet! Many thanks for your excellent work! 💯

PR #781, opened 2021-03-19
[Alertmanager] command podannotations

ngc104:

This PR adds 2 new abilities to Alertmanager chart :
* add the possibility to specify our own command
* add the possibility add annotations to the statefulset Pod (e.g. podAnnotations)

When you want to use Vault Sidecar Injector (https://www.vaultproject.io/docs/platform/k8s/injector) you first need to be able to specify annotations to your pod.

This will create a file `/vault/secrets/whatever`.

Then, in order to take that file into account, the easier way is to use it on the command line, usually like this : `['source /vault/secrets/whatever; /bin/alertmanager']`

This is why this patch allows adding podAnnotations and command.

#### Checklist

- [X] Chart Version bumped

Note : I did not update Chart Version because it depends on when the PR is merged.
Edit : I finally updated it.

### DIFF
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -20,7 +20,7 @@ spec:
         {{- include "alertmanager.selectorLabels" . | nindent 8 }}
 {{- if .Values.podAnnotations }}
       annotations:
-        {{ toYaml .Values.podAnnotations | nindent 8 }}
+        {{- toYaml .Values.podAnnotations | nindent 8 }}
 {{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
@@ -61,7 +61,7 @@ spec:
                   fieldPath: status.podIP
 {{- if .Values.command }}
           command:
-            {{ toYaml .Values.command | nindent 12 }}
+            {{- toYaml .Values.command | nindent 12 }}
 {{- end }}
           args:
             - --storage.path=/alertmanager
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -18,9 +18,9 @@ spec:
     metadata:
       labels:
         {{- include "alertmanager.selectorLabels" . | nindent 8 }}
-{{- if .Values.statefulSet.podAnnotations }}
+{{- if .Values.podAnnotations }}
       annotations:
-        {{ toYaml .Values.statefulSet.podAnnotations | nindent 8 }}
+        {{ toYaml .Values.podAnnotations | nindent 8 }}
 {{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
@@ -59,9 +59,9 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
-{{- if .Values.statefulSet.command }}
+{{- if .Values.command }}
           command:
-            {{ toYaml .Values.statefulSet.command | nindent 12 }}
+            {{ toYaml .Values.command | nindent 12 }}
 {{- end }}
           args:
             - --storage.path=/alertmanager
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -18,6 +18,10 @@ spec:
     metadata:
       labels:
         {{- include "alertmanager.selectorLabels" . | nindent 8 }}
+{{- if .Values.statefulSet.podAnnotations }}
+      annotations:
+        {{ toYaml .Values.statefulSet.podAnnotations | nindent 8 }}
+{{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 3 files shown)


charts/alertmanager/values.yaml
```diff
@@ -84,6 +84,8 @@ affinity: {}
 
 statefulSet:
   annotations: {}
+  podAnnotations: {}
+  command: []
```
naseemkullah, line 88:
```suggestion

podAnnotations: {}

command: []
```

nit: keep these top level

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -55,6 +59,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
+{{- if .Values.statefulSet.command }}
+          command:
+            {{ toYaml .Values.statefulSet.command | nindent 12 }}
```
naseemkullah, line 64:
```suggestion
{{- if .Values.command }}
          command:
            {{- toYaml .Values.command | nindent 12 }}
```

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -18,6 +18,10 @@ spec:
     metadata:
       labels:
         {{- include "alertmanager.selectorLabels" . | nindent 8 }}
+{{- if .Values.statefulSet.podAnnotations }}
+      annotations:
+        {{ toYaml .Values.statefulSet.podAnnotations | nindent 8 }}
```
naseemkullah, line 23:
```suggestion
{{- if .Values.podAnnotations }}
      annotations:
        {{- toYaml .Values.podAnnotations | nindent 8 }}
```

naseemkullah (changes requested):
nit: put these at top level

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -18,6 +18,10 @@ spec:
     metadata:
       labels:
         {{- include "alertmanager.selectorLabels" . | nindent 8 }}
+{{- if .Values.podAnnotations }}
+      annotations:
+        {{ toYaml .Values.podAnnotations | nindent 8 }}
```
naseemkullah, line 23:
```suggestion
        {{- toYaml .Values.podAnnotations | nindent 8 }}
```

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -55,6 +59,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
+{{- if .Values.command }}
+          command:
+            {{ toYaml .Values.command | nindent 12 }}
```
naseemkullah, line 64:
```suggestion
            {{- toYaml .Values.command | nindent 12 }}
```

ngc104:
@naseemkullah I added a new commit with your suggestion.
I also bumped the version of the chart (not in the initial commits of the PR)

naseemkullah (changes requested):
notice the `-`

naseemkullah:
> @naseemkullah I added a new commit with your suggestion.
> I also bumped the version of the chart (not in the initial commits of the PR)

Thanks, we need the dash at the beggining when using `| nindent`

charts/alertmanager/templates/statefulset.yaml
ngc104: If I add the `-` I should replace `nindent` with `indent` ?
What do you prefer ?

ngc104:
@naseemkullah I guess I should not add the `-` as you suggest. As for me, I really don't care, I can do it if you insist.
But the reason why I shouldn't do it is that we can find such code in the same file. Example : 

file : templates/statefulset.yaml
```
{{- if .Values.statefulSet.annotations }}
  annotations:
    {{ toYaml .Values.statefulSet.annotations | nindent 4 }}
{{- end }}
```
This code already exists. This is not my code.

What do you think ?

charts/alertmanager/templates/statefulset.yaml
naseemkullah: as the suggestion indicates

naseemkullah:
> @naseemkullah I guess I should not add the `-` as you suggest. As for me, I really don't care, I can do it if you insist.
> But the reason why I shouldn't do it is that we can find such code in the same file. Example :
> 
> file : templates/statefulset.yaml
> 
> ```
> {{- if .Values.statefulSet.annotations }}
>   annotations:
>     {{ toYaml .Values.statefulSet.annotations | nindent 4 }}
> {{- end }}
> ```
> 
> This code already exists. This is not my code.
> 
> What do you think ?

please see suggestion

charts/alertmanager/templates/statefulset.yaml
ngc104: done

charts/alertmanager/templates/statefulset.yaml
ngc104: done

ngc104:
I rediscovered `nindent` : It was doing something different than what I thought. It just took the time for me to understand that. Sorry.
I fixed it it in the code @naseemkullah

naseemkullah (approved):
Thanks!

PR #796, opened 2021-03-26
[kube-prometheus-stack] Don't set grafana.adminPassword

sathieu:

By default (i.e. in grafana chart), it is randomly generated which is more secure.

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -617,7 +617,7 @@ grafana:
   ##
   defaultDashboardsEnabled: true
 
-  adminPassword: prom-operator
+  # adminPassword: prom-operator
 
   ingress:
     ## If true, Grafana Ingress will be created
--- a/charts/kube-prometheus-stack/Chart.lock
+++ b/charts/kube-prometheus-stack/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
-  version: 3.1.0
+  version: 3.1.1
 - name: prometheus-node-exporter
   repository: https://prometheus-community.github.io/helm-charts
   version: 1.18.0
 - name: grafana
   repository: https://grafana.github.io/helm-charts
-  version: 6.9.1
-digest: sha256:288156ef9d1e49c047a24ede1b3b8e852aaa08aa4c03c96f65d7f42977f19251
-generated: "2021-05-18T17:09:28.105172864+02:00"
+  version: 6.9.3
+digest: sha256:7fbef041be9173650a0fa9db6cf70b415ef9c9afd702c8ef73471a3a9cbfcc7b
+generated: "2021-06-02T18:54:07.604092583+02:00"
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 16.1.2
+version: 17.0.0
 appVersion: 0.48.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
@@ -44,6 +44,6 @@ dependencies:
   repository: https://prometheus-community.github.io/helm-charts
   condition: nodeExporter.enabled
 - name: grafana
-  version: "6.9.*"
+  version: "~6.9.3"
   repository: https://grafana.github.io/helm-charts
   condition: grafana.enabled


Xtigyro:
@sathieu Thanks for your contribution - it's appreciated 👍

Xtigyro (changes requested):
What happens if you change something and redeploy the chart - do you need to still set the password so it doesn't change on each upgrade?

monotek:
Related discussion: https://github.com/prometheus-community/helm-charts/issues/506

sathieu:
OK. Better to wait for https://github.com/grafana/helm-charts/pull/62#issuecomment-745426838 to be merged first...

sathieu:
Not stale

sathieu:
Not stale, but waiting for https://github.com/grafana/helm-charts/pull/62.

sathieu:
@Xtigyro Please review again, now that grafana/helm-charts#62 is merged.

Xtigyro:
@sathieu On it.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -617,7 +617,7 @@ grafana:
   ##
   defaultDashboardsEnabled: true
 
-  adminPassword: prom-operator
+  # adminPassword: prom-operator
```
Xtigyro, line 620: I find it suboptimal because now the same password is set on each run of `helm upgrade` - hence, I believe we leave that line uncommented and put a comment above it that it can be commented out and a note also saying that the same randomly generated pwd will be set on each `helm upgrade`.

Xtigyro (changes requested):

Xtigyro:
@sathieu Also - we should have a line in the `templates/NOTES.txt` that prints out the randomly generated pwd of Grafana at the end of the deployment process.

charts/kube-prometheus-stack/values.yaml
sathieu: @Xtigyro I don't understand why you find this suboptimal? Querying the cluster (using `lookup`) during manifests generation?

Xtigyro: Because the same randomly generated pwd will be set on each `helm upgrade` run. Hence, it is also shown as a difference between executions of `helm upgrade`. It will also cause issues in tools such as ArgoCD and Flux. So it's just not worth the hassle. We can still give that opportunity to people wanting to have that as stated in my original comment.

steven-sheehy:
> Also - we should have a line in the templates/NOTES.txt that prints out the randomly generated pwd of Grafana at the end of the deployment process.

Printing out a password could be a security concern depending upon how the upgrade is ran. Also, as the password generation is done within the secret.yaml and only on first install, it can't also be printed in the NOTES.txt since it's not available in that context. If it's an upgrade, it could be looked up in the NOTES.txt and in the secret.yaml but would require two lookups, which is inefficient.

charts/kube-prometheus-stack/values.yaml
steven-sheehy: > it is also shown as a difference between executions of helm upgrade

I wouldn't characterize it as a difference between executions. If you run two `helm upgrade --debug` commands back to back, the output of the Secret would be identical between the two. Hence there is no difference between executions when using auto-generation of secrets with lookup functionality.

> It will also cause issues in tools such as ArgoCD and Flux.

I use Flux with kube-prometheus-stack and `adminPassword` explicitly set to empty string, and it works perfectly. It will auto generate a secret only on initial install. On every upgrade it will pull the existing password and use that so it stays the same. Internally, Flux does the equivalent of a helm upgrade.

ArgoCD is a different matter as it uses `helm template` and `lookup` functionality does not work with it. This would cause upgrades for users of `helm template` with default chart values to generate new passwords on every upgrade and restart Grafana.

One could argue that they should not be using default passwords as it's insecure. Most charts do not set default passwords for this reason. Bitnami has even taken to not putting a default and throwing an error if it's empty to make it explicit.

To summarize:
* If they supply a custom `adminPassword`, it's not breaking regardless of deploy method.
* If they use the chart default:
  * helm install: Not breaking
  * helm upgrade: Not breaking. Will pull `prom-operator` from existing secret and reuse
  * helm template: Breaking. New password every upgrade and pod restarted

If we're worried about it being a breaking change for `helm template`, then we just bump the chart major version and denote it in the upgrade notes. This is standard practice.

Xtigyro:
> > Also - we should have a line in the templates/NOTES.txt that prints out the randomly generated pwd of Grafana at the end of the deployment process.
> 
> 
> 
> Printing out a password could be a security concern depending upon where those notes are displayed. The person deploying, the person able to access the deploy logs and the person using Grafana could all have different levels of permission. It's more secure for the user to retrieve them from the cluster using their kubectl credentials. Retrieval is a one time operation and they can store in their password manager of choice.
> 
> 
> 
> Also, as the password generation is done within the secret.yaml and only on first install, it can't also be printed in the NOTES.txt since it's not available in that context. If it's an upgrade, it could be looked up in the NOTES.txt and in the secret.yaml but would require two lookups, which is inefficient.
> 
> 
> 
> Note that Grafana chart NOTES.txt displays doesn't display the password but shows the command  to retrieve the admin password. Perhaps a similar line to kube-prometheus-stack that prints out how to retrieve the credentials should suffice?

Yes - let's print the command. Agreed.

charts/kube-prometheus-stack/values.yaml
Xtigyro: It resets the pwd to the same value on every `helm upgrade`. And when you use the "helm diff" plug-in, it shows it as a diff. It even restarted Grafana on some executions for me. I'll retest but that happened last time.

steven-sheehy: From the [docs](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) 

> Keep in mind that Helm is not supposed to contact the Kubernetes API Server during a helm template or a helm install|update|delete|rollback --dry-run, so the lookup function will return an empty list (i.e. dict) in such a case.

So it won't work with `helm diff`. We should just document that if users are using `helm template` or `helm upgrade --dry-run` (or tools that use these) they should set a manual password.

Xtigyro: And then the obvious question comes - is it worth the hassle? And also - that chart won't be fully compatible with ArgoCD, helm diff, helmfile and probably many other tools? If you've already set a certain pwd - you won't care, that's clear. So all the current users of the chart will need to change their values - so that they don't break their deployment. On the contrary - all the new users will have a more secure pwd by default.

Let's vote I think.
@bismarck @gianrubio @gkarthiks @scottrigby @vsliouniaev

steven-sheehy:
Would like for maintainers to weigh in if this would be accepted or not if rebased.

skluthe:
Can someone reopen this? I would like to see this merged.

sathieu:
@skluthe You can use `adminPassword:` (i.e. `adminPassword: null`) to have the expected behavior.

Example (from [here](https://gitlab.com/kubitus-project/kubitus-installer/-/blob/9f9fee57d6a08655dc6bd65f422171092e2ab2d4/roles/prometheus_stack/templates/values.yaml.j2#L26-28)):

```yaml
grafana:
  adminPassword:
  admin:
    existingSecret: prometheus-stack-grafana-admin

```

guillheu:
Looking forward to this being merged.
The current default password setting gives an insecure direction for less experienced DevOps who might end up putting their actual password in plain text in a values file.
Also, I don't think it is that inconvenient to manually look up the volatile random grafana secret from the cluster secret itself.
If a user doesn't have access to that resource because of RBAC, they probably shouldn't be accessing the default admin account anyways.

PR #806, opened 2021-03-31
[prometheus] Use server.prefixURL for scraping Prometheus metrics

jacksgt:

When the user specifies a value for server.baseURL (which affects --web.external-url configuration of Prometheus),
we also need to use the path server.prefixURL when querying ourselves.
Otherwise, localhost:9090 will just return "404 - Not Found".

#### Checklist

- [x] Chart Version bumped


zanhsieh:
@jacksgt 
Please sign DCO and bump chart version.

charts/prometheus/values.yaml
```diff
@@ -1297,7 +1297,7 @@ serverFiles:
       - job_name: prometheus
         static_configs:
           - targets:
-            - localhost:9090
+            - localhost:9090{{.Values.server.baseURL}}/metrics
```
naseemkullah, line 1300: base URL encompasses scheme + host + base path if I'm not mistaken, whereas the target should be a host, we should strip away the scheme if any

naseemkullah (changes requested):

charts/prometheus/values.yaml
jacksgt: Yes, you are correct. I got confused with prefixURL. Updated.

jacksgt:
I signed the DCO (I hope it's fine in a separate commit)

jacksgt:
@zanhsieh I signed the DCO. Could you check again? Thanks.

charts/prometheus/values.yaml
```diff
@@ -1297,7 +1297,7 @@ serverFiles:
       - job_name: prometheus
         static_configs:
           - targets:
-            - localhost:9090
+            - localhost:9090{{.Values.server.prefixURL}}/metrics
```
monotek, line 1300: this looks like it breaks it for updating users

monotek (changes requested):

charts/prometheus/values.yaml
jacksgt: I was thinking about this, but in which case would it break?

1. The user did not specify a prefixURL (the default) -> nothing changes
2. The user has already specified a prefixURL (e.g. because they are using a baseURL with a path) -> it was already broken for them, since Prometheus was listening on `/prefix/metrics`, but the scrape target did not reflect this

Am I missing something?

monotek: The default without prefix url would change from localhost:9090 to localhost:9090/metrics

naseemkullah: In fact, we should configure `prefixURL/metrics` on `metrics_path` not `static_config` according to https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

naseemkullah (changes requested):

charts/prometheus/values.yaml
jacksgt: @naseemkullah Excellent suggestion! I have updated the PR accordingly.

naseemkullah (dismissed):

zanhsieh (approved):

charts/prometheus/values.yaml
```diff
@@ -1298,6 +1298,7 @@ serverFiles:
         static_configs:
           - targets:
             - localhost:9090
+        metrics_path: {{.Values.server.prefixURL}}/metrics
```
zanhsieh, line 1301: Self reference not allowed.

zanhsieh (changes requested):

charts/prometheus/values.yaml
monotek: Please change to:

```
        metrics_path: /metrics
```

So the default is used for all users and you can still change your path in values.yaml.

monotek (changes requested):

charts/prometheus/values.yaml
naseemkullah: At that point no need to even add that, users with a prefixURL will simply have to reimplement the entire     `scrape_configs` array

naseemkullah (changes requested):

charts/prometheus/values.yaml
monotek: True. 
Adding it would maybe be a bit more verbose and could help to point the next user to the right direction if he wants to change it.
In the end i have no strong opinion about to add it or not.

jacksgt: Hmm, I'm pretty unhappy with this solution then. The thing is, since there is a "prefixURL" value, the user of the Helm Chart would expect since to "just work", instead of having to rewrite the entire (!) scrape_config array.

Maybe one of these days I can come up with another solution.

jacksgt:
@bot I still want to look into this.

charts/prometheus/values.yaml
the-beer-emoji: > Hmm, I'm pretty unhappy with this solution then. The thing is, since there is a "prefixURL" value, the user of the Helm Chart would expect since to "just work", instead of having to rewrite the entire (!) scrape_config array.
> 
> Maybe one of these days I can come up with another solution.

Hello,
I am testing this and I can definitivley confirm that I looked at the prefixURL and expected it to "just work". I had to search through the repo and found this open issue to confirm my suspicion that it does not.

monotek (changes requested):
please fix ci

godinja:
@jacksgt did you ever figure our how to set `metrics_path` for this one specific scrape config without rewriting the entire default `scrape_configs`?

PR #815, opened 2021-04-07
prometheus-postgres-exporter - Allow readiness timeout to be configurable via config

Narayana108:

**What this PR does**
Adds `readinessProbe.timeoutSeconds` to be configurable via values.yaml file. 

**Why**
 We are running slow SQL queries which take about 3 min to complete on postgres exporter and the pod keeps on restarting on start up, as it exceeds the default readiness timeout. By increasing this value we can run our slow queries without issues.

  - fixes #

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.9.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.2.1
+version: 2.3.0
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.9.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.2.0
+version: 2.2.1
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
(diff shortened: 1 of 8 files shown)


zanhsieh (dismissed):

zanhsieh:
@Narayana108 
Could you sign DCO please? Thank you.

Narayana108:
@zanhsieh Done :)

charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.9.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.2.0
+version: 2.2.1
```

charts/prometheus-postgres-exporter/templates/deployment.yaml
```diff
@@ -101,6 +101,7 @@ spec:
               port: http
           readinessProbe:
             initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
+            timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
```
monotek, line 104: The k8s default of one second should be in the values.yaml

monotek (changes requested):

charts/prometheus-postgres-exporter/templates/deployment.yaml
Narayana108: Sorry im not to sure what you mean by this. Can one by default change the k8s readiness timeoutSeconds ?

monotek: You're using a variable .Values.readinessProbe.timeoutSeconds which is not defined in values.yaml.
Therefore variable would be empty by default.
The Kubernetes standard value for timeoutSeconds is 1 second.
Therefore it should be in the values.yaml like this.

monotek (changes requested):

Narayana108:
@monotek I have added the readiness timeout to values.yaml.

monotek (changes requested):
may i ask you to add the same for livenessprobe too?
so both probes can be configured the same way.

Narayana108:
@monotek Done

zanhsieh (approved):

monotek (approved):

PR #819, opened 2021-04-07
[prometheus-blackbox-exporter] Add support for hostNetwork

n0rad:
Add support for host network on blackbox

Probes can require special network setup that is provided only by the host.


rsotnychenko (approved):
@n0rad thanks for the contribution!

PR #836, opened 2021-04-12
[prometheus-operator] Extracting prometheus-operator deployment to separate chart

### LINKED ISSUE #760: Discussion - [kube-prometheus-stack] Prometheus Operator in separate chart?
I have previously worked with the original `prometheus-operator` from the Helm stable repository. When the deprecation of the stable repository was announced we created our own version based on the stable repo - before the Prometheus community charts came along.

While creating the new version we chose to split it up into more smaller charts, e.g., a Prometheus Operator chart, a kube-statemetrics chart, etc. I think this chart is taking responsibility for a lot of different things at the same time. But instead of maintaining our own chart we would rather contribute to the Prometheus community.

Have you considered splitting up this chart? And would you be interested in us contributing at least a Prometheus Operator chart - which in turn then could be used as dependency for the {{kube-prometheus-stack}}?

langecode:
As described in (#760), the `kube-prometheus-stack` incorporates quite a large number of components into the same chart. By splitting out the `prometheus-operator` it will be a more clear lifecycle, it should be possible to simplify the `kube-prometheus-stack` chart a little bit, and finally it will also be possible for users to only install the `prometheues-operator` without any other components from the `kube-prometheus-stack` chart.

Following deployment of the new chart the `kube-prometheus-stack` should be changed to depend on the `prometheus-operator` chart.

This is basically just a copy of the existing `prometheus-operator` deployment of `kube-prometheus-stack` with very few changes. This should also allow for a seamless migration later on if `kube-prometheus-stack` is changed to use this as dependency.

Fixes #760


charts/prometheus-operator/ci/certmanager-values.yaml
```diff
@@ -0,0 +1,3 @@
+admissionWebhooks:
+  certManager:
+    enabled: true
```
monotek, line 3: this should be disabled by default.
not every user will have cert-manager installed.

monotek (changes requested):

monotek:
p.s. a second maintainer would be nice.

charts/prometheus-operator/ci/certmanager-values.yaml
langecode: Yes, it is disabled by default. The `ci` files is just used by `ct` during the chart testing. It will do lint and install for every `*-values.yaml` in the `ci` directory.

However it does cause a problem currently because the Github actions spins up a kind cluster for deploying the chart. Deploying the chart with certmanager enabled but without certmanager deployed _is_ a problem - hence the failure from the Github workflow.

langecode:
> p.s. a second maintainer would be nice.

Yes for sure! I also think there is a few minor issues I would like to fix. So I just created the PR as "draft" to see if it would pass the checks.

monotek:
@bismarck @gianrubio @gkarthiks @scottrigby @vsliouniaev @Xtigyro

thoughts?

bismarck:
@langecode @monotek Sorry for the delay. I'd actually like to see this get done.  My main concern with this change is that it doesn't break existing setup.  Could you document where this diverges from what's in `kube-prometheus-stack`?  I'm guessing since it's taken me so long to review, this might need to be re-sync'd from `kube-prometheus-stack`? Could I also be added as a maintainer?

vsliouniaev:
This chart used to be composed of multiple separate charts and was combined to make it easier to maintain.

langecode:
I haven’t had any time to do work on this either. But I think I will be
able to look into this the coming weeks. We are actually maintaining our
own Prometheus-operator chart right now (because we compose a “kube-stack”
with some other components). So I would really like to see that effort go
into the community to benefit and learn from others since we do it anyway -
so it is cool that you are interested. I’ll look at the remaining issues
here and let you know when ready for a review.

monotek:
Please merge main into your branch to get the "node not ready" problem in ci fixed.
See: https://github.com/prometheus-community/helm-charts/pull/1140

langecode:
Thx, I am on vacation for the next 3 weeks then I'll take a look at promoting this from draft to PR :-)

monotek:
@langecode  
frindly ping :)

langecode:
I'll close this and create a new PR as I wrongly added "signoff" on a lot of old commits!

PR #843, opened 2021-04-14
[kube-prometheus-stack] Avoid using non-ascii char in values.yaml

BarthV:
Ascii charaters are not well supported by kustomize due to pyyaml
underlying lib limitations.

Special chars doesn't bring any value here, so removing them
from the file is probably the most simple solution here.

Signed-off-by: Barthelemy Vessemont <bvessemont@gmail.com>

- [x] Chart Version bumped


bismarck (approved):
lgtm

PR #852, opened 2021-04-15
[prometheus-mysql-exporter] Add relabelings to serviceMonitor

Mahanotrahul:
PR to add relabelings in servicemonitor
  - fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/values.yaml
+++ b/charts/prometheus-mysql-exporter/values.yaml
@@ -29,6 +29,9 @@ serviceMonitor:
   jobLabel: ""
   targetLabels: []
   podTargetLabels: []
+  relabelings:
+#   - targetLabel: instance
+#     replacement: *host
   metricRelabelings: []
 
 resources: {}
@@ -109,6 +112,7 @@ collectors: {}
 mysql:
   db: ""
   host: "localhost"
+  # host: &host "localhost"
   param: ""
   pass: "password"
   port: 3306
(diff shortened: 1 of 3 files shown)


charts/prometheus-mysql-exporter/values.yaml
```diff
@@ -109,6 +112,7 @@ collectors: {}
 mysql:
   db: ""
   host: "localhost"
+  # host: &host "localhost"
```
monotek, line 115: What is that comment for?

monotek (changes requested):

monotek:
Please sign your commits to get the dco check pass.

Juanchimienti (dismissed):
looks good

charts/prometheus-mysql-exporter/values.yaml
Mahanotrahul: Usually prometheus will have the ip of the mysql exporter in the instance label. But if we want to have the actual host name of the mysql in the instance label, we can achieve it using anchors. And so I've added it as a comment for anyone to make use of it.

monotek: Can we remove this please?
Anchors are kind of discuraged from using with helm. See: https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors

monotek (changes requested):

PR #864, opened 2021-04-19
[kube-prometheus-stack] Add prometheus.thanosService.omitClusterIP

### LINKED ISSUE #863: [kube-prometheus-stack] prometheus.thanosService.clusterIP should not be required
**Describe the bug**

`prometheus.thanosService.clusterIP` ought not be a required value

Helm Version:

```console
$ helm version --short
v3.5.3+g041ce5a
```

Kubernetes Version:

```console
$ kubectl version --short
Client Version: v1.19.9
Server Version: v1.19.9
```

**Which chart**: kube-prometheus-stack

**Which version of the chart**: 14.9.0

**What happened**:

```
UPGRADE FAILED: cannot patch "prometheus-operator-thanos-discovery" with kind Service: Service "prometheus-operator-thanos-discovery" is invalid: spec.clusterIP: Invalid value: "": field is immutable
```

**What you expected to happen**: successful no-op upgrade

values.yaml

```console
prometheus:
  thanosService:
    enabled: true
    clusterIP: null # or ""
    type: LoadBalancer
```

For example:

```console
helm install prometheus-operator prometheus-community/kube-prometheus-stack -upgrade --version version --values values.yaml
```

This is essentially the same problem as https://github.com/helm/charts/issues/9227 (fixed in https://github.com/helm/charts/pull/13646)

yurrriq:

Support omission of the Thanos service `clusterIP`.

  - fixes #863

Similar to https://github.com/helm/charts/pull/13646

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/serviceThanosSidecar.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/serviceThanosSidecar.yaml
@@ -16,7 +16,9 @@ metadata:
 {{- end }}
 spec:
   type: {{ .Values.prometheus.thanosService.type }}
+  {{- if not .Values.prometheus.thanosService.omitClusterIP }}
   clusterIP: {{ .Values.prometheus.thanosService.clusterIP }}
+  {{- end }}
   ports:
   - name: {{ .Values.prometheus.thanosService.portName }}
     port: {{ .Values.prometheus.thanosService.port }}
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1597,6 +1597,7 @@ prometheus:
     portName: grpc
     port: 10901
     targetPort: "grpc"
+    omitClusterIP: false
     clusterIP: "None"
 
     ## Service type
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 15.2.0
+version: 15.2.1
 appVersion: 0.47.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus


yurrriq:
rebased. need anything else from me here? this is pretty much a blocker for us

yurrriq:
rebased again.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1597,6 +1597,7 @@ prometheus:
     portName: grpc
     port: 10901
     targetPort: "grpc"
+    omitClusterIP: false
```
monotek, line 1600: Why is this needed?
Imho you could just set `clusterIP: ""` to get rid of the error.

monotek (changes requested):

charts/kube-prometheus-stack/values.yaml
yurrriq: I'll try that and report back.

yurrriq: Same result.
```
UPGRADE FAILED: cannot patch "prometheus-operator-thanos-discovery" with kind Service: Service "prometheus-operator-thanos-discovery" is invalid: spec.clusterIP: Invalid value: "": field is immutable
```

charts/kube-prometheus-stack/templates/prometheus/serviceThanosSidecar.yaml
```diff
@@ -16,7 +16,9 @@ metadata:
 {{- end }}
 spec:
   type: {{ .Values.prometheus.thanosService.type }}
+  {{- if not .Values.prometheus.thanosService.omitClusterIP }}
   clusterIP: {{ .Values.prometheus.thanosService.clusterIP }}
```
yurrriq, line 20: As discussed in the linked tickets (see links in #863) regarding the nginx chart, the ideal solution here would be something like this, but if we want to maintain backward compatibility, we'll probably need something like `omitClusterIP`.

```suggestion
  {{- with .Values.prometheus.thanosService.clusterIP }}
  clusterIP: {{ . }}
  {{- end }}
```
```

doug-ba:
I don't know that this is the correct solution for #863 .  This service is intended as a headless service for service discovery.  I've made an alternate PR in https://github.com/prometheus-community/helm-charts/pull/916 which creates a new service for external  access and does not set ClusterIP.

yurrriq:
Superseded by #916

PR #865, opened 2021-04-19
[prometheus-blackbox-exporter] feat(blackbox-exporter): implement host aliases

secustor:
This PR adds the possibility to inject host aliases into the pods.     
We use this if adding the hosts to the DNS infrastructure is not possible or not desired. 

Further this improves the readability of metrics for this cases. 

#### Checklist

- [x] Chart Version bumped


charts/prometheus-blackbox-exporter/templates/daemonset.yaml
```diff
@@ -48,7 +48,16 @@ spec:
         - name: {{ . }}
       {{- end }}
     {{- end }}
-
+    {{- if .Values.hostAliases }}
```
desaintmartin, line 51: Could you move the indentation of those lines to respect the implicit convention ?

desaintmartin (changes requested):
Thanks ! Could you add / edit a test in the ci directory ?

charts/prometheus-blackbox-exporter/templates/daemonset.yaml
secustor: I have gone simply by the other definitions above, but I have changed now the indention.

desaintmartin: Thanks! Sorry about that, so we have something non-logical, I'll take care of it someday.

desaintmartin (approved):

PR #885, opened 2021-04-23
[kube-prometheus-stack]: fix Capabilities.APIVersions.Has check for ingress API version

ArchiFleKs:
#### What this PR does / why we need it:

fix Capabilities.APIVersions.Has check for ingress API version. Latest chart version fails to install on Kubernetes v1.18 which is kind of stuck in the middle of the v1beta1 et v1 networking API

#### Which issue this PR fixes
  - fixes #884 

#### Checklist

- [X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #900, opened 2021-04-27
#899 - correct test for networking.k8s.io/v1

### LINKED ISSUE #899: [kube-prometheus-stack] Wrong test for "networking.k8s.io/v1"
**Describe the bug**
ingress chart test capability to set the version of Ingress
Unfortunately, this returns always networking.k8s.io/v1

Helm Version: 3 5.3
K8s version : v1.18.15+vmware.1

**Which chart**:
kube-prometheus-stack
Template : prometheus/ingress.yaml
15.2.1

Detect version of ingres

Test should be : 
` {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}`

mxmorin:
Change test 
`{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}`
to
`{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}`

  - fixes #899

### DIFF
--- a/charts/kube-prometheus-stack/templates/alertmanager/ingress.yaml
+++ b/charts/kube-prometheus-stack/templates/alertmanager/ingress.yaml
@@ -4,9 +4,9 @@
 {{- $servicePort := .Values.alertmanager.service.port -}}
 {{- $routePrefix := list .Values.alertmanager.alertmanagerSpec.routePrefix }}
 {{- $paths := .Values.alertmanager.ingress.paths | default $routePrefix -}}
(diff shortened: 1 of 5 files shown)


bismarck:
I believe this was fixed in #885, #911.

DirtyCajunRice:
@bismarck this reverted 100% of the progress made over the past few days :(. The explanation was given in the first MR on why the /Kind doesnt have the outcome you expected in many circumstances. Were you just doing this as a temporary measure until Another PR?

PR #911, opened 2021-04-29
[kube-prometheus-stack]: fix boolean in ingress API version evaluation

ArchiFleKs:

Issue created with https://github.com/prometheus-community/helm-charts/pull/885

#### Checklist

- [X] Chart Version bumped


JoshVee (approved):
LGTM

DirtyCajunRice:
As a thought... since im pretty sure this is not intended behavior... you could also just return "" if false in the template and not have to change it in 5 places. Thats the point of the template! Ill make an issue upstream in helm so it will be fixed by 2027

ArchiFleKs:
> As a thought... since im pretty sure this is not intended behavior... you could also just return "" if false in the template and not have to change it in 5 places. Thats the point of the template! Ill make an issue upstream in helm so it will be fixed by 2027

Alright, I don't mind doing it either way but yes your way seems simpler

bismarck (approved):
I'm going to merge this change now because the chart is broken. @ArchiFleKs @DirtyCajunRice Could one of you fix the template in a follow-up PR?

PR #927, opened 2021-05-03
[kube-state-metrics] Add: Kubernetes recommanded labels

nlamirault:
Next of : https://github.com/kubernetes/kube-state-metrics/pull/1387

Add Kubernetes recommanded labels https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

```yaml
---
# Source: kube-state-metrics/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:    
    app.kubernetes.io/name: kube-state-metrics
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/version: "3.0.1"
    app.kubernetes.io/part-of: kube-state-metrics 
    helm.sh/chart: kube-state-metrics-3.0.1
  name: RELEASE-NAME-kube-state-metrics
  namespace: default
imagePullSecrets:
  []
---
# Source: kube-state-metrics/templates/role.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:    
    app.kubernetes.io/name: kube-state-metrics
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/version: "3.0.1"
    app.kubernetes.io/part-of: kube-state-metrics 
    helm.sh/chart: kube-state-metrics-3.0.1
  name: RELEASE-NAME-kube-state-metrics
rules:

- apiGroups: ["certificates.k8s.io"]
  resources:
  - certificatesigningrequests
  verbs: ["list", "watch"]

- apiGroups: [""]
  resources:
  - configmaps
  verbs: ["list", "watch"]

- apiGroups: ["batch"]
  resources:
  - cronjobs
  verbs: ["list", "watch"]

- apiGroups: ["extensions", "apps"]
  resources:
  - daemonsets
  verbs: ["list", "watch"]

- apiGroups: ["extensions", "apps"]
  resources:
  - deployments
  verbs: ["list", "watch"]

- apiGroups: [""]
  resources:
  - endpoints
  verbs: ["list", "watch"]

- apiGroups: ["autoscaling"]
  resources:
  - horizontalpodautoscalers
  verbs: ["list", "watch"]

- apiGroups: ["extensions", "networking.k8s.io"]
  resources:
  - ingresses
  verbs: ["list", "watch"]

- apiGroups: ["batch"]
  resources:
  - jobs
[...]

PR #930, opened 2021-05-03
[kube-state-metrics] Add the customLabels output to all resources

jaredclarke:
This adds the customLabels output to every resource so that we can add our in-house standard labels to every resource to enable validation / testing using OPA as part of CI/CD pipelines and admission webhooks

No issue

I have re-used customLabels but I would also be happy to have different params for different groups of resources e.g. customRbacLabels (and so on)

#### Checklist

- [x] Chart Version bumped


monotek:
As #927 seems to be stale i would go on with this one.

tariq1890:
@jaredclarke Would you be able to define the template partial for rendering the label as was done in #927 ?

jaredclarke:
I will take a look and add this in

PR #949, opened 2021-05-08
[prometheus-postgres-exporter] Update/prometheus postgres exporter value file to allow either password or passwordSecret

zorro1992:
Signed-off-by: Raghunandana S K <skraghunandan11@gmail.com>

When using passwordSecret with the current default values.yaml the users have to mention both ***password*** and passwordSecret  in custom values.yaml they pass during deployment , because the default values.yaml contains ```password : somepassword. ``` . 
Which means we have to override this using the custom values file and requires an unnecessary mentioning of ***password*** as an empty value. Only then the deployment works because of the condition [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-postgres-exporter/templates/deployment.yaml#L2) 

This PR should  addresses this by removing the default value for ***password*** and leaving it empty. This allows the users to either have password or passwordSecret mentioned in the custom values.yaml they pass.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.0.1
+version: 3.0.2
 appVersion: 2.0.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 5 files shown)


zanhsieh (approved):

zanhsieh:
@zorro1992 
I think you miss something on deployment.yaml and podsecuritypolicy.yaml that cause "Lint and Test Charts" below failure. Please check the "Details" next to "Lint and Test Charts".

PR #953, opened 2021-05-10
[kube-state-metrics] Fix rbac.useClusterRole=false behavior

mac-chaffee:

Looks like there is a bug affecting kube-state-metrics if a user sets `useClusterRole=false` and `namespace=<namespace1>,<namespace2>`. When a user sets those two values, they expect multiple `Role`s to be created (one per namespace), but instead it errors:

```
$ cd charts/kube-state-metrics/
$ helm template --debug --set namespaces="ns1\,ns2" --set namespace="ns1\,ns2" --set rbac.useClusterRole=false .
install.go:172: [debug] Original chart version: ""
install.go:189: [debug] CHART PATH: /Users/machaffe/personal/mac-prom-helm/charts/kube-state-metrics

---
# Source: kube-state-metrics/templates/role.yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    app.kubernetes.io/name: kube-state-metrics
    helm.sh/chart: kube-state-metrics-3.0.2
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: RELEASE-NAME
  name: RELEASE-NAME-kube-state-metrics
  namespace: map[Capabilities:0x35345c0 Chart:0xc000347e60 Files:map[.helmignore:[35 32 80 97 116
  ...
...
Error: YAML parse error on kube-state-metrics/templates/role.yaml: error converting YAML to JSON: yaml: line 10: mapping values are not allowed in this context
helm.go:81: [debug] error converting YAML to JSON: yaml: line 10: mapping values are not allowed in this context
```
This PR fixes that bug by just copying the existing logic from [rolebinding.yaml:2](https://github.com/prometheus-community/helm-charts/blob/d6ae19c6bb7089bd0080952ca6f98b68177e3dcb/charts/kube-state-metrics/templates/rolebinding.yaml#L2)

  - 

Note I renamed `namespace` to `namespaces` to match rolebinding.yaml.

Also note that I set a default value of `""`  for `namespaces`. This is needed because `namespaces` must be a string in order to apply `split` to it, even when the user doesn't override it.

I bumped the minor version instead of the patch version since this may break for users who set `useClusterRole=false` but set `namespace=some-single-namespace`. This was never documented behavior, but they will have to migrate to using `namespaces` instead of `namespace`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.1.2
+version: 3.1.0
 appVersion: 2.0.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.0.2
+version: 3.1.2
 appVersion: 2.0.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/role.yaml
+++ b/charts/kube-state-metrics/templates/role.yaml
@@ -1,11 +1,8 @@
-{{- if and (eq $.Values.rbac.create true) (not .Values.rbac.useExistingRole) -}}
-{{- if eq .Values.rbac.useClusterRole false }}
-{{- range (split "," $.Values.namespace) }}
-{{- end }}
-{{- end -}}
+{{- if and (eq .Values.rbac.create true) (not .Values.rbac.useExistingRole) -}}
+{{- range (split "," .Values.namespaces) }}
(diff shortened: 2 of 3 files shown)


jaanhio:
hi @mrueg & @tariq1890 , is it possible to provide a rough timeline as to when this can be reviewed and possibly merged? 🙏🏻

nc-kfl:
Thanks for the great work. Could be great it this bug, which this PE is fixing could be merged soon 👏🏼👏🏼

charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.0.2
+version: 3.1.2
```
mrueg, line 10: can you change this to 3.1.0?

mac-chaffee: Silly me, good catch!

mrueg (approved):

PR #963, opened 2021-05-12
[alertmanager] Fixing indentation for node selector keys in case persistence is enabled

vedharish:
Also, fixing syntax error related to configMapReload resource declaration

1. When alertmanager chart is used with `nodeSelector` `affinity` or `tolerations` along with `persistence.enabled = True`, the chart deployment fails due to indentation issues. These 3 keys are supposed to be a part of `spec.template.spec` whereas `volumeClaimTemplates` should be a child of `spec`. Fixed this by re-ordering.

2. For `configmapReload.resources`, fixed an indentation issue by chomping whitespace using `{{-`. Also used `nindent` to be consistent with the default container. [Check](https://github.com/prometheus-community/helm-charts/blob/079bcdf9d8d28a30aeb42b22027f6c298dcdd903/charts/alertmanager/templates/statefulset.yaml#L100)

NA

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.10.0
+version: 0.10.1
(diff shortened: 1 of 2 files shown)


naseemkullah (approved):
Thanks!

PR #967, opened 2021-05-12
[kube-state-metrics] Fixes template error in kube-state-metrics' role.yaml:

khalsa-ji:
* Segregates role and clusterrole.
* Defines rules in a common place (`_permissions.tpl`) as both role and clusterrole defines the same set of permissions.
* Minor: Bumps the chart version.

**Problem:** If `rbac.useClusterRole` is set to false i.e. the user wants to use rolebinding. The following error is reported while installing the helm chart:
```
Error: YAML parse error on kube-state-metrics/templates/role.yaml: error converting YAML to JSON: yaml: line 10: mapping values are not allowed in this context
```

  - Fixes https://github.com/kubernetes/kube-state-metrics/issues/1452

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.0.2
+version: 3.0.3
 appVersion: 2.0.0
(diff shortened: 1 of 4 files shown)


nc-kfl:
Clone of #953.

monotek:
fixed with: #953

PR #982, opened 2021-05-19
[prometheus-adapter] Add hostPorts in psp

tvm18860:

When trying to install prometheus-adapter into our EKS cluster with Calico CNI and rbac, psp, and serviceaccount enabled, we hit an issue that the ReplicaSet was unable to spin up a pod due to the following error:

```
 Warning  FailedCreate  106s (x15 over 3m8s)  replicaset-controller  Error creating: pods "prometheus-adapter-85684bb67d-" is forbidden: unable to validate against any pod security policy: [spec.containers[0].hostPort: Invalid value: 6443: Host port 6443 is not allowed to be used. Allowed ports: []]
```

We did set `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet` based on some notes in https://github.com/prometheus-community/helm-charts/pull/375, but our cluster does not allow the use of hostPorts by default. 

This PR adds the optional ability to include the `hostPorts` section in the psp to set it to only allow `.Values.listenPort`. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/psp.yaml
+++ b/charts/prometheus-adapter/templates/psp.yaml
@@ -12,8 +12,6 @@ metadata:
 spec:
   {{- if .Values.hostNetwork.enabled }}
   hostNetwork: true
-  {{- end }}
-  {{- if .Values.hostNetwork.hostPortsEnabled }}
   hostPorts:
     - min: {{ .Values.listenPort }}
       max: {{ .Values.listenPort }}
--- a/charts/prometheus-adapter/templates/psp.yaml
+++ b/charts/prometheus-adapter/templates/psp.yaml
@@ -13,6 +13,11 @@ spec:
   {{- if .Values.hostNetwork.enabled }}
   hostNetwork: true
   {{- end }}
+  {{- if .Values.hostNetwork.hostPortsEnabled }}
+  hostPorts:
+    - min: {{ .Values.listenPort }}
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
new feature should rais minor version

charts/prometheus-adapter/templates/psp.yaml
```diff
@@ -13,6 +13,11 @@ spec:
   {{- if .Values.hostNetwork.enabled }}
   hostNetwork: true
   {{- end }}
+  {{- if .Values.hostNetwork.hostPortsEnabled }}
+  hostPorts:
+    - min: {{ .Values.listenPort }}
+      max: {{ .Values.listenPort }}
+  {{- end }}
```
steven-sheehy, line 20: Do we need this option? Since the container ports can't be customized if you set `hostNetwork=true` it will implicitly use hostPorts for the listen port. Can't we just always set hostPorts to listenPort if hostnetwork is true?
```suggestion
  {{- if .Values.hostNetwork.enabled }}
  hostNetwork: true
  hostPorts:
    - min: {{ .Values.listenPort }}
      max: {{ .Values.listenPort }}
  {{- end }}
```

steven-sheehy (approved):
LGTM

monotek (approved):

PR #986, opened 2021-05-20
[prometheus] Add option to set --storage.tsdb.path

vvanouytsel:
Previously the value of '--storage.tsdb.path' was set via '.Values.server.persistentVolume.mountPath'.
This commit adds the option to specify a 'server.hostPath' value. When
the 'server.hostPath' value is not empty, the 'server.hostPath' value is
used for '--storage.tsdb.path' instead.

Signed-off-by: Vincent Van Ouytsel <vvanouytsel@gmail.com>

Make it possible to configure '--storage.tsdb.path' directly via the values file. 
Previously it was 'hardcoded' to use .Values.server.persistentVolume.mountPath as the value of '--storage.tsdb.path'.
With this PR you can now overrule this functionality and set the '--storage.tsdb.path' directly using the 'storagePath' variable.

n/a

n/a

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: 2.26.0
-version: 14.0.1
+version: 14.1.0
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
 icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/assets/prometheus_logo-cb55bb5c346.png
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -671,6 +671,10 @@ server:
   ## Path to a configuration file on prometheus server container FS
   configPath: /etc/config/prometheus.yml
 
+  ### The data directory used by prometheus to set --storage.tsdb.path
+  ### When empty server.persistentVolume.mountPath is used instead
+  storagePath: ""
+
   global:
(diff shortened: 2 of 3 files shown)


charts/prometheus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: 2.26.0
-version: 14.0.0
+version: 14.0.1
```
monotek, line 4: new feature should raise minor version

monotek (changes requested):

charts/prometheus/Chart.yaml
vvanouytsel: Bumped version to 14.1.0

charts/prometheus/values.yaml
```diff
@@ -671,6 +671,10 @@ server:
   ## Path to a configuration file on prometheus server container FS
   configPath: /etc/config/prometheus.yml
 
+  ### The data directory used by prometheus to set --storage.tsdb.path
+  ### When empty server.persistentVolume.mountPath is used instead
+  # storagePath: ""
```
monotek, line 676: could you please uncomment the example value?

monotek (changes requested):

charts/prometheus/values.yaml
vvanouytsel: done!

zanhsieh (approved):

monotek (approved):

PR #992, opened 2021-05-21
[prometheus-postgres-exporter] Don't apply Secret when using one externally-managed

jonathan-mothership:
Adds a validation on `.Values.config.datasourceSecret` to avoid laying down a secret that isn't used. 

When the operator of this chart specifies `.Values.config.datasourceSecret` (commented as being mutually exclusive to `.Values.config.datasource`), the `Secret` is applied regardless. The `Deployment` uses the `datasourceSecret` exclusively when it's defined, so the `Secret` is never mounted and serves no purpose. 
 
I skipped a patch on the chart bump in light of #991.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.9.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.3.1
+version: 2.3.3
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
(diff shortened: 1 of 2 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.9.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.3.1
+version: 2.3.3
```
monotek, line 5: this should be 2.3.2, or not?

monotek (changes requested):

monotek (approved):

zanhsieh (approved):

PR #1017, opened 2021-05-28
[kube-state-metrics] Add containerSecurityContext

lorenzo-biava:
Signed-off-by: Biava, Lorenzo <lorenzo.biava@nuance.com>

Add parameter containerSecurityContext to add a securityContext to a container. To allow setting options not available at the Pod level securityContext.
See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#securitycontext-v1-core

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -195,6 +195,10 @@ spec:
         resources:
 {{ toYaml .Values.resources | indent 10 }}
 {{- end }}
+{{- if .Values.containerSecurityContext }}
+        securityContext:
+{{ toYaml .Values.containerSecurityContext | indent 10 }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


monotek (approved):

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -195,6 +195,9 @@ spec:
         resources:
 {{ toYaml .Values.resources | indent 10 }}
 {{- end }}
+{{- with .Values.containerSecurityContext }}
```
mrueg, line 198: This uses a different style as above (if vs. with). Ideally we wouldn't mix those.

lorenzo-biava: makes sense; updated with the if style

mrueg (approved):

PR #1037, opened 2021-06-02
[prometheus-statsd-exporter] corret ingress render in k8s 1.19 above

LeoQuote:

when installing chart with `helm install test-statsd prom/prometheus-statsd-exporter --values ingress.enabled=true` in a k8s 1.19 cluster , things will go wrong 

```
Error: UPGRADE FAILED: template: prometheus-statsd-exporter/templates/ingress.yaml:38:32: executing "prometheus-statsd-exporter/templates/ingress.yaml" at <.Values.ingress.pathType>: can't evaluate field Values in type interface {}
``` 
the reason is that the `.Values` varible is not available in a loop. assign it in the header can fix this

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-statsd-exporter/templates/ingress.yaml
+++ b/charts/prometheus-statsd-exporter/templates/ingress.yaml
@@ -2,6 +2,7 @@
 {{- $fullName := include "prometheus-statsd-exporter.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+{{- $pathType := .Values.ingress.pathType -}}
 {{- if semverCompare ">=1.19.0-0" $kubeTargetVersion }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
@@ -35,7 +36,7 @@ spec:
           {{- range .paths }}
           - path: {{ . }}
             {{- if semverCompare ">=1.19.0-0" $kubeTargetVersion }}
-            pathType: {{ .Values.ingress.pathType | default "ImplementationSpecific" }}
+            pathType: {{ $pathType | default "ImplementationSpecific" }}
             {{- end }}
             backend:
               {{- if semverCompare ">=1.19.0-0" $kubeTargetVersion }}
--- a/charts/prometheus-statsd-exporter/templates/ingress.yaml
+++ b/charts/prometheus-statsd-exporter/templates/ingress.yaml
@@ -2,7 +2,6 @@
 {{- $fullName := include "prometheus-statsd-exporter.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
-{{- $pathType := .Values.ingress.pathType -}}
 {{- if semverCompare ">=1.19.0-0" $kubeTargetVersion }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-statsd-exporter/values.yaml
```diff
@@ -85,9 +85,9 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
+  # pathType: ImplementationSpecific # only Kubernetes v1.19+
```
monotek, line 88: Imho wron place for this. The ingress default config for a helm 3.6.0 created shard would look like:

```
ingress:
  enabled: false
  className: ""
  annotations: {}
    # kubernetes.io/ingress.class: nginx
    # kubernetes.io/tls-acme: "true"
  hosts:
    - host: chart-example.local
      paths:
        - path: /
          pathType: ImplementationSpecific
  tls: []
  #  - secretName: chart-example-tls
  #    hosts:
  #      - chart-example.local
```

charts/prometheus-statsd-exporter/templates/ingress.yaml
```diff
@@ -35,7 +36,7 @@ spec:
           {{- range .paths }}
           - path: {{ . }}
             {{- if semverCompare ">=1.19.0-0" $kubeTargetVersion }}
-            pathType: {{ .Values.ingress.pathType | default "ImplementationSpecific" }}
+            pathType: {{ $pathType | default "ImplementationSpecific" }}
```
monotek, line 39: Imho helm 3.6.0 default ingress could be used here too:

```
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "prometheus-statsd-exporter.fullname" . -}}
{{- $svcPort := .Values.service.port -}}
{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
  {{- end }}
{{- end }}
{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
apiVersion: networking.k8s.io/v1
{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
apiVersion: networking.k8s.io/v1beta1
{{- else -}}
apiVersion: extensions/v1beta1
{{- end }}
kind: Ingress
metadata:
  name: {{ $fullName }}
  labels:
    {{- include "prometheus-statsd-exporter.labels" . | nindent 4 }}
  {{- with .Values.ingress.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
  ingressClassName: {{ .Values.ingress.className }}
  {{- end }}
  {{- if .Values.ingress.tls }}
  tls:
    {{- range .Values.ingress.tls }}
    - hosts:
        {{- range .hosts }}
        - {{ . | quote }}
        {{- end }}
      secretName: {{ .secretName }}
    {{- end }}
  {{- end }}
  rules:
    {{- range .Values.ingress.hosts }}
    - host: {{ .host | quote }}
      http:
        paths:
          {{- range .paths }}
          - path: {{ .path }}
            {{- if and .pathType (semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion) }}
            pathType: {{ .pathType }}
            {{- end }}
            backend:
              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
              service:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
              {{- else }}
              serviceName: {{ $fullName }}
              servicePort: {{ $svcPort }}
              {{- end }}
          {{- end }}
    {{- end }}
{{- end }}

```

monotek (changes requested):

charts/prometheus-statsd-exporter/templates/ingress.yaml
LeoQuote: I believe it is fixed now

monotek (changes requested):
Imho we should also add ingressclassname.
You could use "helm create prometheus-statsd-exporter" with helm 3.6.0, to get an updated ingress template.

nikopavlica:
Can this be merged please? Right now this chart is broken for kubernetes >= 1.19

monotek:
I could not push to this branch so i created a new pr, adding also classname: https://github.com/prometheus-community/helm-charts/pull/1355

PR #1048, opened 2021-06-07
Following prometheus-operator change #860

levi-pl:
Service kube-prometheus-stack-kubelet created by prom-operator does not have label  "app.kubernetes.io/managed-by: prometheus-operator". Instead it has recommended label "app.kubernetes.io/name: kubelet"

https://github.com/prometheus-operator/kube-prometheus/issues/860


bismarck:
@levi-pl Thanks for the contribution. You need to bump the chart version.

bismarck (approved):
lgtm

PR #1060, opened 2021-06-10
[kube-prometheus-stack] Add Extra Support for Amazon EKS KubeVersion on Ingress

### LINKED ISSUE #1057: [prometheus-kube-stack] wrong result with semverCompare for ingress apiversion
**Describe the bug**

semverCompare thinks "1.19.x" > "v1.20.4-eks-xxxxxx".

Helm Version:

```console
$ helm version
version.BuildInfo{Version:"v3.6.0", GitCommit:"7f2df6467771a75f5646b7f12afb408590ed1755", GitTreeState:"dirty", GoVersion:"go1.16.4"}
```

Kubernetes Version:

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.4-eks-6b7464", GitCommit:"6b746440c04cb81db4426842b4ae65c3f7035e53", GitTreeState:"clean", BuildDate:"2021-03-19T19:33:03Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**:
prometheus-kube-stack

16.5.0

**What happened**:

We use kubernetes 1.20.4. But we see that ingress of alertmanager and prometheus use the `apiversion: networking.k8s.io/v1beta1`

We want to see the apiversion of ingress to be `apiversion: networking.k8s.io/v1`

Create a configmap: 
```apiVersion: v1
data:
  kubeconfig: |-
    {{ semverCompare ">= 1.19.0-0" ( "v1.20.4-eks-test" )  }}
    {{ semverCompare ">= 1.19.0-0" ( "v1.20.4" )  }}
    {{ semverCompare ">= 1.19.0-0" ( "1.20.4" )  }}
    {{ semverCompare ">= 1.19.x" ( "v1.20.4-eks-test" )  }}
    {{ semverCompare ">= 1.19.x" ( "v1.20.4" )  }}
    {{ semverCompare ">= 1.19.x" ("1.20.4" )  }}
kind: ConfigMap
metadata:
  name: test
```

and run `helm template .`

See the result: 
```
---
# Source: test/templates/test.yaml
apiVersion: v1
data:
  kubeconfig: |-
    true
    true
    true
    false
    true
    true
kind: ConfigMap
metadata:
  name: test
```
I'm not sure, wether it is a bug of semverCompare or not. It can be fixed with a change from `1.19.x` to `1.19.0-0` at this line: https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-16.5.0/charts/kube-prometheus-stack/templates/_helpers.tpl#L106 and similarly this line: https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-16.5.0/charts/kube-prometheus-stack/templates/_helpers.tpl#L123

Kirchen99:
Signed-off-by: Kirchen99 <latias_latios@126.com>

This PR adds support when this Chart is deployed on an Amazon EKS Cluster. 

The Kube-Version of an EKS Cluster is like a pre-release version e.g. `v1.20.4-eks-6b7464`, which makes helm confused thus rendered wrong apiVersion for resources like ingress of Alertmanager and Prometheus.

This PR, for example, replaces `v1.20.4-eks-6b7464` with `1.20.4`, if there is an `eks` in it.

  - fixes #1057

#### Checklist
- [x] Chart Version bumped


monotek:
Why is this needed?
Does kubeVersionOverride not work for you?

stevehipwell:
I missed this PR when I opened issue https://github.com/prometheus-community/helm-charts/issues/1071 but I'm in agreement that this is needed.

Are there any other managed K8s distros that also misuse the `KubeVersion` field that should also be included in this?

stevehipwell:
@monotek `kubeVersionOverride` works but firstly you need to know to use it and secondly it's a static value that would need to be maintained over a cluster lifetime and could cause issues if not updated correctly. This fix is a significant improvement to the UX for this chart.

z0rc:
Actually it isn't EKS specific, other providers like GKE and microk8s do something similar with server version. Check out https://github.com/vmware-tanzu/sonobuoy/pull/1228 to see how other projects deal provider build strings.

Kirchen99:
> Why is this needed?
> Does kubeVersionOverride not work for you?

`kubeVersionOverride` works functional but we want better user experience.

Xtigyro (changes requested):
Thank you for the contribution, @Kirchen99! 🥇 

Could you please resolve the merge conflicts?

Xtigyro:
@Kirchen99 Also - please bump the minor version of the chart (not the patch version).

charts/kube-prometheus-stack/templates/_helpers.tpl
```diff
@@ -98,7 +98,12 @@ Allow the release namespace to be overridden for multi-namespace deployments in
 
 {{/* Allow KubeVersion to be overridden. */}}
 {{- define "kube-prometheus-stack.ingress.kubeVersion" -}}
-  {{- default .Capabilities.KubeVersion.Version .Values.kubeVersionOverride -}}
+  {{- $kubeVersion := default .Capabilities.KubeVersion.Version .Values.kubeVersionOverride -}}
+  {{/* Special use case for Aamzon EKS */}}
```
z0rc, line 102: `Aamzon`

z0rc (changes requested):

charts/kube-prometheus-stack/templates/_helpers.tpl
Xtigyro: @Kirchen99 This one too.

stevehipwell:
@z0rc it's a shame Helm doesn't have lookahead assertions (negative) so we could implement a catch all for non SemVer pre-releases (e.g. keep `alpha`, `beta`, `rc` and trim others). I was thinking an option would be to check for the `.x` pre-release version but I don't know how consistently this is used in practice so I think what we've got here looks to be the best solution.

z0rc:
@stevehipwell @Kirchen99 At least maybe proposed regex can be improved to handle more cases (GKE, microk8s, see test cases from linked sonobuoy PR), instead of being EKS specific?

stevehipwell:
@z0rc that'd make sense if the regex patterns would be compatible (e.g. `$\\d+\\.\\d+\\.\\d+-(?:eks|gke).+$`).

Kirchen99:
This one is closed since I have tried to resolve the conflicts by git rebase. But it brings more problems(DCO) to solve. So I open a new pull request https://github.com/prometheus-community/helm-charts/pull/1083.

PR #1109, opened 2021-06-24
[kube-prometheus-stack] Default to auto-generate Grafana admin password

steven-sheehy:
 The kube-prometheus-stack chart still hardcodes a Grafana admin password of `prom-operator`. The Grafana chart added [support](https://github.com/grafana/helm-charts/pull/62) for auto generating the admin password on install and using [lookup](https://github.com/grafana/helm-charts/blob/2d6bfcc7d2e29cb428f26fb6e05b9361bd885873/charts/grafana/templates/_helpers.tpl#L112) so it doesn't change on upgrades. We should utilize this feature to provide a more secure default for new installs of the chart.

This change does not impact users who manually set a `grafana.adminPassword` or users who left the previous chart default and do an upgrade. In the latter case it will simply lookup `prom-operator` from the existing secret.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 16.11.0
+version: 16.12.0
 appVersion: 0.48.1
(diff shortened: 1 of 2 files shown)


Xtigyro:
There is another PR about this. Will send it here tomorrow morning and we can discuss it more.

Xtigyro:
https://github.com/prometheus-community/helm-charts/pull/796

PR #1147, opened 2021-07-09
[prometheus-adapter] Allow to specify deployment strategy

randrusiak:
Signed-off-by: randrusiak <randrusiak@protonmail.com>

Users should be able to specify a custom deployment strategy. It may be useful especially if you use hostNetwork and pod with promethues-adapter is redeployed on the same worker node. With the default "RollingUpdate" strategy new pod won't be deployed because node port will be still allocated. So if we use hostNetwork we may specify "recreate" strategy which solves this problem. 

Default values are the same as k8s provides. 
n/a
n/a
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/deployment.yaml
+++ b/charts/prometheus-adapter/templates/deployment.yaml
@@ -10,6 +10,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   replicas: {{ .Values.replicas }}
+  strategy: {{ toYaml .Values.strategy | nindent 4 }}
   selector:
     matchLabels:
       app: {{ template "k8s-prometheus-adapter.name" . }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   replicas: {{ .Values.replicas }}
+  strategy:
+{{ toYaml .Values.strategy | trim | indent 4 }}
```
steven-sheehy, line 14: nit: I prefer using nindent and the more succinct single line approach.
```suggestion
  strategy: {{ toYaml .Values.strategy | nindent 4 }}
```

randrusiak: Sure, it's definitely better approach. Thanks

steven-sheehy (approved):
LGTM

PR #1176, opened 2021-07-19
[prometheus-node-exporter] Version bump to 1.2.0; drop helm2 support

mrueg:
* Bumps node exporter to v1.2.0
* Drops support for helm2 by upgrading to Chart apiVersion v2.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -1,8 +1,9 @@
-apiVersion: v1
-appVersion: 1.1.2
+apiVersion: v2
+appVersion: 1.2.0
 description: A Helm chart for prometheus node-exporter
 name: prometheus-node-exporter
-version: 1.18.2
+version: 2.0.0
+type: application
 home: https://github.com/prometheus/node_exporter/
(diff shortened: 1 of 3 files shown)


charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -1,8 +1,9 @@
-apiVersion: v1
-appVersion: 1.1.2
+apiVersion: v2
+appVersion: 1.2.0
 description: A Helm chart for prometheus node-exporter
 name: prometheus-node-exporter
-version: 1.18.2
+version: 1.19.0
```
mrueg, line 5: Can move this to 2.0.0 as well if desired.

vsliouniaev: Since we're breaking backwards-compatibility with Helm v2 it would be a good idea I think! 👍

mrueg: Will do!

monotek (approved):

vsliouniaev (approved):

PR #1181, opened 2021-07-19
[kube-state-metrics] Disable VPA by default

### LINKED ISSUE #1131: [kube-state-metrics] Collectors has optional resource verticalpodautoscalers as a default
**Describe the bug**
The kube-state-metrics helm chart includes a collector for `verticalpodautoscalers` which is an optional resource and not installed in a cluster by default. This causes a large number of errors to be logged. This is compounded by the way the collectors are configured in an array meaning if you wish to turn this off you need to maintain the list of collectors and not get updates from the chart.

Helm Version:
n/a

Kubernetes Version:
v1.18, v1.19, v1.20

**Which chart**:
kube-state-metrics

v3.3.1

**What happened**:
A large number of errors.

> E0702 09:40:01.750260       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.5/tools/cache/reflector.go:167: Failed to watch *v1beta2.VerticalPodAutoscaler: failed to list *v1beta2.VerticalPodAutoscaler: the server could not find the requested resource (get verticalpodautoscalers.autoscaling.k8s.io)

Install chart into bare cluster and watch the logs.

n/a

n/a

**Helm values set after installation/upgrade:**

n/a

I'd suggest removing `verticalpodautoscalers` from the `collectors` value and adding an `additionalCollectors` value that is concated with `collectors` in the chart to enable augmenting the default without having to manage them (`verticalpodautoscalers` could be added to this but I'd suggest not as you will get errors on a default install on a default cluster). I'd happily open a PR for this.

mrueg:
Disables VPA by default as it's not a default resource on clusters

  - fixes #1131 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -162,8 +162,8 @@ collectors:
   - statefulsets
   - storageclasses
   - validatingwebhookconfigurations
-  - verticalpodautoscalers
   - volumeattachments
+  # - verticalpodautoscalers # not a default resource, see also: https://github.com/kubernetes/kube-state-metrics#enabling-verticalpodautoscalers
 
 # Enabling kubeconfig will pass the --kubeconfig argument to the container
 kubeconfig:
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 3.4.0
+version: 3.4.1
 appVersion: 2.1.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:


charts/kube-state-metrics/values.yaml
```diff
@@ -162,7 +162,7 @@ collectors:
   - statefulsets
   - storageclasses
   - validatingwebhookconfigurations
-  - verticalpodautoscalers
+  # - verticalpodautoscalers # not a default resource
```
tariq1890, line 165: I would suggest just removing it

kfox1111:
It doesn't hurt anything by it being there?

kfox1111:
I mean, having the permission for verticalpodautoscalers doesn't hurt when the crd isn't installed. the Role I think still works fine.

charts/kube-state-metrics/values.yaml
tariq1890: https://github.com/kubernetes/kube-state-metrics#enabling-verticalpodautoscalers This is a useful link to add to the chart

mrueg:
```
E0719 16:56:58.222694       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.5/tools/cache/reflector.go:167: Failed to watch *v1beta2.VerticalPodAutoscaler: failed to list *v1beta2.VerticalPodAutoscaler: the server could not find the requested resource (get verticalpodautoscalers.autoscaling.k8s.io)
```
is what it would output instead.

charts/kube-state-metrics/values.yaml
mrueg: I added your link to the comment and moved it to the end of the values. I would not remove it since others might wonder where it is.

tariq1890 (approved):

kfox1111:
Isn't it just going to throw some other error message in that case? Like, it doesn't have permissions to that type? Or maybe the same error message, since the api doesn't exist?

mrueg:
If the VPA collector is disabled it should not throw any error message. Users who want to collect VPA metrics can enable it by overwriting the values.

PR #1193, opened 2021-07-21
[kube-prometheus-stack] allow to create secrets for basic auth usage

### LINKED ISSUE #1186: Configuration for Basic Authentication ( prometheus/alertmanager)
I have deployed kube-prometheus-stack using helm charts. The version I use is  15.1.3 .
Everything works as expected.  Now I want to enable the Basic Auth feature.
According to the official documentation, this is supported and can be enabled using the following attribute

$ prometheus -web.config.file=web.yml

Can I use this feature with kube-prometheus-stack? I want to secure the endpoints (prometheus/alertmanager).

Has anyone achieved it with the approach of using web.config.file?

thirdeyenick:

This PR allows to pass hashed credentials to the helm chart. Those credentials will be stored in a k8s secret. That k8s secret can then be referenced to support basic authentication in ingress resources. The nginx ingress controller has [that feature](https://kubernetes.github.io/ingress-nginx/examples/auth/basic/) for example.

Although Prometheus does [support basic authentication](https://prometheus.io/docs/prometheus/latest/configuration/https/) itself, the prometheus-operator project doesn't seem to allow to configure this via the [Prometheus CRD](https://github.com/prometheus-operator/prometheus-operator/blob/f103831b389be5cf504bdf28002cfa2915b821ed/pkg/apis/monitoring/v1/types.go#L516).

  - fixes #1186

#### Checklist

- [x] Chart Version bumped


gkarthiks:
Thank you for your contribution @thirdeyenick. 

Though I understand the need for this, I have the slightest hesitation on this PR. IMHO, one shouldn't do a basic auth like this. Rather the system should be integrated with an IDP. Because when done something like a secret file with the creds, wouldn't this be insecure? 

I would defer this PR for the rest of the owners @bismarck @gianrubio @scottrigby @vsliouniaev @Xtigyro

Thank you @thirdeyenick for your patience and for helping us in defining the best solution for the community.

thirdeyenick:
Hi,
> Though I understand the need for this, I have the slightest hesitation on this PR. IMHO, one shouldn't do a basic auth like this. Rather the system should be integrated with an IDP. Because when done something like a secret file with the creds, wouldn't this be insecure?

It for sure depends on the implementation of the ingress controller, which kind of credentials are needed in the k8s secret.
Nginx Ingress supports bcrypt for example, so you just push a bcrypt hashed password into the helm values. I wouldn't consider this as insecure. But for sure: using the PLAIN scheme (so username and password in plaintext) would be really insecure. It depends on the user which password scheme will be used.

monotek (changes requested):
I would suggest to go for a more general solution like "extraSecrets".

thirdeyenick:
> I would suggest to go for a more general solution like "extraSecrets".

I like the idea and did some changes. Should now be more generic.

gkarthiks (approved):
lgtm

thirdeyenick:
Anything to do for me still?

gkarthiks:
> Anything to do for me still?

nope, just waiting for @monotek approval. Since he requested a change and Github is waiting for it. I don't want to dismiss his review.

shiron-babi:
This feature is very useful, thanks for the hard work! 

Any estimation when it will be merged ?

gkarthiks (approved):

brsolomon-deloitte:
It looks like this was not documented anywhere. Would've been nice to show a minimally reproducible example.

manidharanupoju24:
Can this be used without ingress?

PR #1200, opened 2021-07-22
Add support for policy/v1, for Kubernetes 1.25+

Fodoj:
policy/v1beta1 is deprecated since 1.21 and will be removed in 1.25. This PR adds a check if policy/v1 is available (meaning any 1.21+ cluster) and uses that one, otherwise falls back to policy/v1beta.

### DIFF
--- a/charts/alertmanager/templates/pdb.yaml
+++ b/charts/alertmanager/templates/pdb.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.podDisruptionBudget -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ template "alertmanager.fullname" . }}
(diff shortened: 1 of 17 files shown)


charts/alertmanager/templates/pdb.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.podDisruptionBudget -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
```
zanhsieh, line 5: @Fodoj 
Move to ._tpl should be clearer.

zanhsieh:
@Fodoj 
Also, it's better to split to each chart PR. Make all approvers to agree to a single overall PR would prolong merging process.

monotek (changes requested):
please one pr for every chart.
the release action will not work for multiple charts.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 17.0.0
+version: 17.0.1
```
gkarthiks, line 21: needs a version bump

gkarthiks (commented):
except for the chart version, rest looks good from the kps chart perspective

Fodoj:
Took me a while to get back to this topic, but here is the first PR of a split, @gkarthiks https://github.com/prometheus-community/helm-charts/pull/1240

PR #1216, opened 2021-07-28
[prometheus] Upgrade dependent kube-state-metrics chart version + additional dependent charts

Sebor:
- *The version of `kube-state-metrics` chart has been not actual.* Upgraded
- *There were no additional dependent charts (see below) in `dependencies` section of main one.* Added

These charts are allowed to be installed:
- prometheus-node-exporter
- prometheus-pushgateway
- alertmanager

#### Checklist

- [x] Chart Version bumped


Sebor:
@gianrubio @monotek @naseemkullah @Xtigyro @zanhsieh hello! Could you loot at my PR?

zanhsieh (approved):

charts/prometheus/Chart.yaml
```diff
@@ -26,6 +26,18 @@ engine: gotpl
...
+  - name: prometheus-node-exporter
+    version: "2.0.*"
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: nodeExporter.enabled
+  - name: prometheus-pushgateway
+    version: "1.10.*"
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: pushgateway.enabled
+  - name: alertmanager
+    version: "1.12.*"
```
zanhsieh, line 41: Error: Error linting charts: Error building dependencies for chart 'prometheus => (version: "14.5.0", path: "charts/prometheus")': Error waiting for process: exit status 1
Error: can't get a valid version for repositories alertmanager. Try changing the version constraint in Chart.yaml

zanhsieh (dismissed):

charts/prometheus/Chart.yaml
Sebor: Sorry, it was typo.
Fixed

charts/prometheus/Chart.yaml
```diff
@@ -26,6 +26,18 @@ engine: gotpl
...
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: nodeExporter.enabled
+  - name: prometheus-pushgateway
+    version: "1.10.*"
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: pushgateway.enabled
+  - name: alertmanager
+    version: "0.12.*"
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: alertmanager.enabled
```
naseemkullah, line 43: While I am very much in favour with the idea of re-using the separate charts, I'm afraid this will clash with the components vendored components?

maybe a compositte condition like (pseudo code do not copy paste) `nodeExporter.enabled and nodeExporter.vendoredComponent: false` and we add a vendoredComponent that defaults to true for each component

Sebor: Hi @naseemkullah !
Could you describe your idea in more details? In my opinion single condition is more clear and simple thing

naseemkullah: > In my opinion single condition is more clear and simple thing
I agree but there are already vendored components which would clash with installing the seperate charts, no?

The idea is either deploy the vendored components or the seperate charts, but not both

Sebor: Honestly I don't understand. I've just wanted to upgrade the versions :)

> The idea is either deploy the vendored components or the separate charts, but not both

Maybe just to set ```nodeExporter.enabled/pushgateway.enabled/alertmanager.enabled``` as `false` by default?

Sebor:
Is it always so hard to be a new contributor for this repo?..

charts/prometheus/Chart.yaml
naseemkullah: > Honestly I don't understand. I've just wanted to upgrade the versions :)
> 

But you are not only upgrading the version of kube state metrics, you are also adding 3 subcharts that don't currently exist in the chart. Have you tried deploying this locally?

Have you tried deploying this locally? I am concerned about the clash between the added dependency charts and the corresponding components that are already in the prometheus chart.

> Maybe just to set nodeExporter.enabled/pushgateway.enabled/alertmanager.enabled as false by default?

I don't see how this would help and it would just be a breaking change for folks not explicitly enabling those components.

naseemkullah (approved):

charts/prometheus/Chart.yaml
Sebor: Finally I've understood you. You are right - for this PR kube-state chart dependency upgrade is enough

monotek (approved):

PR #1234, opened 2021-08-06
[prometheus] server: support running in a service mesh

harti2006:
I'd like to deploy the Prometheus server into a setup with Consul Connect Service Mesh. That means the main Prometheus API must not be reachable from outside the Pod directly, but only through the so-called "sidecar-proxy". This proxy enforces mutualTLS and only allows authorized traffic to reach the application. To optimally support this setup, I would like to add three changes:
- [x] Add a flag to bind the server only to localhost
- [x] Allow customization of the readiness and liveness probes, to call to the proxy instead of port 9090
- [x] Optionally disable the Kubernetes Service. It's just not needed in our case.

#### Checklist

- [x] Chart Version bumped

Can you please take a look @gianrubio @zanhsieh @Xtigyro @monotek @naseemkullah

### DIFF
--- a/charts/prometheus/templates/server/deploy.yaml
+++ b/charts/prometheus/templates/server/deploy.yaml
@@ -110,6 +110,9 @@ spec:
           {{- if .Values.server.baseURL }}
             - --web.external-url={{ .Values.server.baseURL }}
           {{- end }}
+          {{- if .Values.server.listenOnLocalhost }}
+            - --web.listen-address=localhost:9090
+          {{- end }}
           ports:
             - containerPort: 9090
           readinessProbe:
--- a/charts/prometheus/templates/server/deploy.yaml
+++ b/charts/prometheus/templates/server/deploy.yaml
@@ -116,18 +116,26 @@ spec:
           ports:
             - containerPort: 9090
           readinessProbe:
+          {{- if .Values.server.customReadinessProbe }}
+{{ toYaml .Values.server.customReadinessProbe | indent 12}}
+          {{- else }}
             httpGet:
               path: {{ .Values.server.prefixURL }}/-/ready
               port: 9090
+          {{- end }}
             initialDelaySeconds: {{ .Values.server.readinessProbeInitialDelay }}
             periodSeconds: {{ .Values.server.readinessProbePeriodSeconds }}
             timeoutSeconds: {{ .Values.server.readinessProbeTimeout }}
             failureThreshold: {{ .Values.server.readinessProbeFailureThreshold }}
             successThreshold: {{ .Values.server.readinessProbeSuccessThreshold }}
           livenessProbe:
+          {{- if .Values.server.customLivenessProbe }}
+{{ toYaml .Values.server.customLivenessProbe | indent 12}}
+          {{- else }}
             httpGet:
               path: {{ .Values.server.prefixURL }}/-/healthy
               port: 9090
+          {{- end }}
             initialDelaySeconds: {{ .Values.server.livenessProbeInitialDelay }}
             periodSeconds: {{ .Values.server.livenessProbePeriodSeconds }}
             timeoutSeconds: {{ .Values.server.livenessProbeTimeout }}
--- a/charts/prometheus/templates/server/deploy.yaml
+++ b/charts/prometheus/templates/server/deploy.yaml
@@ -110,8 +110,8 @@ spec:
           {{- if .Values.server.baseURL }}
             - --web.external-url={{ .Values.server.baseURL }}
           {{- end }}
-          {{- if .Values.server.listenOnLocalhost }}
-            - --web.listen-address=localhost:9090
+          {{- if .Values.server.listenOnHost }}
+            - --web.listen-address={{ .Values.server.listenOnHost }}:9090
           {{- end }}
           ports:
             - containerPort: 9090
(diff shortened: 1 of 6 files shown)


zanhsieh (approved):

Xtigyro (approved):

harti2006:
@prometheus-community/helm-charts-admins can you please also check this PR? I added a small addition to the `.gitignore` file.

charts/prometheus/templates/server/deploy.yaml
```diff
@@ -110,21 +110,32 @@ spec:
           {{- if .Values.server.baseURL }}
             - --web.external-url={{ .Values.server.baseURL }}
           {{- end }}
+          {{- if .Values.server.listenOnLocalhost }}
```
monotek, line 113: Imho if we make use of "--web.listen-address" it should be configurable and not bound to localhost only.

charts/prometheus/templates/server/sts.yaml
```diff
@@ -110,21 +110,32 @@ spec:
           {{- if .Values.server.baseURL }}
             - --web.external-url={{ .Values.server.baseURL }}
           {{- end }}
+          {{- if .Values.server.listenOnLocalhost }}
+            - --web.listen-address=localhost:9090
```
monotek, line 114: Imho if we make use of "--web.listen-address" it should be configurable and not bound to localhost only.

charts/prometheus/values.yaml
```diff
@@ -696,6 +696,13 @@ server:
   ### When empty server.persistentVolume.mountPath is used instead
   storagePath: ""
 
+  ## If true, binds the Prometheus server to localhost and hence does not accept traffic from outside the
+  ## Kubernetes Pod. Useful in service mesh setups (e.g. Consul Connect, Istio, ...), where traffic flows through
+  ## sidecar proxies.
+  ## If using that feature, you most likely also need to configure "server.customReadinessProbe" and "server.customLivenessProbe"
+  ## and expose the respective HTTP paths on different ports outside of the service mesh (e.g. via Consul's exposed paths feature)
+  listenOnLocalhost: false
```
monotek, line 704: Imho if we make use of "--web.listen-address" it should be configurable and not bound to localhost only.

monotek (changes requested):

charts/prometheus/templates/server/deploy.yaml
harti2006: Thanks for the feedback. I got similar feedback from one of my colleagues as well. I did the change now to 
```yaml
listenOnHost: localhost
```
so users can freely configure the host name. The port number 9090 is not configurable.

monotek: Why not also make the port configurable like?

```
listen:
  host:
  port:
```

harti2006: I like the config style:
```yaml
listen:
  host: ...
```
because it is extendable. However I'd rather not implement switching the port in this PR, because the 9090 is hard-coded in so many places. Tomorrow I will adjust the values.yaml according to your suggestion.

charts/prometheus/templates/server/sts.yaml
```diff
@@ -110,21 +110,32 @@ spec:
           {{- if .Values.server.baseURL }}
             - --web.external-url={{ .Values.server.baseURL }}
           {{- end }}
+          {{- if .Values.server.listen.host }}
+            - --web.listen-address={{ .Values.server.listen.host }}:9090
```
monotek, line 114: i guess this will not match with "listen: {}" from the values.

monotek (changes requested):

charts/prometheus/templates/server/sts.yaml
harti2006: What do you mean? I initialized this "listen" block with an empty map "{}", because otherwise I get this error when checking `if .Values.server.listen.host`
>  <.Values.server.listen.host>: nil pointer evaluating interface {}.host

The alternative is to do `if .Values.server.listen and .Values.server.listen.host`, both in deploy.yaml and sts.yaml.

fktkrt:
We are using multiple Prometheus instances for more than a year with Istio meshes, and based on our experience, these changes are not really needed (at least for Istio, but I have the same feeling regarding Consul as well).

Usually, you don't want to inject Prometheus components w/ sidecars, but even if you do, service meshes should take care of the workarounds needed and not the Prometheus configurations. Istio takes care of this by having best practices around not injecting  Prometheus workloads, and providing workarounds by having custom annotations and suggesting custom scrape configs.

After a quick search, this official guide came up regarding Consul: https://learn.hashicorp.com/tutorials/consul/kubernetes-layer7-observability

It also mentions dropping sidecars for Prometheus.

Based on these, I'd suggest approaching this problem from another angle.

harti2006:
In our setup we are already running Grafana in the Consul Connect service mesh. Now I'd just like to make it able to use several instances of Prometheus as datasources. Each Prometheus runs in a separate Kubernetes cluster. The service mesh connects all these clusters and ensures traffic is encrypted, authenticated and authorized.
This already works perfectly fine with the current version of the Prometheus helm chart. My suggested changes are just "on-top" to make sure that no unauthorized traffic reaches Prometheus.
Yes there is the "Transparent Proxy" coming to Consul Connect which would also solve the problem (block unauthorized traffic + rewrite health probes), but this feature is very new and does not yet work in an environment with multiple Kubernetes clusters.

https://learn.hashicorp.com/tutorials/consul/kubernetes-layer7-observability is a nice tutorial that shows how Consul and Prometheus are integrated together, but it doesn't elaborate why Prometheus should not be protected by the service mesh.

PR #1248, opened 2021-08-12
[prometheus-operator] (Re)adding prometheus-operator chart

langecode:
Replacing PR #836 - extract Prometheus Operator deployment from `kube-prometheus-stack` into separate chart (see #760) in time the code in `kube-prometheus-stack` could then be replaced with reference to this chart - however this will make it possible to install just Prometheus Operator instead of the full monitoring stack.

Signed-off-by: Thor Anker Kvisgård Lange <tal@netic.dk>


monotek (approved):

charts/prometheus-operator/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+type: application
+maintainers:
+  - name: langecode
+    email: tal@netic.dk
+  - name: bismarck
+name: prometheus-operator
+sources:
+  - https://github.com/prometheus-community/helm-charts
+  - https://github.com/prometheus-operator/prometheus-operator
+version: 0.2.0
```
monotek, line 14: why are you starting with 0.2.0?

monotek (changes requested):

charts/prometheus-operator/Chart.yaml
langecode: Good catch - I think it was just a leftover - should probably just start out `1.0.0`

langecode:
@monotek Bump... The version was update a while ago 😀

monotek (approved):

monotek:
@prometheus-community/helm-charts-admins 
we need a review to get the codeowners change approved.

vsliouniaev:
It's currently already possible to install prometheus-operator without installing all the dependencies using the kube-prometheus-stack chart. The original approach to the chart was one composed of individual pieces, which was then combined to simplify maintenance. One of the reasons for it was the requirement for cross-bumping version numbers across interdependent charts. There's already quite a lot of time spent doing that by various contributors and I'd hesitate do go back to that model.

langecode:
Obviously I do not agree with this reasoning since I proposed this chart 😀
for multiple reasons.

First of all installing Prometheus Operator using the kube-stack chart - by
switching off all other components - would really imply a lot of version
bumping without any changes to the Prometheus Operator deployment -
wouldn’t it. The lifecycle of kube-stack is that of the combination of all
the components which the kube-stack chart aggregates. If I were to use that
chart, a lot of version bumps would really not concern Prometheus Operator
but instead the components switched off any way.

For the same reason I would also have chosen a design (we already maintain
some similar aggregating charts) were each individual software component
had its own chart following the lifecycle of that component. Making it
clear what a new release is concerned with.

I cannot see why it should be more work to bump the version of a dependency
than having to bump the image version and do the appropriate changes to an
“in-lined” chart. On the contrary having a chart for each component should
make each chart a lot more manageable and easier to test.

Naturally the combination of versions of the components have to be tested
in the aggregating kube-stack chart but this cannot be different if you
bump the version or change one of the “in-lined” components. And still Helm
chart dependencies is versioned so it is not like new versions need to come
in automatically.

In essence I see something like the kube-stack chart as some (a lot)
configuration combined with a collection of individual components having
their own lifecycle. However, currently the lifecycle of each component is
merged into one versioning of the whole chart.

Which is why we cannot use the kube-stack chart for installing Prometheus
Operator. But you are, of course, free to accept the contribution and the
offer to help maintain it or not. We can maintain it out of our own
repository instead.

monotek:
Imho some people would like to use this chart even if its not meant to be used as a dependency chart for kube-prometheus-stack later. 

So imho no reason to not have it as a standalone chart.

zanhsieh (dismissed):

charts/prometheus-operator/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+type: application
+maintainers:
+  - name: langecode
+    email: tal@netic.dk
+  - name: bismarck
+name: prometheus-operator
+sources:
+  - https://github.com/prometheus-community/helm-charts
+  - https://github.com/prometheus-operator/prometheus-operator
+version: 1.0.0
```
monotek, line 14: Imho we should start with 0.1.0

monotek (changes requested):

charts/prometheus-operator/Chart.yaml
langecode: Yeah, sure we can do `0.1.0` for a start

monotek (approved):

daenney:
I might be misunderstanding something, but something feels broken here:

```
$ helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
$ helm repo update
$ helm show chart prometheus-community/prometheus-operator
apiVersion: v1
appVersion: 0.38.1
dependencies:
- condition: kubeStateMetrics.enabled
  name: kube-state-metrics
  repository: https://kubernetes-charts.storage.googleapis.com/
  version: 2.8.*
- condition: nodeExporter.enabled
  repository: https://kubernetes-charts.storage.googleapis.com/
  version: 1.10.*
- condition: grafana.enabled
  name: grafana
  repository: https://kubernetes-charts.storage.googleapis.com/
  version: 5.3.*
deprecated: true
description: DEPRECATED - This chart will be renamed. See https://github.com/prometheus-community/community/issues/28#issuecomment-670406329
home: https://github.com/coreos/prometheus-operator
icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/assets/prometheus_logo-cb55bb5c346.png
keywords:
- operator
- prometheus
name: prometheus-operator
sources:
- https://github.com/coreos/kube-prometheus
- https://github.com/coreos/prometheus-operator
- https://coreos.com/operators/prometheus
version: 9.3.2
```

I seem to be picking up on an old chart with a very different version? Did something name clash?

monotek:
Damn.... I will delete the chart again....

daenney:
Since a chart by that name and with that version seems to already have existed in the past, shouldn't this get bumped to something like `0.40.0` to avoid any confusion?

monotek:
It's another chart now and it seems to be a bad idea to add it again. I'll remove it....

daenney:
That means the `prometheus-operator` chart will be removed entirely? I was kinda hoping to use it as the `kube-prometheus-stack` umbrella chart does too much for me right now.

monotek:
Yes. also see this comment: https://github.com/prometheus-community/helm-charts/pull/1248#issuecomment-931433976

siegenthalerroger:
Sorry to ask the question but why is your argumentation here that "you can disable everything you don't need", yet you maintain seperate charts for kube-state-metrics and node-exporter. What is the difference in these two cases?

monotek:
node-exporter and kube-state-metrics are completly different applications.

langecode:
Thanks for the effort @monotek - I am sorry that the chart is removed again.

As stated earlier on, I also find it rather counter intuitive to take an aggregation kind of chart and disable all but one component (which by itself is a separate application with its own lifecycle, https://github.com/prometheus-operator/prometheus-operator). I find it rather problematic to have to use the `kube-prometheus-stack` chart to only install "prometheus-operator" for a number of reasons, but I think these are already stated earlier on.

@daenney FWIW we have a chart solely intended for prometheus-operator deployment here: https://github.com/neticdk/k8s-oaas-observability/tree/main/charts/prometheus-operator

It is MIT licensed so it is there for everyone to use  but without any warranties :-)

As I was hoping to contribute it to the prometheus community collection of charts we have not kept it completely up-to-date so it may be a little bit out of date right now but we will make sure to update it soon as we are using this ourselves on a number of production clusters.

PR #1286, opened 2021-08-24
[prometheus-node-exporter] Fix port of pod in daemonset

### LINKED ISSUE #1279: [prometheus-node-exporter] Port of daemonset is incorrect

The values of the chart allow to configure the service and especially its ports:
```yaml
service:
  type: ClusterIP
  port: 9100
  targetPort: 9100
  nodePort:
  listenOnAllInterfaces: true
  annotations:
    prometheus.io/scrape: "true"
```

In the daemonset, its port is configured with the variable `.Values.service.port`
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/templates/daemonset.yaml#L66
```yaml
ports:
  - name: metrics
    containerPort: {{ .Values.service.port }}
    protocol: TCP
```
when it should be `.Values.service.targetPort`.

This inversion is present three more times in the daemonset file at the lines 47, 71 and 75.

It works if the values `service.port` and `service.targetPort` are identical, what is their defaults, but it fails if we override one of these values when installing the chart.

### What's your helm version?

3.5.2

1.20.4

### Which chart?

prometheus-node-exporter

### What's the chart version?

2.0.4

### What happened?

The pod of the daemonset does not listen to the expected port, and the service can not redirect to the pod.

The daemonset should listen to the correct port.

### How to reproduce it?

Install the chart by overriding either `service.port` or `service.targetPort`. 

```yaml
service:
  targetPort: 9101
```

helm install node-exporter prometheus-community/prometheus-node-exporter --values values.yaml

_No response_

fredgate:

Fix the port the pod of prometheus-node-exporter is listening for. It used the variable `.Values.service.port` which is the port of the service, not the pod, instead of the variable `.Values.service.port` which is the port to which the service redirects in the pod.

  - fixes #1279 

#### Checklist

- [x] Chart Version bumped


mrueg:
Thanks for the contribution, can you bump the chart version as well?

fredgate:
Yes I can bump the chart version. Is it okay to bump from `2.0.4` to `2.0.5` ? And `appVersion` remains the same.

monotek (changes requested):
I'm not sure if this is a simple bugfix version change as some user could have changed the service port but not the target port and the update would be a breaking change fo them.

fredgate:
This is not a problem because users could not change the port but not the target port, because it could not work.

For example :
| | Service port | Service targetPort  | Pod port |
|---|---|---|---|
| Variables | service.port  | service.targetPort | service.port |
| Default values | 9100  | 9100 | 9100 |
| Custom service port (ko) | 9105 | 9100 | 9105 |

So the service could not redirect to the pod on port 9100 because it was listening on port 9105.

vsliouniaev (approved):

vsliouniaev (changes requested):
I would still like to see the chart major version bumped for this. Even though there was no valid configuration for the chart itself, someone could have used a chart in part somewhere else and is relying on `.Values.service.port`

monotek:
As soon as we have the info in the readme....

monotek (changes requested):
A major change should be described in the updating section of the readme...

PR #1323, opened 2021-09-08
[kube-prometheus-stack] Flags to enable/disable burnrate/histogram rules

commixon:
In this PR we add flags to enable/disable burnrate/histogram rules for kube-apisever.

All rules follow the same convention, that is something like:
```
{{- if and (semverCompare ">=1.14.0-0" $kubeTargetVersion) (semverCompare "<9.9.9-9" $kubeTargetVersion) .Values.defaultRules.create .Values.kubeApiServer.enabled .Values.defaultRules.rules.kubeApiserverAvailability }}

```
Burnrate and histogram rules for kube-apiserver were lacking an explicit flag, like the rest of the rules. This resulted in them being deployed by default and not being able to exclude them.

Here we add `kubeApiserverBurnrate` and `kubeApiserverHistogram` to be used as flags (by default true to keep backwards compatibility)

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/alertmanager.rules.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/alertmanager.rules.yaml
@@ -29,7 +29,7 @@ spec:
     - alert: AlertmanagerFailedReload
       annotations:
         description: Configuration has failed to load for {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.pod{{`}}`}}.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}alert-name-alertmanagerfailedreload
+        runbook_url: https://runbooks.prometheus-operator.dev/runbooks/alertmanager/alertmanagerfailedreload
         summary: Reloading an Alertmanager configuration has failed.
       expr: |-
         # Without max_over_time, failed scrapes could create false negatives, see
@@ -44,7 +44,7 @@ spec:
     - alert: AlertmanagerMembersInconsistent
       annotations:
         description: Alertmanager {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.pod{{`}}`}} has only found {{`{{`}} $value {{`}}`}} members of the {{`{{`}}$labels.job{{`}}`}} cluster.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}alert-name-alertmanagermembersinconsistent
+        runbook_url: https://runbooks.prometheus-operator.dev/runbooks/alertmanager/alertmanagermembersinconsistent
         summary: A member of an Alertmanager cluster has not found all other cluster members.
       expr: |-
         # Without max_over_time, failed scrapes could create false negatives, see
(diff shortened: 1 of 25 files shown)


bismarck (changes requested):
Please see [note](https://github.com/prometheus-community/helm-charts/blob/2fae1d5ffb04849a149633693e0d21b68eec8454/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kube-apiserver-burnrate.rules.yaml#L1-L5) at the top of the rule files.

commixon:
> Please see [note](https://github.com/prometheus-community/helm-charts/blob/2fae1d5ffb04849a149633693e0d21b68eec8454/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kube-apiserver-burnrate.rules.yaml#L1-L5) at the top of the rule files.

:facepalm: I completely missed that.  Thank you very much @bismarck , will do that and update accordingly

monotek (approved):

commixon:
Hey @bismarck and @monotek !

I have run the prometheus rules sync script to properly generate the yaml files. It seems the sync function brings some other changes from the upstream as well. Is this ok?

charts/kube-prometheus-stack/templates/prometheus/rules-1.14/alertmanager.rules.yaml
```diff
@@ -29,7 +29,7 @@ spec:
     - alert: AlertmanagerFailedReload
       annotations:
         description: Configuration has failed to load for {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.pod{{`}}`}}.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}alert-name-alertmanagerfailedreload
+        runbook_url: https://runbooks.prometheus-operator.dev/runbooks/alertmanager/alertmanagerfailedreload
```
monotek, line 32: not sure if this should be hardcoded?

monotek (changes requested):

charts/kube-prometheus-stack/templates/prometheus/rules-1.14/alertmanager.rules.yaml
commixon: Me neither, but this seems to be the choice upstream if I understand correctly (have seen this in other places as well). I could skip those changes, but this will be something to be tackled in general when the script is being invoked

commixon: There is always the possibility of me messing something up :smile:, but when I run the script locally this is what it was produced. If we don't want to tackle this now I can just keep the changes in the script, that add the flags for burnrate and histogram rules and the respective changes in the templated files

commixon: Could add something like this:
```
    'https://runbooks.prometheus-operator.dev/runbooks': {
        'replacement': '{{ .Values.defaultRules.runbookUrl }}',
        'init': ''},
```
Which produces a diff like the following:
```
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}alert-name-alertmanagerconfiginconsistent
+        runbook_url: {{ .Values.defaultRules.runbookUrl }}/alertmanager/alertmanagerconfiginconsistent
```

HailD:
Could we get a review and merge on this one? It seems like a trivial thing, and definitely something that should be fixed.

monotek:
This is absolete in the meantime. you can disable every single rule now.

PR #780, opened 2021-03-19
[kube-prometheus-stack] unify timezone in grafana dashboards

steinbrueckri:
Signed-off-by: Richard Steinbrück <richard.steinbrueck@googlemail.com>

#### What this PR does / why we need it:
- unify timezone in grafana dashboards

#### Which issue this PR fixes
- fixes #779

#### Special notes for your reviewer:
I found the following commit and PR from the old helm stable repo where many resource changes inc. the timezone and in the PR nobody has mentioned this changed and in my eyes, that's pretty wired.

- c9542ad
- https://github.com/helm/charts/pull/22900/files#diff-2a7f38bd4b1ecfa7d2043fc1ffaa362a4d1b0226737fa08dd409c1e6081bd942

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #1326, opened 2021-09-08
[kube-prometheus-stack] Allow to set a timezone for the default grafana dashboards

krufab:
Allow to set a timezone for the default grafana dashboards.

The default dashboards have different timezones (utc, UTC, browser) and are loaded as read only in Grafana. Therefore, any change the user performs regarding the timezone, it cannot be saved.
This MR gives the user the possibility to set the some default timezone for all the dashboards (i.e. his / her own local timezone), without the need of modifying them.

Fixes #780 

#### Checklist
- [x] Chart Version bumped

Signed-off-by: Fabio Kruger <10956489+krufab@users.noreply.github.com>

### DIFF
--- a/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
+++ b/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
@@ -150,7 +150,7 @@ def patch_json_for_multicluster_configuration(content, multicluster_key):
 
 def patch_json_set_timezone_as_variable(content):
     # content is no more in json format, so we have to replace using regex
-    return re.sub(r'\"timezone\"\s*:\s*\"[^"]*\"\s*,', '"timezone": "\{\{ .Value.grafana.defaultDashboardsTimezone \}\}",', content, flags=re.IGNORECASE|re.MULTILINE)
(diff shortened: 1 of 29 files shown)


charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
```diff
@@ -147,6 +148,11 @@ def patch_json_for_multicluster_configuration(content, multicluster_key):
     return content
 
 
+def patch_json_set_timezone_as_variable(content):
+    # content is no more in json format, so we have to replace using regex
+    return re.sub(r'\"timezone\":.*$', '"timezone": "\{\{ .Value.grafana.defaultDashboardsTimezone \}\}",', content, flags=re.IGNORECASE|re.MULTILINE)
```
vsliouniaev, line 153: Seems like this could go wrong, can we be more specific with the regex or parse a sub-section of the input file as json?

krufab: Yes, the regex was a bit weak.
I'd have liked to manipulate directly the json object, but '\{\{' is not an allowed json sequence, as '\' must be escaped all the times => '\\{\\{'. However the '\\{\\{' is not converted correctly by yaml_str_repr on line 97, as it replaces exactly '\{\{'.
I didn't want to change too much yaml_str_repr, nor introduce a complex function as patch_json_for_multicluster_configuration.

vsliouniaev (dismissed):

vsliouniaev (approved):

PR #1329, opened 2021-09-09
[prometheus-redis-exporter] allow using securityContext in containers

chesnovsky:
Signed-off-by: Stanislav Chesnovskiy <s.chesnovskiy@semrush.com>

This PR allows using security context in containers

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-redis-exporter/templates/deployment.yaml
+++ b/charts/prometheus-redis-exporter/templates/deployment.yaml
@@ -36,6 +36,10 @@ spec:
           {{- range $key, $value := .Values.extraArgs }}
             - --{{ $key }}={{ $value }}
           {{- end }}
+          securityContext:
+          {{- range $key, $value := .Values.securityContext }}
+            {{ $key }}: {{ $value }}
+          {{- end }}
           ports:
             - name: exporter-port
               containerPort: 9121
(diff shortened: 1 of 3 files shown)


charts/prometheus-redis-exporter/templates/deployment.yaml
```diff
@@ -36,6 +36,10 @@ spec:
           {{- range $key, $value := .Values.extraArgs }}
             - --{{ $key }}={{ $value }}
           {{- end }}
+          securityContext:
+          {{- range $key, $value := .Values.securityContext }}
+            {{ $key }}: {{ $value }}
+          {{- end }}
```
zanhsieh, line 42: what if securityContext in values.yaml is empty?

zanhsieh (changes requested):

acondrat:
Do you want to provide the option to configure securityContext at both POD and Container levels?

charts/prometheus-redis-exporter/templates/deployment.yaml
chesnovsky: In that case, we will see empty securityContext in the manifest, and container will run under the root user by default.
I could add to the values.yaml something like this
`securityContext: {
    runAsUser: 0,
    runAsGroup: 0
}`
or this
`securityContext: {}`

chesnovsky:
> Do you want to provide the option to configure securityContext at both POD and Container levels?

As far as I see, the POD security policy described at podsecuritypolicy.yaml has default values that could not be changed with templates.
If securityContext is in values.yaml it will not be implemented to the POD policy only to the container securityContext.

zanhsieh (approved):

PR #1335, opened 2021-09-10
[prometheus-pingdom-exporter] Minor tweaks to README

vitaliyf:
@monotek

I don't think the version for this chart needs to be bumped for these changes.

### DIFF
--- a/charts/prometheus-pingdom-exporter/Chart.yaml
+++ b/charts/prometheus-pingdom-exporter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-pingdom-exporter
-version: 2.4.0
+version: 2.4.1
 appVersion: 20190610-1
 home: https://github.com/giantswarm/prometheus-pingdom-exporter
 description: A Helm chart for Prometheus Pingdom Exporter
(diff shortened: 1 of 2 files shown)


torstenwalter:
@vitaliyf change looks good. Please bump the chart version. It needs to be increased for all changes. Otherwise no new release can be created.

torstenwalter (approved):

vitaliyf:
Worth mentioning that Pingdom deprecated v2 APIs in December 2020, so the container that this chart uses from [camptocamp/prometheus-pingdom-exporter](https://github.com/camptocamp/prometheus-pingdom-exporter) (which itself is a fork from [giantswarm/prometheus-pingdom-exporter](https://github.com/giantswarm/prometheus-pingdom-exporter)) does not work anymore. There is a pull-request to https://github.com/giantswarm/prometheus-pingdom-exporter/pull/26 by @timkra to make it work with v3 API but it hasn't been merged.

I was able to build a container from that pull request and tweak this chart to work with and would be happy to submit those chart changes once that code is merged by someone and built/uploaded to a registry.

rpahli (approved):

PR #1346, opened 2021-09-15
[kube-prometheus-stack] Add scrapeTimeout in kubeStateMetrics

commixon:
This is a simple addition to the ServiceMonitor values of **kube-state-metrics**.

The `scrapeTimeout` is added as an option. The intention is for this value to be configured to something else than the default when needed (for example because of timeouts due to wildcards being used with `--metric-labels-allowlist` etc)

#### Checklist

- [x] Chart Version bumped


vsliouniaev (approved):

stevehipwell:
Out of interest, why isn't the service monitor in the _kube-state-metrics_ chart brought up to spec to be used from _kube-prometheus-stack_ and then this service monitor could be removed? The same could also be done for the _prometheus-node-exporter_.

PR #1351, opened 2021-09-17
kube-state-metrics servicemonitor relabeling

bojleros:
Hi,

This change makes kube-state-metrics relabeling/metricsRelabeling possible. Useful thing especially when you configure custom labels in kube-state-metrics, particularly when you want to remove label prefix.

Regards,
Bart


monotek (changes requested):
Please sign your commits and raise minor chart version.

PR #1371, opened 2021-09-24
[alertmanager] Add support for ingressClassName

NiklasWagner:
This PR adds support for ingressClassName since the old [ingress.class annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation) is now considered as deprecated.

I had to introduce a breaking change to support the new pathType field.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,8 +6,8 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.12.2
-appVersion: v0.22.1
+version: 0.13.0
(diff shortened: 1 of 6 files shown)


monotek (changes requested):
As we're already updating the ingress we could just use the default one the command `helm create alertmanager` creates, if helm 3.7.0 is used.

Thoughts?

NiklasWagner:
Yeah that is a good idea to avoid different implementation, I've updated the PR :)

monotek (changes requested):
Please fix ci.

And did you also copied the ingress config from the helm created values file?

NiklasWagner:
Hi, yeah I did copy the original helm ingress from the helm template. I've fixed the CI.

monotek (approved):

PR #1372, opened 2021-09-24
[kube-prometheus-stack] Configurable thresholds for KubeQuotaAlmostFull alert rule

amuraru:

Optionally, allow KubeQuotaAlmostFull threshold condition to be overwritten. Defaults to the current hard-coded value.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 20.0.1
+version: 20.0.2
 appVersion: 0.52.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -60,6 +60,10 @@ defaultRules:
 
   ## Runbook url prefix for default rules
   runbookUrl: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#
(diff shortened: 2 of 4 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -60,6 +60,8 @@ defaultRules:
 
   ## Runbook url prefix for default rules
   runbookUrl: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#
+  ## Thresholds for KubeQuotaAlmostFull alert
+  kubernetesResourceAlmostFullQuotaCondition: ' > 0.9 < 1'
```
vsliouniaev, line 64: Since we're starting a new paradigm here, do you think it would be a good idea to have the values in a separate block structure and also match them to the alert name itself?

amuraru: @vsliouniaev thanks for suggestion. I extended the rules sync script to support the notion of `alert_expressions_replacement_map ` that be used to override alert expressions only:
https://github.com/prometheus-community/helm-charts/pull/1372/files#diff-90d0cb7c99a0aca67494fd6a5957bac966b4a6a7a7e5fb59759251d7e9ca2fbbR151

LMK

shinebayar-g:
Cool, can you do the same for CPUThrottling one? 😁

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 19.2.2
+version: 19.2.3
```
monotek, line 21:
```suggestion
version: 19.2.4
```

monotek (changes requested):

amuraru:
rebased

PR #1375, opened 2021-09-27
[kube-prometheus-stack] Separate config reloader limit and request

ctrox:

Previously it was not possible to set different values for requests/limits of the config reloader. This simply separates `configReloaderCpu` and `configReloaderMemory` into two each for specifying both the limit and request.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 20.0.1
+version: 21.0.0
 appVersion: 0.52.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1645,20 +1645,23 @@ prometheusOperator:
   ##
   # alertmanagerDefaultBaseImage: quay.io/prometheus/alertmanager
 
-  ## Prometheus-config-reloader image to use for config and rule reloading
+  ## Prometheus-config-reloader
   ##
-  prometheusConfigReloaderImage:
-    repository: quay.io/prometheus-operator/prometheus-config-reloader
-    tag: v0.52.0
-    sha: ""
+  prometheusConfigReloader:
+    # image to use for config and rule reloading
+    image:
+      repository: quay.io/prometheus-operator/prometheus-config-reloader
+      tag: v0.52.0
+      sha: ""
 
-  ## Set the prometheus config reloader side-car CPU limit
-  ##
-  configReloaderCpu: 100m
-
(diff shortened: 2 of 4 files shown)


charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 18.1.0
+version: 18.2.0
```
monotek, line 21: As this changes the current value i would consider this a breaking change. Therefor major version change needed.

```suggestion
version: 19.0.0
```

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1608,13 +1608,15 @@ prometheusOperator:
     tag: v0.50.0
     sha: ""
 
-  ## Set the prometheus config reloader side-car CPU limit
+  ## Set the prometheus config reloader side-car CPU resources
```
monotek, line 1611: As we break this already i would use the chance to do it like this:

```
  ## Prometheus-config-reloader
  ##
  prometheusConfigReloader:
    # image to use for config and rule reloading
    image:
      repository: quay.io/prometheus-operator/prometheus-config-reloader
      tag: v0.50.0
      sha: ""

    # resource config for prometheusConfigReloader
    resources:
      requests:
        cpu: 100m
        memory: 50Mi
      limits:
        cpu: 100m
        memory: 50Mi
```

deployments.yaml needs to be adjusted acordingly.

monotek (changes requested):

charts/kube-prometheus-stack/Chart.yaml
ctrox: Makes sense, changed it.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 19.2.2
+version: 20.0.0
```
gkarthiks, line 21: apologies for the delayed response. can you please fix the conflicts, I will approve this PR

gkarthiks (changes requested):

charts/kube-prometheus-stack/Chart.yaml
ctrox: No worries, I also completely forgot about this PR. Just rebased it and bumped the version again.

monotek (approved):

JoKeRGreGre:
Is it possible to externalize the configuration of the config-loader for the Prometheus StatefulSet as well?

In case the initContainer doesn't have enough resources, Prometheus can't start, and it's not possible to change this configuration in the StatefulSet because it's overwritten by the Prometheus object.

For those who also encounter this issue, one solution would be to scale down the Prometheus to 0, clone the StatefulSet by removing the labels of the operator: operator.prometheus.io/*, and then modify the resources of the initContainers.

PR #1343, opened 2021-09-14
[kube-prometheus-stack] Use namespaceOverrides when set

sathieu:
#### What this PR does / why we need it:

Uses `grafana.namespaceOverride`, `kube-state-metrics.namespaceOverride` and `prometheus-node-exporter.namespaceOverride` when set, for created resource.

(includes #1342).

#### Checklist
- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #1383, opened 2021-09-29
[kube-prometheus-stack] Move ServiceMonitor back to main namespace, and use namespaceSelector only when needed

### LINKED ISSUE #1382: Don't specify namespaceSelector when running in same namespace
#1343 broke namespace independence of `ServiceMonitor` resources.
`namespaceSelector` shouldn't be set unless it's pointing at a different namespace, or explicitly overridden

sathieu:

Quoting @james-callahan:

> #1343 breaks namespace independence.
> namespaceSelector shouldn't be set unless it's pointing at a different namespace

  - fixes #1382

#### Checklist

- [x] Chart Version bumped


sathieu:
@vsliouniaev Please review. This follows [@james-callahan comment](https://github.com/prometheus-community/helm-charts/pull/1343#issuecomment-929851971).

james-callahan:
You don't need to necessarily move them back to their previous namespace; 
a better fix may be to make `namespaceSelector` conditional on differing values; essentially doing:
```
  {{ if ne (template "kube-prometheus-stack-kube-state-metrics.namespace" .) (printf "%s" (include "kube-prometheus-stack-kube-state-metrics.namespace" .) | quote ) }}
  namespaceSelector:
    matchNames:
      - {{ printf "%s" (include "kube-prometheus-stack-kube-state-metrics.namespace" .) | quote }}
  {{- end }}
```

This can probably be made nicer with a helper or the right logic elsewhere.

sathieu:
@james-callahan I went for a simpler fix: Remove all `namespaceSelector`s when they match the `ServiceMonitor`'s namespace.

vsliouniaev:
This is still preventing scraping these components deployed independently into a different namespace.

sathieu:
@vsliouniaev I couldn't find in the doc, but thought that the default was to use the `ServiceMonitor`'s namespace ...

vsliouniaev:
It used to be possible to deploy things like kube-state-metrics and node-exporter independently of this chart and have the servicemonitor point to a different namespace. 

With the changes introduced the servicemonitors must be always be deployed into the same namespace as the target, which may be problematic when the account doesn't have rights to other namespaces.

sathieu:
@vsliouniaev OK. So the first version of this PR (4fdc00d) is preferred. Is it?

(see also previous comment https://github.com/prometheus-community/helm-charts/pull/1383#issuecomment-929993738)

sathieu:
@vsliouniaev Please review this new version.

PR #1393, opened 2021-10-04
[kube-prometheus-stack] Support new default ports for controlplane

### LINKED ISSUE #1310: [kube-prometheus-stack] Make default kube-controller-manager and kube-scheduler ports depend on Kubernetes versions v1.22 and v1.23

Right now, all deployments to Kubernetes v1.22+ must change the kube-controller-manager port to 10257 explicitly to be able to scrape metrics from it, as default port has been changed in upstream Kubernetes:
> liveness/readiness probes to controller-manager MUST use HTTPS now, and the default port has been changed to 10257.

From https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#deprecation

Change default port to 10257 or make it dynamic based on detected Kubernetes version.

NONE

### Additional context.

Note that this might be a breaking change for clusters <v1.22.

As a workaround, when deploying on 1.22, one can add the following snippet to their values.yaml:
```yaml
kubeControllerManager:
  service:
    port: 10257
    targetPort: 10257
  serviceMonitor:
    https: true
    insecureSkipVerify: true
```

invidian:
    kube-controller-manager since Kubernetes 1.22 and kube-scheduler since
    1.23 by default only serve secure endpoints on new ports with HTTPS
    enabled, this commit makes default values for scraping dynamic, based on
    used Kubernetes version.

  - Closes #1310

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -975,8 +975,11 @@ kubeControllerManager:
   ##
   service:
     enabled: true
-    port: 10252
-    targetPort: 10252
+    ## If null or unset, the value is determined dynamically based on target Kubernetes version due to change
+    ## of default port in Kubernetes 1.22.
+    ##
+    port: null
+    targetPort: null
     # selector:
     #   component: kube-controller-manager
 
@@ -991,9 +994,10 @@ kubeControllerManager:
     proxyUrl: ""
 
     ## Enable scraping kube-controller-manager over https.
-    ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
+    ## Requires proper certs (not self-signed) and delegated authentication/authorization checks.
+    ## If null or unset, the value is determined dynamically based on target Kubernetes version.
     ##
-    https: false
+    https: null
 
     # Skip TLS certificate validation when scraping
     insecureSkipVerify: null
@@ -1200,8 +1204,11 @@ kubeScheduler:
   ##
   service:
     enabled: true
-    port: 10251
-    targetPort: 10251
+    ## If null or unset, the value is determined dynamically based on target Kubernetes version due to change
+    ## of default port in Kubernetes 1.23.
+    ##
+    port: null
+    targetPort: null
     # selector:
     #   component: kube-scheduler
 
@@ -1214,9 +1221,10 @@ kubeScheduler:
     ##
     proxyUrl: ""
     ## Enable scraping kube-scheduler over https.
-    ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
+    ## Requires proper certs (not self-signed) and delegated authentication/authorization checks.
+    ## If null or unset, the value is determined dynamically based on target Kubernetes version.
     ##
-    https: false
+    https: null
 
     ## Skip TLS certificate validation when scraping
     insecureSkipVerify: null
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 27.1.0
+version: 28.0.0
 appVersion: 0.53.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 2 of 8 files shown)


invidian:
Activity

charts/kube-prometheus-stack/values.yaml
```diff
@@ -965,8 +965,8 @@ kubeControllerManager:
   ##
   service:
     enabled: true
-    port: 10252
-    targetPort: 10252
+    port: 10257
```
monotek, line 968: I dont like this solution, as it will fail for people who want to update the chart but are still on k8s 1.19.
Imho there should be some logic, using the ports depending on the k8s version or custom ports.

Please also rebase.

monotek (changes requested):

charts/kube-prometheus-stack/values.yaml
invidian: Sure, adding logic sounds reasonable. So I'll change those defaults to `null` and they will be determined dynamically based on K8s version (helm provides that).

invidian: Done.

monotek: Id like to wait for one of the chart maintainers what the think about the solution.

mrueg (approved):
Can you rebase on latest changes, bump the chart version (it's probably worth to raise major as it changes default configuration) and add some documentation into the values yaml and readme? Folks might be confused why it's set to null.

invidian:
>Can you rebase on latest changes, bump the chart version (it's probably worth to raise major as it changes default configuration) and add some documentation into the values yaml and readme? Folks might be confused why it's set to null.

Thanks for reviewing, done :+1:

charts/kube-prometheus-stack/values.yaml
```diff
@@ -993,7 +995,7 @@ kubeControllerManager:
     ## Enable scraping kube-controller-manager over https.
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
     ##
-    https: false
+    https: null
```
mrueg, line 998: Can you add a line above that says something like:
"If null or unset, determined dynamically based on target Kubernetes version

charts/kube-prometheus-stack/values.yaml
```diff
@@ -975,8 +975,10 @@ kubeControllerManager:
   ##
   service:
     enabled: true
-    port: 10252
-    targetPort: 10252
+    # Default value is determined dynamically based on target Kubernetes version due to change
```
mrueg, line 978:
```suggestion
    # If null or unset, the value is determined dynamically based on target Kubernetes version due to change
```

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1216,7 +1220,7 @@ kubeScheduler:
     ## Enable scraping kube-scheduler over https.
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks
     ##
-    https: false
+    https: null
```
mrueg, line 1223: Same here

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1200,8 +1202,10 @@ kubeScheduler:
   ##
   service:
     enabled: true
-    port: 10251
-    targetPort: 10251
+    # Default value is determined dynamically based on target Kubernetes version due to change
```
mrueg, line 1205:
```suggestion
    # If null or unset, the value is determined dynamically based on target Kubernetes version due to change
```

charts/kube-prometheus-stack/values.yaml
invidian: Sure.

monotek (approved):

mrueg (approved):

desaintmartin:
Nice!
Note that even with kube < 1.22, some (most?) of us still would like to use https. In my setup, I basically did by hand everything that you automate in your PR. Should we add an optional flag (or two?) that triggers this even with kube < 1.22 / 1.23? Or is that better to keep it manual?

invidian:
>Note that even with kube < 1.22, some (most?) of us still would like to use https. In my setup, I basically did by hand everything that you automate in your PR. Should we add an optional flag (or two?) that triggers this even with kube < 1.22 / 1.23? Or is that better to keep it manual?

Not sure if having an extra flag make sense here. In non-default setups then, it's just a matter of setting 2 fields (targetPort and https), adding separate field to reduce setting 2 fields to one (with better name though), seems like more complexity than actual gain from it.

mrueg:
> Nice! Note that even with kube < 1.22, some (most?) of us still would like to use https. In my setup, I basically did by hand everything that you automate in your PR. Should we add an optional flag (or two?) that triggers this even with kube < 1.22 / 1.23? Or is that better to keep it manual?

I'd expect k8s<1.22 to phase out this year, so it might not be worth adding that extra logic, as new users hopefully start with a recently published version of kubernetes.

PR #1394, opened 2021-10-04
[prometheus] Use extraArgs for alertmanager configmap-reload container

ktaborski:
values.yaml contains extraArgs for alertmanager configMapReload, however this dictionary is not used in container. I need it to add extra argument `webhook-retries`.

#### Checklist

- [x] Chart Version bumped

@gianrubio @zanhsieh @Xtigyro @monotek @naseemkullah


charts/prometheus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: 2.26.0
-version: 14.8.1
+version: 14.8.2
```
monotek, line 4: New feature should raise minor version

monotek (changes requested):

charts/prometheus/Chart.yaml
ktaborski: That's partially true. You can treat it also as bug-fix, because dictionary was mentioned in values.yaml and not used, see:
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L412
Anyway I will not insist to not change it ;).

zanhsieh (approved):

charts/prometheus/Chart.yaml
ktaborski: @monotek can you take a look?

PR #1492, opened 2021-11-05
[prometheus-node-exporter] add ability to set scheme if TLS is configured

rcanderson23:
Adds the ability to configure the probe schemes in case node exporter is configured with TLS. This also defaults the ServiceMonitor to use `http` and will use `https` if the user configures `HTTPS` on `probeScheme`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -21,7 +21,6 @@ prometheus:
     enabled: false
     additionalLabels: {}
     namespace: ""
-    scheme: http
     bearerTokenFile:
     tlsConfig: {}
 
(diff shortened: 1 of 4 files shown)


charts/prometheus-node-exporter/values.yaml
```diff
@@ -21,7 +21,6 @@ prometheus:
     enabled: false
     additionalLabels: {}
     namespace: ""
-    scheme: http
```
monotek, line 24: why are we changing an existin value?
this would need to raise major version as this is a breaking change for updating users.

monotek (changes requested):

charts/prometheus-node-exporter/values.yaml
rcanderson23: Hmm, I was trying to avoid requiring specifying the same information in two places when configuring scheme. Perhaps it is best to leave monitor configuration with just the probe configuration added?

monotek: Can't we just use the old var?

PR #1508, opened 2021-11-15
[prometheus-statsd-exporter] Fixed a bug that does not add mapping config when configuring a config map

andrewgkew:
When configuring statsd exporter and configuring the config mappings using mappingConfigMapName field in values file, so creating a config map with the mappings, the config map gets mounted on the pod but the process is not started with the statsd.mapping-config= parameter because the conditional logic was only checking the mappingConfig field which is setting the config inline in the values file.

  - fixes #

#### Checklist

- [x] Chart Version bumped
- [x ] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

### DIFF
--- a/charts/prometheus-statsd-exporter/Chart.yaml
+++ b/charts/prometheus-statsd-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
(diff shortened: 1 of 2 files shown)


scDisorder:
LGTM, thanks!
Did not use that for a while...

scDisorder (approved):

monotek (approved):

PR #1485, opened 2021-11-04
[kube-prometheus-stack] Prometheus operator version upgrade v0.52.0

andrewgkew:
#### What this PR does / why we need it:
Upgrading Prometheus Operator version (and CRD changes) to version v0.52.0. This adds additional features one in question the sigv4 setting to allow remote writing to Amazon Managed Prometheus

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Special notes for your reviewer:
The new version v0.52.0 was only released by the operator team today, 4th November 2021
I have a project that is dependent on this change, though I have tested it as a local chart and it has successfully worked, but having it pushed to the main chart is important.

Thank you

@torstenwalter - added you here but cant add you as a reviewer. Noticed you have approved a PR recently. I am looking for this change to get pushed through as quickly as possible :) thanks

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #1510, opened 2021-11-17
[kube-prometheus-stack] Use kubectl replace to upgrade prometheuses CRD

### LINKED ISSUE #1500: [kube-prometheus-stack] Can't upgrade CRD from 0.50.0 to 0.52.0: metadata.annotations: Too long: must have at most 262144 bytes

When applying the new 0.52.0 CRD, I get this result:

```console
customresourcedefinition.apiextensions.k8s.io/alertmanagerconfigs.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/alertmanagers.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/podmonitors.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/probes.monitoring.coreos.com configured
The CustomResourceDefinition "prometheuses.monitoring.coreos.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
customresourcedefinition.apiextensions.k8s.io/prometheusrules.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/servicemonitors.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/thanosrulers.monitoring.coreos.com configured
```

### What's your helm version?

version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}

Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.11", GitCommit:"27522a29febbcc4badac257763044d0d90c11abd", GitTreeState:"clean", BuildDate:"2021-09-15T19:21:44Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"} Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.11", GitCommit:"27522a29febbcc4badac257763044d0d90c11abd", GitTreeState:"clean", BuildDate:"2021-09-15T19:16:25Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}

### Which chart?

kube-prometheus-stack

### What's the chart version?

20.0.1

### What happened?

When applying the new 0.52.0 CRD, I get this result:

```console
customresourcedefinition.apiextensions.k8s.io/alertmanagerconfigs.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/alertmanagers.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/podmonitors.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/probes.monitoring.coreos.com configured
The CustomResourceDefinition "prometheuses.monitoring.coreos.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
customresourcedefinition.apiextensions.k8s.io/prometheusrules.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/servicemonitors.monitoring.coreos.com configured
customresourcedefinition.apiextensions.k8s.io/thanosrulers.monitoring.coreos.com configured
```

So I can't upgrade the CRD `prometheuses.monitoring.coreos.com`.

I expected to upgrade CRD without errors.

### How to reproduce it?

Starting with a 19.2.3 kube-prometheus-stack instance and 0.50.0 installed CRD, try to upgrade the chart to 20.0.1 then applying the new 0.52.0 CRD (yes, it was a mistake from myself to upgrade the chart before upgrading CRD, but I don't think this is the problem here because I read the commits and the chart seems to be backward-compatible).

NONE

```sh
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_probes.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
```

_No response_

sathieu:

The `prometheuses` CRD is too long for ` kubectl apply` since prometheus-operator v0.52.0 (#1485).

This PR fixes this in the upgrade notes and when using ArgoCD.

  - fixes #1500

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
+++ b/charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
@@ -5,6 +5,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    argocd.argoproj.io/sync-options: Replace=true
     controller-gen.kubebuilder.io/version: v0.6.2
   creationTimestamp: null
   name: prometheuses.monitoring.coreos.com
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 28.0.1
+version: 28.1.0
 appVersion: 0.53.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -164,7 +164,7 @@ kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheu
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_probes.yaml
-kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
 kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
--- a/charts/kube-prometheus-stack/hack/update_crds.sh
+++ b/charts/kube-prometheus-stack/hack/update_crds.sh
@@ -22,5 +22,9 @@ for line in "${FILES[@]}" ; do
     URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/$VERSION/example/prometheus-operator-crd/$SOURCE"
     echo "# ${URL}" > ../crds/"${DESTINATION}"
     curl -L "${URL}" >> ../crds/"${DESTINATION}"
-
+    # CRD is too long
+    # https://github.com/prometheus-community/helm-charts/issues/1500
+    if [ "$SOURCE" = monitoring.coreos.com_prometheuses.yaml ]; then
+      sed -i 's@^  annotations:@  annotations:\n    argocd.argoproj.io/sync-options: Replace=true@' ../crds/"${DESTINATION}"
+    fi
 done


mosheavni (dismissed):

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
```diff
@@ -5,6 +5,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    argocd.argoproj.io/sync-options: Replace=true
```
monotek, line 8: i dont think we should add deployment specific annotations to the crd.
imho the change in the docs for manual upgrade should be enough.

monotek (changes requested):

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
sathieu: @monotek The problem here, is that CRDs are not templated. There is no way to add annotations except hacky workarounds.

jplanza-gr: @sathieu Great idea to add this directly to the CRD; it solves the problem for ArgoCD users and the annotation will simply be ignored for everyone else.  The "manual upgrade" isn't really a viable option when using ArgoCD; every sync attempt will fail because the CRD is too large.  Without this change, the only way for an ArgoCD user to fix this is to replace *everything* on each sync operation (risky) or to extract the CRD from the chart so it can be updated locally.

monotek: As helm is not supposed to update crds anyway, most people add and manage these crds "by hand" or via some other automated step, outside of helm..
Therefore the crd does not need such annotations. 
In the long term this would make the CRD also kind of messy, if everybody starts to add things for his special deployment method, which in the end are not realy needed.

monotek (changes requested):

sathieu:
@monotek While I understand the general rule to avoid using tool-specific annotations, the current situation is broken. Expanding the topic, the crds upgrade handling is currently a manual step, this could be improved.

A few charts handling this correctly:

- gatekeeper
  - [crds directory](https://github.com/open-policy-agent/gatekeeper/tree/3ba8e93049b542a0c273e60a57ca03bd24d95dd7/charts/gatekeeper/crds)
  - upgrade with a ["hook"](https://github.com/open-policy-agent/gatekeeper/blob/master/charts/gatekeeper/templates/upgrade-crds-hook.yaml) (ClusterRole+ClusterRoleBinding+ServiceAccount+Job)
 - velero:
   - [crds directory](https://github.com/vmware-tanzu/helm-charts/tree/58e695b586b80a7cec593955ba49f503bba77406/charts/velero/crds)
   - [upgrade-crds Job](https://github.com/vmware-tanzu/helm-charts/blob/58e695b586b80a7cec593955ba49f503bba77406/charts/velero/templates/upgrade-crds.yaml)
   - [cleanup-crds Job](https://github.com/vmware-tanzu/helm-charts/blob/58e695b586b80a7cec593955ba49f503bba77406/charts/velero/templates/cleanup-crds.yaml)

Some other charts use a subchart (like [here](https://github.com/elastic/cloud-on-k8s/tree/master/deploy/eck-operator/charts/eck-operator-crds)) but requires [extra manual steps](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-upgrading-eck.html#k8s-beta-to-ga-upgrade).

If we want to implement a solution like gatekeeper and velero, [this PR](https://github.com/prometheus-community/helm-charts/pull/1510) is still needed as crds don't allow templating.

If we want to implement the subchart solution, we'll have problem as kube-prometheus-stack uses the CRDs it creates (and we'll also have the upgrade problems).

J0n4t4n:
I also just hit this issue and support this change.
I think it is important to have a solution for the problem now even if it is not perfect, as this gives more time to think about and implement a better solution.

andrewgkew:
Would be good to separate the changes between the README change and then new argoCD addition so that the README can go up and users be aware of the new larger CRD from prometheus operator. 

Is that team aware of the larger CRD and have plans to shrink it?

monotek:
I guess one of the chart maintainers should decide how to proceed.

sathieu:
@monotek What can we do to move this PR forward?

It has fixes for upgrade with kubectl, and install/upgrade with ArgoCD.

Other methods will need other fixes (using kustomize, like https://github.com/cloudnativedaysjp/dreamkast-infra/pull/1262, or anything else), but at least this PR improves things.

While I understand your point about limiting tool-specific annotations (and this should be done when possible),  I think that perfect is the enemy of good.

This PR is currently a blocker for us (see [here](https://gitlab.com/kubitus-project/kubitus-installer/-/merge_requests/392) and [here](https://gitlab.com/kubitus-project/kubitus-installer/-/merge_requests/393))

monotek:
> @monotek What can we do to move this PR forward?

As i will not approve it you have to wait for one of the chart maintainers to decide.

sathieu:
@bismarck, @gianrubio, @gkarthiks, @scottrigby, @vsliouniaev, @Xtigyro Please review :pray:. This is a blocker for us currently...

mrueg:
Probably this change should be requested upstream in prometheus-operator. In my opinion this chart should not alter CRDs when importing them, as it is potentially error prone and the chance that someone checks imported CRDs for changes is almost zero.

sathieu:
@mrueg Upstream use kubebuilder, and I don't see any way to add annotations with it.

sathieu:
I've reported this upstream https://github.com/prometheus-operator/prometheus-operator/issues/4439.

@mrueg What about using a kustomize patch instead of sed? This is the method used in crossplane (https://github.com/crossplane/crossplane/pull/1020).

monotek:
by the way... server side apply should work too without error:

```
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_probes.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.52.0/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
```

sathieu:
@monotek My usecase is within ArgoCD (where server-side apply is not implemented https://github.com/argoproj/argo-cd/issues/2267). The only solutions I see is shrinking the CRD, or adding the annotation `argocd.argoproj.io/sync-options: Replace=true`.

alex-mozejko:
hello,

just a thought, would the prometheus people be open to a section within the values somewhere like `global.crd.annotations` where people can specify any additional annotations which should be added to crds?

then each crd template can be updated to something like: 

```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.6.2
    {{- if .Values.global.crd.annotations }}
    {{ toYaml .Values.global.crd.annotations | indent 4 }}
    {{- end }}
  creationTimestamp: null
  name: prometheuses.monitoring.coreos.com
```

this should avoid any implementation specific stuff getting into the vanilla helm chart, but will allow for defining annots without mucking around with kustomize / etc

stephaneetje:
This is very blocking for deploying with argocd (which is mainly used in lot of places to push helm charts). What are the workarounds if you are not going to approve this @monotek  ? We actually just can't deploy anymore without deleting the crd... which is a pretty bad solution

sathieu:
As this tiny patch is not accepted here (nor the upstream one https://github.com/prometheus-operator/prometheus-operator/pull/4443), I ended up doing a complex and hackish fix 😞 .

See https://github.com/argoproj/argocd-example-apps/issues/103#issuecomment-990793051 for detailed instructions.

Still, perfect is enemy of good, and refusing this PR makes things hard for everybody using ArgoCD.

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
Xtigyro: Guys, we have to pay attention to the way the CRDs are updated in this chart. It's not thru manual changes like the above but thru an automated process that pulls them from the Prometheus Operator repo. So this change here is even pointless because it will get overwritten upon the next major update of the CRDs.

Xtigyro (changes requested):

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
steinarox: @Xtigyro Wont it be added because of the changed update_crd script?

Xtigyro: @steinarox Ah, apologies - didn't realize we were updating that file after every pull from the Prometheus Operator repo. You are right.

Xtigyro: And one of the reasons I didn't realize that was because that script was not meant to modify the CRDs.

Xtigyro: Maybe the other maintainers might have another take on the matter but I'm with @monotek on this one. The chart is not meant to create new opinionated versions of the used external code. 

Having said that - @sathieu how about adding this functionality in a completely optional way. That is to say - adding another section in the `README.md` about tools like ArgoCD, Flux, etc. - and state what particular changes are needed to make the CRDs fully compatible with third-party tools that operate in a non-standard way. Maybe even putting another small script that does the changes upon manual execution?

Xtigyro:
@sathieu On top of the other comments I made - I would encourage fling an issue with the ArgoCD team and maybe solve the issue in a similar way how Flux solved it: https://github.com/fluxcd/flux2/issues/1889.

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
sathieu: @Xtigyro There is no easy way to workaround this on the ArgoCD side.

Possible correct solutions:

- Implement server-side apply in ArgoCD https://github.com/argoproj/argo-cd/issues/2267
- Support helm post renderers in ArgoCD https://github.com/argoproj/argo-cd/issues/3698

Heavy workaround:

- https://github.com/argoproj/argocd-example-apps/issues/103#issuecomment-990793051

I won't work more on this as I have a workaround (thanks to Ansible, see https://gitlab.com/kubitus-project/kubitus-installer/-/issues/70). But this workaround is clearly not easily actionable to everyone.

steinarox:
@Xtigyro link to issue has already been posted https://github.com/prometheus-community/helm-charts/pull/1510#issuecomment-987879810 And it seems like a solution is not going to happen in the short term. 

I understand that the proposed solution is suboptimal and "should" be solved upstream. But it seems like this is the fastest way to solve a big headache for now. 

Someone suggested to add a 

````
{{ with .Valueas.crd.annotations}}
annotations: 
```

irizzant:
Sorry guys but ArgoCD does work if you use a kustomized Helm approach: https://github.com/argoproj/argocd-example-apps/tree/master/plugins/kustomized-helm

Just add the patched CRD:
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    argocd.argoproj.io/sync-options: Replace=true
  name: prometheuses.monitoring.coreos.com
```

to the kustomization file and you're good to go

pblgomez:
> Sorry guys but ArgoCD does work if you use a kustomized Helm approach: https://github.com/argoproj/argocd-example-apps/tree/master/plugins/kustomized-helm
> 
> Just add the patched CRD:
> 
> ```yaml
> apiVersion: apiextensions.k8s.io/v1
> kind: CustomResourceDefinition
> metadata:
>   annotations:
>     argocd.argoproj.io/sync-options: Replace=true
>   name: prometheuses.monitoring.coreos.com
> ```
> 
> to the kustomization file as follows:
> 
> ```yaml
> patchesStrategicMerge:
> - crd-patch.yaml
> ```
> 
> and you're good to go

Could you please write step by step this process?

sathieu:
> Could you please write step by step this process?

@pblgomez , this is not as simple as @irizzant said. I've written a more complete step-by-step doc procedure at https://github.com/argoproj/argocd-example-apps/issues/103#issuecomment-990793051.

irizzant:
> @pblgomez , this is not as simple as @irizzant said.

Maybe, I don't think it's difficult either.
In my company we use the following ArgoCD ConfigMap

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
data:
  configManagementPlugins: |
    - name: kustomized-remote-helm
      init:
        command: ["/bin/sh", "-c"]
        args: ["helm repo add $HELM_REPO_URL && helm repo update"]
      generate:
        command: [sh, -c]
        args: ["helm template $OPTIONS --include-crds > all.yaml && kustomize build"]
```
Then you create the ArgoCD App as follows:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: prometheus
  finalizers:
    - resources-finalizer.argocd.argoproj.io
spec:
  project: default
  source:
    repoURL: '<<< my repo url >>>'
    path: <<< path >>>
    targetRevision: << branch >>
    plugin:
      name: kustomized-remote-helm
      env:
        - name: HELM_REPO_URL
          value: "prometheus-community https://prometheus-community.github.io/helm-charts"
        - name: OPTIONS
          value: "--release-name kube-prometheus-stack --namespace monitoring --version 23.1.5 -f values-sdbit.yaml prometheus-community/kube-prometheus-stack"
```

The "<<< path >>>" above is a directory that contains the following files:
* kustomization.yaml
```yaml
resources:
- all.yaml

patchesStrategicMerge:
- prometheus-crd-patch.yaml
```
* prometheus-crd-patch.yaml
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    argocd.argoproj.io/sync-options: Replace=true
  name: prometheuses.monitoring.coreos.com
```

charts/kube-prometheus-stack/crds/crd-prometheuses.yaml
brsolomon-deloitte: > As helm is not supposed to update crds anyway, most people add and manage these crds "by hand" or via some other automated step, outside of helm..

What examples exist that this is how "most people" managed CRDs? The kube-prometheus-stack Helm chart includes the CRDs and applying/creating CRDs through Helm/ArgoCD is common practice and works.

> Great idea to add this directly to the CRD; it solves the problem for ArgoCD users and the annotation will simply be ignored for everyone else.

Agreed; this is a simple and harmless 1-line annotation that solves what is otherwise a huge pain for administrators, and does not affect those not using ArgoCD. Is it really worth it to take a philosophical stance against "add[ing] deployment specific annotations to the crd"?

pblgomez:
too messy, I even have argocd managed by argocd so modifying the cm without values in helm are not really an option.

irizzant:
It's not messy at all, and the additional plugin can be used in a number of different scenarios.
We have ArgoCD managed by itself and if you installed it with Helm it's just a metter of adding the `configManagementPlugins` [here](https://github.com/argoproj/argo-helm/blob/0ee1099de92a77365823de079f2b2a463718c54f/charts/argo-cd/values.yaml#L1078) and create 2 files (kustomization and Prometheus CRD patch).

altitudems:
I think @irizzant 's solution is actually great. Appreciate you taking the time to share. Working great here.

patrickjahns:
One thing in regards to @irizzant solution - be careful with helm-charts that test for k8s versions.

See the remarks in the [ArgoCD Discussion on using customize with helm together](https://github.com/argoproj/argo-cd/issues/2789#issuecomment-960271294)

A suggestion from [this comment](https://github.com/argoproj/argo-cd/issues/2789#issuecomment-967836081) would be, to improve it as following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
data:
  configManagementPlugins: |
    - name: kustomized-remote-helm
      init:
        command: ["/bin/sh", "-c"]
        args: ["helm repo add $HELM_REPO_URL && helm repo update"]
      generate:
        command: [sh, -c]
        args: ["helm template $OPTIONS --kube-version $KUBE_VERSION --api-versions $KUBE_API_VERSIONS --include-crds > all.yaml && kustomize build"]
```

irizzant:
The [above comment](https://github.com/prometheus-community/helm-charts/pull/1510#issuecomment-1004307590) is a good improvement, though ArgoCD does not signal any out-of-sync resource in kube-prometheus-stack chart after applying it meaning that this chart does not use customized output based on k8s version.

patrickjahns:
> The above comment is a good improvement, though ArgoCD does not signal any out-of-sync resource in kube-prometheus-stack chart after applying it meaning that this chart does not use customized output based on k8s version.

Please be aware, that especially for ingress and pdb this is the case for this chart.  See:

Helpers:
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl#L132-L166

Ingress:
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/ingress.yaml#L7-L8

PDB: 
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/podDisruptionBudget.yaml#L2


So in your configuration and your use case the k8s version does not make a difference for the final rendered chart, for other users it might very well

irizzant:
Ah I see, I indeed don't use Ingress and PodDisruptionBudget in my case so that makes sense

FrediWeber:
@irizzant Please note, that even if this solution works for you, Helm discourages the use of `helm template` for generating resources that are to be deployed directly.
https://github.com/helm/helm/issues/3553#issuecomment-887770050
https://github.com/helm/helm/issues/3553#issuecomment-888681145

This issue is a huge blocker for us too and we would be glad, if we could add this annotation as a workaround until a more permanent solution is found upstream (e.g. shrinking the CRD).

sathieu:
@FrediWeber ArgoCD uses `helm template` internaly, so the same limitations apply.

irizzant:
@FrediWeber if you refer to the [linked discussion](https://github.com/argoproj/argo-cd/issues/2789#issuecomment-960271294) you'll see that using `helm template` to render the manifests is a solution that works not only for me.
Kustomize for example does exactly that when you enable Helm support to it using `kustomize.buildOptions: --enable-helm`.

Also I don't see where they're discouraging this, since the comment you linked ends up with:
> But we recognize that many users DO take the output of helm template to integrate parts of Helm into their projects in new and exciting ways. And we're happy to encourage that to improve upon what's been built by others in the community.

`helm template` is a suppported and perfectly legal Helm command to render the manifests which of course doesn't expect a cluster running so it doesn't have the possibility to check what Kubernetes API are supported, and that's why `--kube-version` and `--api-versions` can be added as parameters.

FrediWeber:
Yes, you sure can use it but it is not expected to work exactly as expected by chart developers. Not many projects test their charts against `helm template` (e.g. add all namespace fields in all templates).
But if ArgoCD uses `helm template` anyway, why not use the - in that case perfectly valid solution - from @irizzant and close this Pull Request? Maybe we could add something about it in some docs?

monotek:
There can't be a shrink of the CRD.
If you shrink it, you'll likely will remove stuff which is needed.
Even if you only remove some comments now, at some point, with some new version implementing new features, it will likely be to big again.

Please talk to ArgoCD developers to implement server side apply.
Thats the way to go. 
Not to adjust the chart for some particular deployment tool.

I still vote against changing the CRD in the chart.
No matter if permanently or temporarly.

irizzant:
@FrediWeber 
Helm 2 use to have an in-cluster agent to work with charts, but Helm 3 competely removed this component.
Now Helm first use the templates and values files and then translate them into manifests (and this step is the same as `helm template` except for the already mentioned in-cluster checks). 
Finally it deploys the rendered manifests using Kubernetes API.

Consequently yes, there are differences between manifests rendered with `helm template` but if you add the already mentioned parameters to `helm template` they end up to be the same. There is no need to test specifically for this.

Adding ArgoCD annotations to the upstream CRD is innocuous as a change and I agree on this, but my suggested approach (with [this update](https://github.com/prometheus-community/helm-charts/pull/1510#issuecomment-1004307590) ) is still a viable way to work with kube-prometheus-stack chart whether they change the upstream CRD or not.

monotek (changes requested):
Please remove update of python code and use `kubectl apply --server-side -f ...` instead of `kubectl replace` as we're already using it this way here now: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-24x-to-25x

FrediWeber:
@monotek I understand that your hesitant to implement this workaround because it is "just one solution" amongst many and there is the chance, that it will not be removed again. On the other side, there is an imminent problem of many ArgoCD users with this chart. Of course, this needs to be addressed in ArgoCD (e.g. by adding support for server side apply) but what is the downside of temporarily adding the annotation? Even if the CRDs doesn't get shrunk in the upstream chart it will take some time to implement server side apply in ArgoCD. This would be a viable workaround - and I say it again - that would spare many users some time without any downsides.

@irizzant Thank you very much for pointing this out. Actually I didn't know that and your solutions seems absolutely fine and we'll also using it if there is no other workaround. I just think that we could implement this workaround and thereby remove the necessity to modify every ArgoCD instance managing this chart.

FrediWeber:
@monotek Would it be okay for you, if I write something about this in the readme, in the "breaking changes" section you linked? I would include the solution of @irizzant. Strictly speaking it is not a problem of this project but it could help some people.

jplanza-gr:
I've been following along with the discussion here; I was hoping we'd get the ArgoCD annotation added but I can understand why the maintainers don't want to include something specific to a particular deployment tool.  I'd like to offer a compromise; would you folks consider providing a flag in the chart that makes CRD installation optional?

The `cert-manager` chart uses this to great effect: https://cert-manager.io/docs/installation/helm/#3-install-customresourcedefinitions

The problem they're working against is that if you remove and replace a chart with CRDs, all objects using the new resources are deleted... this is mighty inconvenient if those resources are certificates that you have to renew.

If ArgoCD users could suppress the installation of the CRDs, it's pretty easy to install them independently and make any necessary annotations or changes.  This allows `kube-prometheus-stack` to stay "vendor agnostic" and still provides a straightforward workaround for those of us who need it.

monotek:
Helm install already supports "--skip-crds" as argument out if the box. So no need to add it to the chart.

Cheshirez:
Workaround with `--skip-crds` would work, as it's supporting by `helm template` as well.
Reg. ArgoCD - there's PR to add this ability there https://github.com/argoproj/argo-cd/pull/8012/files

irizzant:
As already reported, [here](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-24x-to-25x) the suggested way is using `--server-side` kubectl parameter.

The way I see it another possibility could be adding a kubernetes Job to the chart to upgrade CRD which would execute the `kubectl apply --server-side` commands.

jplanza-gr:
@monotek @irizzant Have either of you used ArgoCD before?  When using a declarative system, we don't have direct access to "helm install" or "kubectl"... the entire point is to manage your helm charts with version control and to not run CLI commands directly.  The other suggestions in the thread (kustomize, etc) require a bunch of additional complexity that will make this installation harder to maintain.

Again, I respect that you don't want to make a vendor-specific change; that completely makes sense.  But providing the ability to suppress CRD installation seems like it would be trivial, vendor neutral, and help out many of the folks here who want to use Prometheus.  Instead, it seems like the approach here is to hold fast and wait for ArgoCD to innovate around you; is that really in the best interests of the community?

----
> Helm install already supports "--skip-crds" as argument out if the box. So no need to add it to the chart.
----
> As already reported, [here](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-24x-to-25x) the suggested way is using `--server-side` kubectl parameter.
> 
> The way I see it another possibility could be adding a kubernetes Job to the chart to upgrade CRDs which would execute the `kubectl apply --server-side` commands, similar to what is done [here](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/templates/upgrade-crds.yaml)

irizzant:
> @monotek @irizzant Have either of you used ArgoCD before?
No, never! Indeed [this comment](https://github.com/prometheus-community/helm-charts/pull/1510#issuecomment-998938176) was written by having absolutely no knowledge of ArgoCD.

>  When using a declarative system, we don't have direct access to "helm install" or "kubectl"... the entire point is to manage your helm charts with version control and to not run CLI commands directly
If a Job is added to this chart as I suggested [here](https://github.com/prometheus-community/helm-charts/pull/1510#issuecomment-1007241923), would you mind explaining what the required changes would be for your Application definition in ArgoCD? Or for your ArgoCD setup more in general?

jplanza-gr:
Sure, no problem.  Unless there are future changes to ArgoCD, here's what I'd do today.

1. Disable deployment of the CRDs by the main chart.
2. Create a new ArgoCD "Application" to manage just the CRDS; with the following [sync options](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/):
  -- ApplyOutOfSyncOnly=true
  -- Replace=true

When the chart deploys the CRDs, the main problem an ArgoCD user faces is that the scope of these sync options is applied to *everything* in the chart.  This means that we run the risk of having resources completely replaced rather than updated.  Off the top of my head, the most important thing we could lose is any persistent caching of metrics; a sync would run and those resources (including PVs) could be replaced rather than updated in-place.

If we can disable CRD installation from the main chart, it's easy to download the CRDs, store them elsewhere, and manage them with the right options.  CRDs don't change very often, so we can get the desired behavior for both the CRD and chart resources.   As I mentioned above, the "cert-manager" chart uses this to explicitly separate the CRD from the chart; it stores CA-validated certificates in custom resources, so you'd run the risk of destroying all your certificates during any operation that removed the CRDs.

I think the "kustomize" workarounds are viable if you're already using it, but anyone on vanilla Helm 3 charts would need to introduce that just for this one application.  Quite frankly, every org has practitioners with varying levels of experience and we'd rather not introduce complexity that will trip up newer folks.

Back to my earlier suggestion, a toggle to deploy the CRDs with a default of true would be perfect here; your typical users won't notice any change, but those of us who have to workaround the issue can do so.  Thanks for hearing me out on this.

monotek:
> @monotek @irizzant Have either of you used ArgoCD before? When using a declarative system, we don't have direct access to "helm install" or "kubectl"..

No. And reading this issue there is no reason i would.

The point here is, that you use tooling, which does not implement all options of Helm and Kubernetes completly, but still trying to use Helm charts.

Maybe you want to try out FluxCD, which supports --skip-crds and also server-side apply.

As there a workarounds desribed and to reduce noise from now on i will close this issue and lock the conversation. If you want to discuss the issue further, i suggest using the ArgoCD issue tracker.

PR #1545, opened 2021-11-29
[prometheus-postgres-exporter] Add support for queries in external configmap

otassetti-talend:

Currently queries are defined directly in the values, we want to be able to use a config-map outside of the helm-chart. (Our use case is to have several exporters and we want to use a common configmap as source of truth).

None

From design choice currently externalQueries will disable inline values queries.
Thanks

#### Checklist

- [x] Chart Version bumped


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.10.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 2.3.7
+version: 2.3.8
```

monotek (changes requested):

charts/prometheus-postgres-exporter/Chart.yaml
otassetti-talend: Hello,
Sure i will set it to 2.4.0.
Thanks

zanhsieh (approved):

zanhsieh (approved):

monotek (approved):

Issue #1425, opened 2021-10-14
[kube-prometheus-stack] Use sub-chart service monitors


stevehipwell:
### Is your feature request related to a problem ?

The current _kube-prometheus-stack_ chart creates custom service monitors for _prometheus-node-exporter_ and _kube-state-metrics_ despite both charts having built in service monitor support and being in this repo (I know this wasn't always the case). This adds additional complexity and the potential for inconsistent behaviour depending on how these components have been installed.

### Describe the solution you'd like.

I'd think that if the _prometheus-node-exporter_ and _kube-state-metrics_ chart service monitors could be brought up to spec with the bundled ones then they could be removed leaving only one service monitor per service to be maintained.

### Describe alternatives you've considered.

NONE

monotek:
Would you mind to create an pr for that?

stevehipwell:
@monotek sure I can take a look at this. It'd need to be 3 PRs, one each for the sub charts and one for the stack chart.

PR #1556, opened 2021-12-03
[kube-state-metrics] Improve ServiceMonitor

stevehipwell:
This PR add capabilities to the _kube-state-metrics_ chart ServiceMonitor that are present in the override ServiceMonitor in the _kube-prometheus-stack_ chart with the intention that the _kube-prometheus-stack_ should use the _kube-state-metrics_ chart ServiceMonitor.

This PR is the first PR of 3 for #1425.

I've not implemented the template logic from the _kube-prometheus-stack_ as it's not a valid pattern, only strings should be run through the `tpl` command. It's possible to implement it correctly here if required and I'm happy to do so.

I've bumped the chart minor version as this is a feature release.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.0.2
+version: 4.1.0
 appVersion: 2.2.4
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/servicemonitor.yaml
+++ b/charts/kube-state-metrics/templates/servicemonitor.yaml
@@ -6,38 +6,57 @@ metadata:
   namespace: {{ template "kube-state-metrics.namespace" . }}
   labels:
     {{- include "kube-state-metrics.labels" . | indent 4 }}
-    {{- if .Values.prometheus.monitor.additionalLabels }}
-{{ toYaml .Values.prometheus.monitor.additionalLabels | indent 4 }}
-    {{- end }}
+  {{- with .Values.prometheus.monitor.additionalLabels }}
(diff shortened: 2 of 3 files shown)


stevehipwell:
@monotek if you're happy with this approach I'll do the _prometheus-node-exporter_ chart too. Once both have been merged I'll open a PR for the _kube-prometheus-stack_ to use the sub-chart ServiceMonitors.

charts/kube-state-metrics/Chart.yaml
```diff
@@ -2,18 +2,18 @@ apiVersion: v2
 name: kube-state-metrics
 description: Install kube-state-metrics to generate and expose cluster-level metrics
 keywords:
-- metric
-- monitoring
-- prometheus
-- kubernetes
+  - metric
```
mrueg, line 5: could we leave those formatting changes out? Unless there's a general decision to format all charts in a specific way, let's leave it as it is.

stevehipwell: Could I suggest adding something like an EditorConfig spec into the repo to keep formatting consistent?

mrueg: That sounds like a great idea, maybe also some sort of yamllint to enforce it during CI. :) 
(I'm not opposed to the change, just would want to have the same style enforced everywhere)

stevehipwell: I'll look at undoing the change here to get this merged and can look at opening another PR for the formatting if you'd like?

mrueg: Yeah, ideally start with an issue that describes what you'd like to set in terms of standards and how to enforce them on this repo.

stevehipwell: @mrueg I've reverted the _Chart.yaml_ formatting.

Issue #467, opened 2020-12-09
[kube-prometheus-stack] node-exporter crashes at startup


DaviGia:
**Describe the bug**
Node exporter pod underlying container fails to startup. This is the error reported by the deamon:

``` console
Error: failed to start container "node-exporter": Error response from daemon: path / is mounted on / but it is not a shared or slave mount
```

This error is related to #325 and it appears to be caused by this [pull request](https://github.com/prometheus-community/helm-charts/pull/80#issuecomment-728716134). I've create this issue as suggested by @monotek in https://github.com/prometheus-community/helm-charts/issues/80#issuecomment-728922569 (it appears that the original reporter of this issue has not created a new issue).

Moreover, I've tried to change kube-prometheus-stack chart manually to revert the commit above and I am able to deploy the chart without any error.

I am using `docker-desktop` (v2.5.0.1) on Windows 10 1909 and I have enabled Kubernetes cluster from it.

I'm not sure if this error is caused by `docker-desktop`, in any case, if it does not happen in other type of clusters, I would suggest to make the **mount rootfs** *mountPropagation* configurable via the chart values (and maybe enable it by default if you think it should not be optional) to prevent this error.

Another (better) solution would be to mount / as shared, but I've not found a way to do it.

**Version of Helm and Kubernetes**:

Helm Version:

```console
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.14.11"}
```

Kubernetes Version:

```console
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.1", GitCommit:"7879fc12a63337efff607952a323df90cdc7a335", GitTreeState:"clean", BuildDate:"2020-04-08T17:38:50Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

**Which chart**: kube-prometheus-stack

**Which version of the chart**: 12.7.0

**What happened**:  node-exporter pod fails to startup

**What you expected to happen**: node-exporter should not fail

[...]

PR #1583, opened 2021-12-09
[prometheus-node-exporter] Parameterise host root FS mount propagation

weibeld:
Signed-off-by: Daniel Weibel <danielmweibel@gmail.com>

The mounting of the node's root (`/`) file system was enabled by default in PR #80 and it was made optional in PR #757. However, in the state of PR #757, the [mount propagation](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation) is still hardcoded to HostToContainer. However, this feature requires certain [settings](https://kubernetes.io/docs/concepts/storage/volumes/#configuration) in the container runtime to work properly. Without these settings, the container may fail to start up with an error of the following form:

```
Error: failed to start container "node-exporter": Error response from daemon: path / is mounted on / but it is not a shared or slave mount
```

This PR allows to freely define the mount propagation type in the values file:

```yaml
hostRootFsMount:
  enabled: true
  mountPropagation: HostToContainer
```

In this way, it's possible to disable mount propagation when mounting the host root file system (by setting the field to None or omitting it), which allows to mount the host root file system on nodes that don't support mount propagation.

The default mount propagation value is set to HostToContainer to maintain backward compatibility with older versions of the chart.

Various reports of the default mount propagation type causing the container to crash at startup can be found in #467.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -92,9 +92,15 @@ hostNetwork: true
 # Share the host process ID namespace
 hostPID: true
 
-## If true, node-exporter pods mounts host / at /host/root
-##
-hostRootFsMount: true
+# Mount the node's root file system (/) at /host/root in the container
+hostRootFsMount:
+  enabled: true
+  # Defines how new mounts in existing mounts on the node or in the container
+  # are propagated to the container or node, respectively. Possible values are
+  # None, HostToContainer, and Bidirectional. If this field is omitted, then
+  # None is used. More information on:
+  # https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
+  mountPropagation: HostToContainer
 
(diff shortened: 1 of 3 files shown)


charts/prometheus-node-exporter/values.yaml
```diff
@@ -92,9 +92,15 @@ hostNetwork: true
 # Share the host process ID namespace
 hostPID: true
 
-## If true, node-exporter pods mounts host / at /host/root
-##
-hostRootFsMount: true
+# Mount the node's root file system (/) at /host/root in the container
+hostRootFsMount:
```
monotek, line 96: this is a breaking change which would need to raise major version of the chart.

monotek (changes requested):

monotek (dismissed):

PR #1617, opened 2021-12-20
reintroduce [prometheus-node-exporter] and [kube-state-metrics] service monitor selectorOverride

jurgen-weber-deltatre:
I am bringing back the service monitor selectorOverride option that were removed in v24 (https://github.com/prometheus-community/helm-charts/commit/c75ec1c8ead17a745b9e141da4ebd51922626794) when deprecating the kube-prom-stack Service Monitors in favour of the tools specific chart Service Monitors.

While I could not find the old issue from many years ago, this override was introduced to solve problems with various deployment tools that take over labels. In our case ArgoCD which changes the value the instance label. Once I upgraded to 24 I realised I lost a lot of metrics and that is because the service monitor labels were all wrong and I could find no equivalent new field.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.lock
+++ b/charts/kube-prometheus-stack/Chart.lock
@@ -1,10 +1,10 @@
 dependencies:
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
-  version: 4.2.0
+  version: 4.2.1
 - name: prometheus-node-exporter
   repository: https://prometheus-community.github.io/helm-charts
-  version: 2.4.0
+  version: 2.4.1
 - name: grafana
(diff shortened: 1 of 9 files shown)


jurgen-weber-deltatre:
not sure how to resolve the lint test issue there, trying locally I get the same issue. Do I need to split the kube-state-metrics and node-exporter updates from the kube-prom-stack update into 2 PR's?

mrueg:
Yes, please split those changes up into one PR per helm chart

jurgen-weber-deltatre:
As you can imagine, I have been on holidays. I will split all of this up this week.

PR #1626, opened 2021-12-22
[kube-prometheus-stack] make common part of runbook_url overridable

hairmare:

Make it possible to override the common part of the `runbook_url`s in the provided prometheus rules. The upstream mixin allows doing this and it would be nice to have the same feature available to users of `kube-prometheus-stack`.

I decided to default the runbookUrlPrefix in the individual yaml files so i could keep the magic `https://runbooks.prometheus-operator.dev/runbooks/` string in one place but i'd be happy to move the default to values.yaml if requested to do so.

I also considered using some `helm template | kustomize` magic to acheive the same results. This change feels like it has less tradeoffs.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 25.1.0
+version: 25.2.0
 appVersion: 0.53.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -69,6 +69,9 @@ defaultRules:
   ## Additional labels for PrometheusRule alerts
   additionalRuleLabels: {}
 
+  ## Prefix for runbook URLs. Use this to override the first part of the runbookURLs that is common to all rules. Defaults to `https://runbooks.prometheus-operator.dev/runbooks/` if not specified.
+  runbookUrl: ""
+
 ## Deprecated way to provide custom recording or alerting rules to be deployed into the cluster.
 ##
 # additionalPrometheusRules: []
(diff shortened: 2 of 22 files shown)


mrueg:
If folks are using it and willing to support it, let's add it back. :) Can you add a few additional words to the changelog https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/README.md#from-24x-to-25x ?

hairmare:
i changed it to use the same value that was previously used and will update the changelog towmorrow.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -69,6 +69,9 @@ defaultRules:
   ## Additional labels for PrometheusRule alerts
   additionalRuleLabels: {}
 
+  ## Prefix for runbook URLs
+  runbookUrl: ""
```
eyenx, line 73: As it looks like, the runbookUrl needs to end with a trailing slash. Should we add this to the comment above?

Xtigyro: yeah - I think we should.

hairmare: The new default runbookUrl needs to end in a trailing slash. If you override it there are a lot of possibilities some of which would not need to end in a trailing slash. The following would both build perfectly valid URLs:

* `https://example.com/path/to/endpoint?alert=`
* `https://example.com/path/to/html-page#`

hairmare:
I added info to the changelog, reworded the comment in `values.yaml` and rebased this PR.

monotek (changes requested):
Why is the trailing slash not just added to the template?

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -17,7 +17,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 25.1.0
+version: 25.2.0
```
monotek, line 20:
```suggestion
version: 25.3.0
```

monotek:
@hairmare could yuo please rebase / fix the merge conflict?
I cant push to your branch...

monotek:
superseeded by: https://github.com/prometheus-community/helm-charts/pull/1646

PR #1665, opened 2022-01-04
[kube-state-metrics] Fix code duplication

skhalash:
Signed-off-by: Stanislav Khalash <stanislav.khalash@sap.com>

- Unifies the clusterrole labels with the rest of the chart resources by using the `kube-state-metrics.labels` template

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.2.0
+version: 4.2.1
 appVersion: 2.3.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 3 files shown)


monotek (dismissed):

mrueg:
Have you tested this with namespaces and roles as well? 
I believe this code is not duplicated, as it's using different scoping.

skhalash:
@mrueg Good hint, I tested it and found a bug in the `main` version. When setting `rbac.useClusterRole=false` the chart from `main` is not renderable. The problem is the `rolebinding` which uses incorrect scoping.

skhalash:
@mrueg Could you please take a look? I compared the `helm template` output with the `main` branch and it produces the same set of labels. The only difference is `app.kubernetes.io/version`,  which is now set to the app version instead of the chart version like in the rest of the resources.

mrueg:
Thanks for the fix! Looks good to me. :)

mrueg (approved):

PR #1673, opened 2022-01-05
[prometheus-pushgateway] Use a map for the extraArgs value in the pushgateway chart, handle alternate web prefix

devec0:
This PR allows the use of an alternate web prefix, using the same mechanism as the standalone prometheus chart to conditionally modify the liveness and readiness checks on the pushgateway deployment if an alternate web root is configured.

This is required as otherwise, the checks fail with a 404 error using the current hard-coded check paths, as these paths must change if the web prefix is customised.

Signed-off-by: James Hebden <james@ec0.io>

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-pushgateway/Chart.yaml
+++ b/charts/prometheus-pushgateway/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.4.2"
+appVersion: "1.5.2"
 description: A Helm chart for prometheus pushgateway
 name: prometheus-pushgateway
 version: 1.14.0
--- a/charts/prometheus-pushgateway/templates/deployment.yaml
+++ b/charts/prometheus-pushgateway/templates/deployment.yaml
@@ -40,7 +40,14 @@ spec:
         {{- end }}
         {{- if .Values.extraArgs }}
           args:
-{{ toYaml .Values.extraArgs | indent 12 }}
+          {{- range $key, $value := .Values.extraArgs }}
+          {{- $stringvalue := toString $value }}
+          {{- if eq $stringvalue "true" }}
+            - --{{ $key }}
+          {{- else }}
+            - --{{ $key }}={{ $value }}
+          {{- end }}
+          {{- end }}
         {{- end }}
           ports:
             - name: metrics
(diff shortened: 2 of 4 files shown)


monotek (changes requested):
Imho web.route-prefix should not be a part of extraArgs type should not be changed with a bugfix version as this would be a breaking change.

devec0:
> Imho web.route-prefix should not be a part of extraArgs and extraArgs type should not be changed with a bugfix version as this would be a breaking change.

Thank you for looking at these changes @monotek! 

I've fixed the version number to better represent that project's versioning.

I also agree with your thought that the usage of the extraArgs map for controlling the prefix is awkward. I implemented it this way however to ensure consistency between this and the pushegateway configuration in the standalone prometheus chart in this repository. In my opinion, it would be cleaner to simply have a `webPrefix:` value (potentially under a `pushgateway` top level key) which was checked conditionally in the `deployment.yaml` and `servicemonitor.yaml` templates and its value used - in the `deployment.yaml`, the value could be checked and then the appropriate argument injected into the args list for the container, then any extraArgs (in list form, rather than map) could be appended if specified.

I'm happy to go either way, if you're prefer I implement it in the way I've described above, I can absolutely do that.

monotek:
> > Imho web.route-prefix should not be a part of extraArgs and extraArgs type should not be changed with a bugfix version as this would be a breaking change.
> 
> Thank you for looking at these changes @monotek!
> 
> I've fixed the version number to better represent the project's versioning.
> 
> I also agree with your thought that the usage of the extraArgs map for controlling the prefix is awkward. I implemented it this way however to ensure consistency between this and the pushegateway configuration in the standalone prometheus chart in this repository. In my opinion, it would be cleaner to simply have a `webPrefix:` value (potentially under a `pushgateway` top level key) which was checked conditionally in the `deployment.yaml` and `servicemonitor.yaml` templates and its value used - in the `deployment.yaml`, the value could be checked and then the appropriate argument injected into the args list for the container, then any extraArgs (in list form, rather than map) could be appended if specified.
> 
> I'm happy to go either way, if you're prefer I implement it in the way I've described above, I can absolutely do that.

Sounds good :)

PR #1674, opened 2022-01-05
[kube-prometheus-stack]: image digest pinning

Dentrax:
Image tags can’t always be relied on to point to a consistent specific image, so a common piece of advice is to use SHA-256 hashes to identify your images. [^1]

Probably it's nice to have to maintain image `digest` for each version that we bump'ed.

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

[^1]: https://blog.aquasec.com/docker-image-tags

  - fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -405,7 +405,7 @@ alertmanager:
     image:
       repository: quay.io/prometheus/alertmanager
       tag: v0.23.0
-      sha: ""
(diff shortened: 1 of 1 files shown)


monotek (changes requested):
Not sure if this should be default as it might be error prone updating the images, if you have to set two variables from now on.

I guess this would also break for people already using a non default tag but no sha, if they are updating the chart and keeping their values.

Normally using no sha should not be a problem as long as "latest" is not used as an image tag, something with the docker image build pipline is wrong or somebody is doing something bad...

PR #1686, opened 2022-01-06
[kube-prometheus-stack] PodSecurityPolicy apiVersion upgrade

andrewgkew:
As of k8s 1.21 the PodSecurityPolicy CRD apiVersion is now policy/v1 so updating that, and using an already created helper to ensure this change is backwards compatible

  - fixes #
Warnings being thrown when installing the chart which will become errors as of 1.25

#### Checklist

- [x] Chart Version bumped


andrewgkew:
Misread the documentation, sorry this is being deprecated completely and removed and we need to use an OPA instead. Will close this and look into that instead.

PR #1688, opened 2022-01-07
[kube-prometheus-stack] Removed outdated template

amuraru:

Grafana dashboards are now yaml templates and not raw jsons.
Removing the old template.

#### Checklist

- [ ] Chart Version bumped


monotek (changes requested):
What makes you believe that?
Somebody could still place a dashboard there using a json installing locally....

amuraru:
> Somebody could still place a dashboard there using a json installing locally....

Hmm - is this something for testing? The user can make use of standard ConfigMap based dashboards right?

monotek:
But taking it away might stil be a breaking change for some users.

PR #1700, opened 2022-01-11
[kube-prometheus-stack] Increase default CPU for config-reloader

### LINKED ISSUE #1587: [kube-prometheus-stack] High CPU throttling for config-reloader after v22/v23 upgrade

After upgrading from v21 to v23, the config-reloader container inside the Prometheus pod began experiencing high CPU throttling:
image

We don't specify a custom value for the config-reloader container, so it still had its default 100m CPU limit before and after the upgrade.

### What's your helm version?

v3.7.1

v1.22.1

### Which chart?

kube-prometheus-stack

### What's the chart version?

23.1.5

### What happened?

Upgrading to v23 caused higher-than-normal CPU throttling.

No spike in CPU usage.

### How to reproduce it?

1. On a large-ish cluster (13 nodes, 230 pods), install kube-prometheus-stack v21
2. Observe little-to-no CPU throttling on the config-reloader container
3. Upgrade to v23
4. Observe CPU throttling on the config-reloader container

Relevant section:
```
kube-prometheus-stack:
  prometheus:
    prometheusSpec:
      ruleSelector: {}
      ruleSelectorNilUsesHelmValues: false
      serviceMonitorSelector: {}
      serviceMonitorSelectorNilUsesHelmValues: false
      retention: 90d
      resources:
        requests:
          cpu: 2500m
          memory: 4000Mi
        limits:
          cpu: 2500m
          memory: 4000Mi
```

`helm upgrade --install -n monitoring monitoring . -f values.yaml`

I know that the values for config-reloader did change slightly, but I don't think that changed the default CPU limit of 100m, that's always been there IIRC.

And I don't think config-reloader itself changed substantially recently: https://github.com/prometheus-operator/prometheus-operator/commits/main/cmd/prometheus-config-reloader

Maybe some other part of the ecosystem changed that could make config-reloader's job more CPU-intensive?

mac-chaffee:

See #1587 for context. Essentially, after the v23 upgrade, config-reloader seems to be doing more CPU work which causes it to experience CPU throttling.

image

\>25% CPU throttling triggers [CPUThrottlingHigh alerts from kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/resource_alerts.libsonnet#L143), so this PR fixes that issue.

  - fixes #1587 

#### Checklist

- [x] Chart Version bumped
    - **NOTE:** I bumped the minor version instead of the patch version since this update will the prometheus statefulset to rollout new pods. Let me know if this is incorrect


desaintmartin:
I've seen a small increase as well. My take: it would be better to have higher limits (say 1 cpu), since average usage on a 3000-containers cluster is closer to 0.005 cpu than 0.2 cpu.

arthurk:
I've also noticed CPU throttling with the default values, but only between 10%-20% in my case. I think increasing them a little would be good

Xtigyro (approved):
Thanks for the contribution, @mac-chaffee! 🥇 

LGTM.

monotek:
@Xtigyro 
Why was this merged as an feature release and without changing the values first to something lower?

The requested memory should now be way above the actual usage for most users, which wastes ram. Do you need this much requests in your cluster?

Xtigyro:
@monotek 

Feel free to tweak it further if you want to. I think it's a reasonable increase and `200m` is not a high value.

Xtigyro:
@monotek 
Apologies if the new value is not what you thought was right.
Feel free to tweak it further if you want to.
I think it's a reasonable increase and 200m is not that high but we can always improve it more.

PR #1705, opened 2022-01-12
kube-prometheus-stack: Add missing tolerations field

erikkn:

Just nitpicking I guess, but the `kube-prometheus-stack` Helm chart is missing `tolerations` for Grafana & kube-state-metrics field.

#### Checklist

- [ x] Chart Version bumped


monotek (changes requested):
The filed is not missing. 
It's just not used and therefore there is no reason to add it to the subchart config.
Only things which are different to the defaults should be in the subchart config.

PR #1706, opened 2022-01-12
[prometheus-mongodb-exporter] Add volumes support

realrill:
PR gives option to attach volumes from various sources (configMap, secret, etc) to the running container. It can also reduce complexity for consumers in many use cases.

In my case it's a secret that contains a certificate. This certificate then used in the URI `?option` part to secure communication.

Without this, I need to use this chart as dependency and overwrite the deployment simply to support volumes management.
#### Checklist

- [ ] Chart Version bumped


realrill:
Hi @steven-sheehy can you please take a look?

steven-sheehy (approved):
LGTM

PR #1707, opened 2022-01-12
[kube-prometheus-stack] Allow using template for additionalScrapeConfigs

BulatSaif:
1. It will be very useful to be able to template content of `additionalScrapeConfigs` variable. Currently, `tpl` allows only changing the string value in yaml not yaml itself.
**Old behavior:**
```
    additionalScrapeConfigs:
      - job_name: "node-exporter"
        gce_sd_configs:
        - project: "project1"
          zone: "zone1"
          port: 9100
        - project: "project1"
          zone: "zone2"
          port: 9100
        - project: "project1"
          zone: "zone3"
          port: 9100
        ...
```
**New behavior:**
```
    additionalScrapeConfigs: |
      - job_name: "node-exporter"
        gce_sd_configs:
         {{range $zone := .Zones}} 
        - project: "project1"
          zone: "{$zone}}"
          port: 9100
         {{end}}
        ...
```
 
#### Checklist

- [x] Chart Version bumped


desaintmartin:
Thanks!
This is a breaking change as one has to re-write all already-defined `additionalScrapeConfigs`.

As such, I believe we should have several maintainers give their opinion. On my part, I do not see particular downsides (I would be happy if one can have "mergeable" values which is not possible because new type is string, but it was not previously possible because old type is list).

charts/kube-prometheus-stack/templates/prometheus/additionalScrapeConfigs.yaml
```diff
@@ -12,5 +12,9 @@ metadata:
     app: {{ template "kube-prometheus-stack.name" . }}-prometheus-scrape-confg
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
 data:
-  additional-scrape-configs.yaml: {{ tpl (toYaml .Values.prometheus.prometheusSpec.additionalScrapeConfigs) $ | b64enc | quote }}
```
monotek, line 15: the posssibility to use string ot tpl should be reflected in some info in the values.yaml too.

monotek (changes requested):

charts/kube-prometheus-stack/templates/prometheus/additionalScrapeConfigs.yaml
BulatSaif: Thanks for reviewing,  `values.yaml` updated. 
If you have ideas what can be improved in PR, please let me know.

BulatSaif:
Can someone please review and merge it?

Xtigyro (approved):
Thanks for the contribution, @BulatSaif! 🥇

LGTM.

PR #1710, opened 2022-01-13
[prometheus-adapter] adding maxage 

AlonGluz:
- Signed-off-by: Alon Gluzman <alon.gluzman@torq.io>
- Signed-off-by: Alon Gluzman <alon.gluzman@torq.io> bump version
- useful when using Stackdriver or any other monitoring tool

Adds --metrics-max-age flag to deployment.yaml so it will be easier to understand why external metrics (Stackdriver) are missing from the kubectl get --raw "/apis/external.metrics.k8s.io/v1beta1" request 
  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.0.1
+version: 3.0.2
 appVersion: v0.9.1
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.0.0
+version: 3.0.1
 appVersion: v0.9.1
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.0.2
+version: 3.1.0
 appVersion: v0.9.1
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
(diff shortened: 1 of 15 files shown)


charts/prometheus-adapter/values.yaml
```diff
@@ -10,6 +10,9 @@ logLevel: 4
 
 metricsRelistInterval: 1m
 
+# https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/README.md#Configuration
+# metricsMaxAge: 2m
```
monotek, line 14:
```suggestion
metricsMaxAge: ""
```

monotek (changes requested):

charts/prometheus-adapter/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.0.1
+version: 3.0.2
```
monotek, line 3: new feature should raise minor version

monotek (changes requested):

monotek (changes requested):
Please sign your commits, to get DCO check pass.

AlonGluz:
@monotek just did, don't know why the DCO is failing

monotek:
https://github.com/prometheus-community/helm-charts/pull/1710/checks?check_run_id=4802830313

zanhsieh (approved):

charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -55,6 +55,9 @@ spec:
         - --logtostderr=true
         - --prometheus-url={{ tpl .Values.prometheus.url . }}{{ if .Values.prometheus.port }}:{{ .Values.prometheus.port }}{{end}}{{ .Values.prometheus.path }}
         - --metrics-relist-interval={{ .Values.metricsRelistInterval }}
+        {{- if .Values.metricsMaxAge }}
+        - --metrics-max-age={{ .Values.metricsMaxAge }}
```
steven-sheehy, line 59: Why do we need an explicit value for this when we have `extraArguments` that you can use to set `--metrics-max-age=`? I'm not in favor of more properties when we have an existing mechanism to support your use case.

AlonGluz: IMHO I think this metric is important because it comes hand in hand with External metrics. If the Helm chart should work out of the box with all the API's it is important that this property will be explicitly listed. 
It took me a while to understand why the Stack driver metrics don't appear even if all is working properly

steven-sheehy: I'm not arguing whether the flag is useful or not, I'm arguing that you can already configure the current chart with that flag. Since you're adding a disabled by default flag there is little difference in people discovering the flag in the `values.yaml` or finding in the official Prometheus Adapter [documentation](https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/README.md#configuration). It does not add any value and I'm a proponent of keeping things simple.

PR #1716, opened 2022-01-17
[prometheus-json-exporter] Fix typo in deployment nodeselector 

roelvdberg:
There is a random bracket in the deployment of the json exporter which blocks the usage of the nodeselector in the prometheus json exporter

  - fixes #

#### Checklist

- [ ] Chart Version bumped

@schmiddim


schmiddim (approved):

schmiddim (approved):
Thanks

schmiddim (changes requested):
Thanks for your fix!
could you please increase 
- version to 0.1.1
- appVersion to "1.0.3"
in *values.yaml*?

schmiddim (approved):
Could you also check the Boxes in the PR? This is my first approve in this repo.

charts/prometheus-json-exporter/Chart.yaml
```diff
@@ -19,13 +19,13 @@ type: application
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.1.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.0.2"
+appVersion: "1.0.3"
```
monotek, line 28: why is this changed if the used docker image is the same?

monotek: I guess 0.3.0 would be the correct value here?
```suggestion
appVersion: "0.3.0"
```

monotek: and image tag could be renmoved, so app version is used: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-json-exporter/values.yaml#L11

rouke-broersma: This seems like an improvement that should be picked up separately no?

monotek: @rouke-broersma why? IMHO it's all Bugfix scope so no need to do it in another pr.

@schmiddim Take your time and get well soon :)

roelvdberg: I have changed the version number

monotek (changes requested):
can we please also use an empty var here https://github.com/prometheus-community/helm-charts/blob/c0f13f952c83137e7a8d8f867020334d759aa513/charts/prometheus-json-exporter/values.yaml#L11  

so the default from the Charty.yaml can be used.
See: https://github.com/prometheus-community/helm-charts/blob/c0f13f952c83137e7a8d8f867020334d759aa513/charts/prometheus-json-exporter/templates/deployment.yaml#L34

roelvdberg:
I removed the tag and updated the appVersion to include the v.

charts/prometheus-json-exporter/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   repository: quay.io/prometheuscommunity/json-exporter
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "v0.3.0"
+  # tag: "v0.3.0"
```
monotek, line 11:
```suggestion
  tag: ""
```

monotek (changes requested):

roelvdberg:
Are there any other changes before I do another attempt to fix the signoff? 

I wanted to fix one single typo and I get the idea that I am now responsible for for fixing a whole version number issue which I am not familiar with. I would suggest to do this in another issue the next time so we can move forward with the initial fix.

schmiddim:
Hi, just fix the Bracket issue - I will clean the version mess up.

roelvdberg:
Ok, I have updated the chart version and fixed the typo.

mrueg (dismissed):

monotek (approved):

PR #1760, opened 2022-02-01
[prometheus-blackbox-exporter] Added runAsGroup to Deployment

shaardie:
Signed-off-by: Sven Haardiek <sven.haardiek@uni-muenster.de>

It is possible to change the `runAsUser`, but not the `runAsGroup` in the deployment. So despite the fact that the Pod is running as the user 1000, it is still running with the root group. This pull request makes in configurable and also set the default to 1000. Same id as the user.

I do not this there is an issue about it.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
 version: 5.3.2
-appVersion: 0.19.
+appVersion: 0.19.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
   - https://github.com/prometheus/blackbox_exporter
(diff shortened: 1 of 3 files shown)


rsotnychenko (approved):
@shaardie thanks for the contribution!

LGTM

charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 5.3.1
-appVersion: 0.19.0
+version: 5.3.2
+appVersion: 0.19.
```
rsotnychenko, line 5: Why do we need this change?

rsotnychenko: Thanks! I was asking only about `appVersion`, sorry for the misunderstanding

rsotnychenko (approved):

PR #1786, opened 2022-02-09
[prometheus-mysql-exporter] Add missing namespaceSelector to servicemonitor

leo108:

Add missing namespaceSelector to MySQL exporter chart

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-mysql-exporter/templates/servicemonitor.yaml
@@ -12,6 +12,9 @@ metadata:
   namespace: {{ .Values.serviceMonitor.namespace }}
 {{- end }}
 spec:
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
   selector:
     matchLabels:
       {{- include "prometheus-mysql-exporter.selectorLabels" . | nindent 6 }}
--- a/charts/prometheus-mysql-exporter/Chart.yaml
+++ b/charts/prometheus-mysql-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for prometheus mysql exporter with cloudsqlproxy
 name: prometheus-mysql-exporter
-version: 1.5.0
+version: 1.6.0
 home: https://github.com/prometheus/mysqld_exporter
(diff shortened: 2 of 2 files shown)


monotek (changes requested):
what about?

 https://github.com/prometheus-community/helm-charts/blob/9533608d204129e479fee86300a62a4d6e527751/charts/prometheus-mysql-exporter/templates/servicemonitor.yaml#L15-L17

leo108:
@monotek I believe all other charts in this repo is using that code, e.g.

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-redis-exporter/templates/servicemonitor.yaml#L38-L41
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-elasticsearch-exporter/templates/servicemonitor.yaml#L37-L40

I think it's better to make the code consistent

monotek (changes requested):
there is no convention in the repo how every chart has to look like.
but this chart uses already templated labels, what should be the way to go.

my initial point was, that you have declared it a second time.

leo108:
@monotek Fixed, stupid error.

charts/prometheus-mysql-exporter/templates/servicemonitor.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
   namespace: {{ .Values.serviceMonitor.namespace }}
 {{- end }}
 spec:
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
```
monotek, line 17: IMHO this should be configurable. 
Otherwise you still can only use it in the same namespace it's installed too.


See: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#namespaceselector

monotek (changes requested):

charts/prometheus-mysql-exporter/templates/servicemonitor.yaml
leo108: `namespaceSelector` is used for finding metric exporter, it should be always the same as the namespace the chart installed, I don't think it needs to be customized

monotek (approved):

PR #1811, opened 2022-02-18
[kube-prometheus-stack] Expose the container-specific security context for Prometheus Operator

rdauncey53:
We want to further configure the Kubernetes container security context for the Prometheus Operator, so we can set other container-specific settings such as capabilities -> drop -> ALL. Therefore we expose the container-specific security context using `containerSecurityContext` as the input (as used in other prometheus helm-charts e.g. [node-exporter](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/values.yaml#L79)).

We understand that some other Prometheus containers such as the node-exporter do not support dropping capabilities (in that case, because it did not have any in the first place) - this pull request assumes that isn't the case for Prometheus Operator, but please do let us know if that is an incorrect assumption.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 32.2.2
+version: 32.3.0
 appVersion: 0.54.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 32.2.1
+version: 32.2.2
```
monotek, line 26: new feature should raise minor version
```suggestion
version: 32.3.0
```

monotek (changes requested):

charts/kube-prometheus-stack/Chart.yaml
rdauncey53: Sure, have done that.

rdauncey53:
(28/2/2022 Fixed some merge conflicts)

monotek (approved):

Issue #1720, opened 2022-01-18
[kube-prometheus-stack] Allow to set grafana datasource UID


TeddyAndrieux:
### Is your feature request related to a problem ?

No

### Describe the solution you'd like.

Since we should use UID to reference a Grafana Datasource from a Grafana dashboard, could be cool to be able to set the UID for Prometheus Datasource created from the helm chartPrometheus. (also rework a bit the layout so that we can set some variables for "default Datasource" and "replicas Datasource")

From helm option file being able to do something like that:
```yaml
grafana:
  sidecar:
    datasources:
      default:
        enabled: true
        uid: my-default-prometheus-ds
```

### Describe alternatives you've considered.

As an alternative solution with today helm chart we can just do it by disabling the default Datasource and creating an "additional" one with the UID

```yaml
grafana:
  sidecar:
    datasources:
      defaultDatasourceEnabled: false

  additionalDataSources:
    - name: Prometheus
      uid: my-default-prometheus-ds
      type: prometheus
      access: proxy
      isDefault: true
      url: <prometheus service URL>
      jsonData:
        timeInterval: 30s
```

TeddyAndrieux:
:up: (I may have time to do a PR for this next week)

TeddyAndrieux:
:up: I will I will, I just not have time right now

PR #1816, opened 2022-02-21
[kube-prometheus-stack] Add uid to grafana default datasources

Keramblock:
Signed-off-by: Michael Kirpichev <m.kirpichev@haut.ai>
Today I tried to migrate to new grafana alerting and noticed following diff in one of my dashboards:
Screenshot 2022-02-21 at 15 17 08
So now grafana relates on uid to reference data source from dashboard. To avoid any potential issues I think that is a good idea to fix uid of default data sources. I dont think this change will affect anybody with old alerts and will ease migration to new system.

P.S. 
Grafana dashboard migration results after uid was specified:
изображение

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
+++ b/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
@@ -37,8 +37,8 @@ data:
 {{- if .Values.grafana.sidecar.datasources.createPrometheusReplicasDatasources }}
 {{- range until (int .Values.prometheus.prometheusSpec.replicas) }}
     - name: Prometheus-{{ . }}
-      uid: promeheus-replica-{{ . }}
       type: prometheus
+      uid: promeheus-replica-{{ . }}
       url: http://prometheus-{{ template "kube-prometheus-stack.fullname" $ }}-prometheus-{{ . }}.prometheus-operated:9090/{{ trimPrefix "/" $.Values.prometheus.prometheusSpec.routePrefix }}
       access: proxy
       isDefault: false
--- a/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
+++ b/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
@@ -24,6 +24,7 @@ data:
 {{- if .Values.grafana.sidecar.datasources.defaultDatasourceEnabled }}
     - name: Prometheus
       type: prometheus
+      uid: promeheus
       {{- if .Values.grafana.sidecar.datasources.url }}
       url: {{ .Values.grafana.sidecar.datasources.url }}
       {{- else }}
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
```diff
@@ -24,6 +24,7 @@ data:
 {{- if .Values.grafana.sidecar.datasources.defaultDatasourceEnabled }}
     - name: Prometheus
       type: prometheus
+      uid: promeheus
```
hlomzik, line 27: I guess it should be "prometheus" with "t" missing. And also in the second uid below

Keramblock: Thx! gixed

TeddyAndrieux:
Good I was about to open a PR to do this :+1: 

If I may, @Keramblock could you make those UID configurable please see #1720 (otherwise I can open another PR to do this)

Keramblock:
> Good I was about to open a PR to do this 👍
> 
> If I may, @Keramblock could you make those UID configurable please see #1720 (otherwise I can open another PR to do this)

I am not sure if this is good idea to make only UID configurable in that PR, because:

1. It is defaults and we can make our own, if we need something else
2. Other fields of it are not parametrised
3. In guidlines there is line that we need to create PR that solves only one problem at a time, so we need another one for that problem
4. I am not sure if maintainers will approve such change

TeddyAndrieux:
> I am not sure if this is good idea to make only UID configurable in that PR, because:

> 1. It is defaults and we can make our own, if we need something else

I agree we can, kind of, work around it, but we may want to just override the default UID so that we don't need to "know" the content of the Datasource config file when we apply an helm chart

> 2. Other fields of datasource are not parametrised

Agree but we need to start somewhere :smiley: and I think this field makes sense since we can use it in dashboard

> 3. In guidlines there is line that we need to create PR that solves only one problem at a time, so we need another one for that problem

Fine, I can do another PR for it, don't worry was just asking since it related but never mind

> 4. I am not sure if maintainers will approve such change

I think/hope they will :)

TeddyAndrieux:
> I am not sure if this is good idea to make only UID configurable in that PR, because:

> 1. It is defaults and we can make our own, if we need something else

I agree we can, kind of, work around it, but we may want to just override the default UID so that we don't need to "know" the content of the Datasource config file when we apply an helm chart

> 2. Other fields of datasource are not parametrised

Agree but we need to start somewhere :smiley: and I think this field makes sense since we can use it in dashboard

> 3. In guidlines there is line that we need to create PR that solves only one problem at a time, so we need another one for that problem

Fine, I can do another PR for it, don't worry was just asking since it related but never mind

> 4. I am not sure if maintainers will approve such change

I think/hope they will :)

Xtigyro (dismissed):

Xtigyro (approved):

Xtigyro:
Thanks for the contribution, @Keramblock! 🥇 

LGTM.

PR #1833, opened 2022-02-26
[kube-prometheus-stack] Replace invalid characters in the version

dpiessens:
Fixes an issue if this chart is wrapped and given a version that is on a PR or branch with a +. The chart ID is protected but the main version is not.


dpiessens:
Hi it looks like the title was changed, this is causing us issues with chart wrapping. Could it be reviewed and merged if acceptable?

Xtigyro (approved):
Thanks for the contribution, @dpiessens! 🥇 

LGTM.

PR #1909, opened 2022-03-24
[prometheus] Make it possible to disable the Service for Prometheus server

mdonkers:

When Prometheus is running in 'Agent' mode (via
`--enable-feature=agent`) there is no need to create the Kubernetes
`Service` because Prometheus won't run any reachable service endpoints.

Therefore make it possible to disable the service explicitly, but
enabled by default like currently.

#### Checklist

- [ ] Chart Version bumped


zanhsieh (approved):

charts/prometheus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: 2.34.0
-version: 15.5.4
+version: 15.5.5
```
monotek, line 4: new feature should raise minor version

monotek (changes requested):

charts/prometheus/Chart.yaml
mdonkers: :+1: Changed version to `15.6.0`

monotek (approved):

PR #1915, opened 2022-03-25
[alertmanager] Allow exposing configmap reloaders port

Sadzeih:
Needed for scraping reloader metrics

Signed-off-by: Alexis Guerville <alexis.guerville@messagebird.com>

This allows exposing ports on the configmapreloaders containers. It is necessary to allow scraping the reloaders metrics.

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -62,9 +62,9 @@ spec:
             - --webhook-url=http://127.0.0.1:{{ .Values.service.port }}/-/reload
           resources:
             {{- toYaml .Values.configmapReload.resources | nindent 12 }}
-          {{- if .Values.configmapReload.port }}
+          {{- if .Values.configmapReload.containerPort }}
           ports:
-            - containerPort: {{ .Values.configmapReload.port }}
+            - containerPort: {{ .Values.configmapReload.containerPort }}
           {{- end }}
           volumeMounts:
             - name: config
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -62,6 +62,10 @@ spec:
             - --webhook-url=http://127.0.0.1:{{ .Values.service.port }}/-/reload
           resources:
             {{- toYaml .Values.configmapReload.resources | nindent 12 }}
+          {{- if .Values.configmapReload.port }}
+          ports:
+            - containerPort: {{ .Values.configmapReload.port }}
+          {{- end }}
           volumeMounts:
             - name: config
               mountPath: /etc/alertmanager
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -172,6 +172,10 @@ configmapReload:
     tag: v0.5.0
     pullPolicy: IfNotPresent
 
+  ## container port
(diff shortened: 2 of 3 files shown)


charts/alertmanager/values.yaml
```diff
@@ -172,6 +172,10 @@ configmapReload:
     tag: v0.5.0
     pullPolicy: IfNotPresent
 
+  ## container port
+  ##
+  # port: 9533
```
naseemkullah, line 177:
```suggestion
  # containerPort: 9533
```

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -62,6 +62,10 @@ spec:
             - --webhook-url=http://127.0.0.1:{{ .Values.service.port }}/-/reload
           resources:
             {{- toYaml .Values.configmapReload.resources | nindent 12 }}
+          {{- if .Values.configmapReload.port }}
+          ports:
+            - containerPort: {{ .Values.configmapReload.port }}
```
naseemkullah, line 67:
```suggestion
          {{- if .Values.configmapReload.containerPort }}
          ports:
            - containerPort: {{ .Values.configmapReload.containerPort }}
```

naseemkullah (dismissed):
Thanks!

nit: port => containerPort removes need for comment in values file

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -62,6 +62,10 @@ spec:
             - --webhook-url=http://127.0.0.1:{{ .Values.service.port }}/-/reload
           resources:
             {{- toYaml .Values.configmapReload.resources | nindent 12 }}
+          {{- if .Values.configmapReload.containerPort }}
+          ports:
+            - containerPort: {{ .Values.configmapReload.containerPort }}
```
naseemkullah, line 67: im not totally familiar with the reloader, is there any relation to the webhook-url port?

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -62,6 +62,10 @@ spec:
...
+          {{- end }}
```
naseemkullah, line 68:
```suggestion
          ports:
            - containerPort: 9533
```

Would this be bad?

Sadzeih: I didn't want my change to affect every user of the chart, that's why I kept it optional.

naseemkullah: True, I appreciate that, but now I am thinking what if we did, what would be the worst case scenario?

Sadzeih: People might already use the `9533` port number?

naseemkullah: Unlikely enough that I say let's just hardcode da thing... @monotek thoughts?

monotek: Imho it does not hurt to have it configurable.

monotek (approved):

charts/alertmanager/templates/statefulset.yaml
naseemkullah: very well

PR #1922, opened 2022-03-28
[prometheus-adapter] support facultative use of sha256 hash in container image reference

sbeginCoveo:
For fast feedback, please @-mention maintainers that are listed in the Chart.yaml file.
@MattiasGees @steven-sheehy @hectorj2f 

We have strict requirements to use sha256 hashes on all our container image used in production, thus this PR allows users of this helm chart to configure that from the values.yaml file. The sha256 is facultative, back-wards compatible and defaults to none so there are no breaking changes.

Have a wonderful day

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/deployment.yaml
+++ b/charts/prometheus-adapter/templates/deployment.yaml
@@ -42,7 +42,11 @@ spec:
       {{- end}}
       containers:
       - name: {{ .Chart.Name }}
+        {{- if .Values.image.sha256 }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}@sha256:{{ .Values.image.sha256 }}"
+        {{- else }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- end }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -42,7 +42,11 @@ spec:
       {{- end}}
       containers:
       - name: {{ .Chart.Name }}
+        {{- if .Values.image.sha256 }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}@sha256:{{ .Values.image.sha256 }}"
```
steven-sheehy, line 46: Why can't you just populate the sha256 on the existing `image.tag`? The sha256 is specific to any version so you would have to adjust it for every chart release anyway. You don't gain anything by making it separate. In general I'm against increased complexity if there's a valid way to accomplish the same thing.

```yaml
image:
  tag: v0.9.1@sha256:abcd...
```

sbeginCoveo: You're right. I'm ok with that strategy :ok_hand: 
Thanks for the quick response

Issue #1781, opened 2022-02-08
[all] Supported Kubernetes versions policy


dotdc:
### Is your feature request related to a problem ?

Hi everyone,
After the issue https://github.com/prometheus-community/helm-charts/issues/1774, I wanted to know if we had a policy or a list of supported version of Kubernetes but couldn't find it.

Currently, each chart has a Kubernetes version in the "Pre-requisites" section of their README.md, but I think most of them are not up-to-date (a few claim to support 1.8+).

As some charts of this repository works together, I think it would make sense to have a global policy to let users know what to expect from the charts included in this repository.

### Describe the solution you'd like.

Create a global policy that informs users on what versions are likely to be supported by the charts of this repository.

Kubernetes supports the last 3 minor version ([source](https://kubernetes.io/releases/version-skew-policy/) + 1 year of patch support on 1.19+.
I think it would make sense to match this policy, as most users and components of the Kubernetes ecosystem will already try to match it.

We could also consider do to a "best effort" support for older versions on some charts, if this doesn't impact the up-to-date users.

What to you think about this?

@scottrigby @monotek @torstenwalter @mrueg 

### Describe alternatives you've considered.

Otherwise, we could keep the way it works right now and just checks and update the README.md in each chart.

XC-:
This is something that I have been looking for as well. In our use case, we have clusters with different versions and need to figure out which version of prometheus-community/prometheus is compatible with which version of kubernetes. Somehow, I once managed to dig this for one of our clusters, but sadly forgot to bookmark the page I used as a starting point and now I cannot find this information from anywhere. It would be really useful to have a compatibility matrix (such as kubernetes-operator has) for each chart.

dotdc:
not stale, will make a compatibility matrix for kube-state-metrics to start with

dotdc:
Closing this issue related to https://github.com/prometheus-community/helm-charts/pull/1930

PR #1930, opened 2022-03-30
[kube-state-metrics] Compatibility matrix

dotdc:
Signed-off-by: David <davidcalvertfr@gmail.com>

This is will allow chart users to have a up-to-date compatibility matrix between the 5 latest chart releases and the 5 lastest Kubernetes releases. It is inspired by [kube-state-metrics compatibility matrix](https://github.com/kubernetes/kube-state-metrics/blob/master/README.md#kubernetes-version).

  - Issue linked: https://github.com/prometheus-community/helm-charts/issues/1781

I have made a wrapper around [kubepug](https://github.com/rikatz/kubepug) to test the compatibility of the charts.
If you like this format, I can work on a GitHub Action CI to generate this automatically for all the charts of this repository.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 34.10.0
+version: 34.10.1
 appVersion: 0.55.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -6,12 +6,31 @@ See the [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus
 
 _Note: This chart was formerly named `prometheus-operator` chart, now renamed to more clearly reflect that it installs the `kube-prometheus` project stack, within which Prometheus Operator is only one component._
 
-## Prerequisites
+## Compatibility matrix
 
-- Kubernetes 1.16+
-- Helm 3+
+### Helm Chart versions
 
-## Get Helm Repository Info
+This matrix shows the compatibility between the 5 latest chart releases with the 5 latest [Kubernetes releases](https://github.com/kubernetes/kubernetes/releases).
+
+The compatibility is based on the Kubernetes API resources addition, deprecation and deletion.
+
+<!-- START COMPATIBILITY MATRIX -->
+| **Chart Version** | **Kubernetes 1.19** | **Kubernetes 1.20** | **Kubernetes 1.21** | **Kubernetes 1.22** | **Kubernetes 1.23** |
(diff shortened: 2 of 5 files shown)


monotek (commented):
Who is supposed to update/maintain that table and test new versions of kube-state-metrics chart against all new and old k8s versions? Imho this only makes sense if our ci would test on multiple k8s versions.

Whats the currrent benefit if everything is supported anyway?

dotdc:
Sorry, maybe I didn't explain enough.
Of course the idea is to add this to our CI and make theses tests automatically after each release.

What I want to know is:
- What do you think about the format?
- Do you see anything missing?
- Are you ok with the 5 versions?

The benefit is for the users, they will have a compatibility matrix to know which version(s) will work on their k8s cluster.
It could also help us track deprecated or deleted API resources in case we missed some (I think it's the case for some other, less used charts).

I can maintain this for kube-state-metrics in the meantime (it's already half automated).

mrueg:
The Chart.yaml supports `kubeVersion` https://helm.sh/docs/topics/charts/#the-kubeversion-field
We should use this as well.

dotdc:
Thank you @mrueg, I wasn't aware of this feature!

It's a good idea, but if we want to use `kubeVersion` we'll need to test the compatibility with a wider range of Kubernetes versions to generate it.
This is because some charts will probably work on older versions of Kubernetes (older than N-5).

This is fine, but it will probably increase the CI run time (will make further tests to see how much longer it can be).
I will share the numbers to see if it's worth it, CI should remain quite fast in my opinion.

dotdc:
Hi 👋,

Just pushed an update with:
- the script that will create the compatibility matrix of the charts
- the required README.md blocks in `kube-state-metrics` & `kube-prometheus-stack `

This script will:
- get the latest 5 releases of a chart (vX.X.0)
- tests the compatibility of each of them using [kubepug](https://github.com/rikatz/kubepug)
- generate a compatibility matrix
- insert the matrix between a pre-defined block in the chart's README.md

I was originally thinking calling this script in a GitHub workflow that will do that on a vX.X.0 tag.
But then, I realized that I cannot commit a change in a chart's README.md without bumping the chart version.

Would like to know if you have an idea of the best place/moment to do this?

Note: 
- I've locked the script to `kube-prometheus-stack` (forced var) until I know how it will be called
- I didn't implement `kubeVersion` in this first version

monotek:
If compatibility should be tested it should be in a github action, running on pr event like all the others.
Such things can be done with matrix builds but i'm not sure if it's wort to have it.
This would need a lot more resources. So even if it's free one could have the environemtn in mind too.

dotdc:
Hi @monotek,

That's true, and it's useless to run this at each release. 
Like I said, it should run only on vX.X.0 releases and only on the chart being updated. 
I will try to figure out the GitHub workflow part, in any case, the pipeline will need to do a commit to update the README.md's.

I personally think that this would be a really nice information to have for our users, [at least one of them agree](https://github.com/prometheus-community/helm-charts/issues/1781#issuecomment-1040099628) 😅

Would be awesome if more users and the others maintainers tell what they think about this.

Can we @ call them?

PS: We can also enable the matrix only for a pre-defined list of charts depending on the maintainers choices.

Let me know

dotdc:
Closing this for the lack of interest.

PR #1944, opened 2022-03-31
[kube-prometheus-stack] Add sample value to grafana section of values file

nicosingh:
Required in some k8s versions, as documented for other components 😊

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 34.6.0
+version: 34.7.0
 appVersion: 0.55.0
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -648,6 +648,9 @@ alertmanager:
 ##
 grafana:
   enabled: true
+  # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
```
monotek, line 651: i don´t think that info is needed as the ingress is disabled by default.
the comment is also missleading, as it should be under .Values.grafana.ingress.ingressClassName

monotek (changes requested):

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 34.6.0
+version: 34.7.0
```
monotek, line 26: this should als obe only a bugfix version bump.

PR #1975, opened 2022-04-12
[prometheus-adapter] remove helm 2 commands from readme / cleanup values.yaml

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

* remove helm 2 commands from readme 
* cleanup values.yaml (indet of commented out stuff / added remove some blank lines)

  - fixes #

To make it easier to use this chart as a metrics-server replacement together with kube-promehteus-stack chart i've thought about to change `prometheus.url` to "http://kube-prometheus-stack-prometheus.monitoring.svc" and enable `rules.resource`by default. But this would need a new major version so i'm not sure if its worth to add this change. Maybe there are also users who want to use metrics-server and proemtheus-adapter together, so might be unwanted anyway.

#### Checklist

- [ ] Chart Version bumped


steven-sheehy (dismissed):

steven-sheehy (approved):

steven-sheehy:
We definitely shouldn't tie this chart and kube-prometheus-stack together, so I'm against changing the default URL. Welcome to add a section to the README that clarifies how to configure it with kube-prometheus-stack.

I use prometheus-adapter on GKE and GKE already comes with a metrics-server that can't be disabled. So definitely shouldn't make any change that would break that.

Issue #1773, opened 2022-02-07
[kube-prometheus-stack] Info alert inhibition rule evaulation failures


timbuchwaldt:
### Describe the bug a clear and concise description of what the bug is.

Hey! We love the newest release with the rule inhibition feature. We sadly see rule evaluation failures:

```
ts=2022-02-07T14:42:17.636Z caller=manager.go:609 level=warn component="rule manager" group=general.rules msg="Evaluating rule failed" rule="alert: InfoInhibitor\nexpr: ALERTS{severity=\"info\"} == 1 unless on(namespace) ALERTS{alertname!=\"InfoInhibitor\",alertstate=\"firing\",severity=~\"warning|critical\"}\n  == 1\nlabels:\n  severity: none\nannotations:\n  description: |\n    This is an alert that is used to inhibit info alerts.\n    By themselves, the info-level alerts are sometimes very noisy, but they are relevant when combined with\n    other alerts.\n    This alert fires whenever there's a severity=\"info\" alert, and stops firing when another alert with a\n    severity of 'warning' or 'critical' starts firing on the same namespace.\n    This alert should be routed to a null receiver and configured to inhibit alerts with severity=\"info\".\n  runbook_url: https://runbooks.prometheus-operator.dev/runbooks/general/infoinhibitor\n  summary: Info-level alert inhibition.\n" err="vector contains metrics with the same labelset after applying alert labels"
```

### What's your helm version?

3.8.0

### What's your kubectl version?

1.23

### Which chart?

kube-prometheus-stack

### What's the chart version?

v32

### Enter the command that you execute and failing/misfunctioning.

 none

onedr0p:
I am receiving something simulair everyone once in awhile since upgrading to `32.0.0`

![image](https://user-images.githubusercontent.com/213795/152842053-d45fbd34-f0c3-44af-9fb3-2510b3fd9034.png)

cortex3:
hi, im running into the same issue. What does your config look like and have you resolved it already?

cortex3:
So I think the following works for me as a workaround. But I'm not sure why I need this extra configuration on a fresh install.
```      routes:
      - match:
          alertname: 'InfoInhibitor'
        receiver: 'null'
```

irizzant:
[...]

PR #2000, opened 2022-04-20
[kube-prometheus-stack] Add null receiver to InfoInhibitor rule in default AlertManager configuration

jnoordsij:
Add `null` as receiver for the InfoInhibitor rule, and also replace the deprecated `match` syntax with a `matchers` directive. Based on the discussion in #1773.

This prevents the InfoInhibitor rule from causing noise on other receivers, as recommended in the rule description.

#### Checklist

- [x] Chart Version bumped


zanhsieh (dismissed):

zanhsieh (dismissed):

jemag:
would it make more sense to define the inhibit rule as well? Since I believe this is the intent of this alert.

Also in general wondering if inhibit_rules should be defined by default, since upstream kube-prometheus project is doing it : https://github.com/prometheus-operator/kube-prometheus/blob/0ac0d7f9bf5ab1de3e596ff3bdd4b36c52e27267/jsonnet/kube-prometheus/components/alertmanager.libsonnet#L30-L42

zanhsieh (approved):

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 35.2.0
+version: 35.2.1
```
monotek, line 26:
```suggestion
version: 35.2.1
```

zanhsieh (approved):

monotek (approved):

gkarthiks:
@jnoordsij  @zanhsieh Thank you for the contribution. LGTM. 

A small request, next time please squash your commits into one. So that it would be easy for review.

gkarthiks (approved):

jemag:
@gkarthiks any particular reason to merge this without the inhibit rule? 

My understanding is that the sole purpose of this alert is to inhibit other `Info` alerts, as is mentioned in the alert description:
```
...
          This alert should be routed to a null receiver and configured to inhibit alerts with severity="info".
``` 

As far as I am aware, now it is only routed to null but not actually inhibiting anything. It will also not notify users of the chart to do the proper modifications to their own inhibit rules since, by default, it will be routed to the null receiver, effectively rendering it useless to unaware users.

jnoordsij:
> @gkarthiks any particular reason to merge this without the inhibit rule?
> 
> My understanding is that the sole purpose of this alert is to inhibit other `Info` alerts, as is mentioned in the alert description:
> 
> ```
> ...
>           This alert should be routed to a null receiver and configured to inhibit alerts with severity="info".
> ```
> 
> As far as I am aware, now it is only routed to null but not actually inhibiting anything. It will also not notify users of the chart to do the proper modifications to their own inhibit rules since, by default, it will be routed to the null receiver, effectively rendering it useless to unaware users.

I think this is exactly what the current rule definition (https://github.com/prometheus-community/helm-charts/blob/e2af4613e9c79f3f663c697c2e33b711a676b142/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/general.rules.yaml#L82) already does, though I'm not sure on the particulars. The `InfoInhibitor` alert has been firing quite a few times in my setup, without additional config. So it seems to be already in place...

jnoordsij:
@jemag ahhh I see the difference now, I was a little confused about the difference between the alert firing and it actually inhibiting and thus muting the `info` alerts. I created #2114 to implement the actually inhibit rules in the same way kube prometheus does it, would love your feedback on that!

PR #2017, opened 2022-04-29
[prometheus-elasticsearch-exporter] Allow jobLabel configuration

jleloup:

Current configuration of `jobLabel` in Elasticsearch Exporter ServiceMonitor does not allow users to define this value themselves. Also its current values has no effect as it tries to use the release name as a value whereas this parameter is supposed to match a Kubernetes Endpoints label to let Prometheus Operator use its value for Prometheus `job` label.

This PR uses the same strategy as available in prometheus-node-exporter chart:

* allow users to provide `jobLabel` in helm values
* default value refers to Endpoints label `app.kubernetes.io/name`: a [well-known Kubernetes label](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-name)
* since this label is not set to the service resource as of now: default behavior for elasticsearch-exporter would remain the same (Prometheus `job` label valued with the Endpoint name)

No opened issues for this.

I choosed to apply the same strategy as in prometheus-node-exporter with the same implicit behavior regarding the absence of the `app.kubernetes.io/name` label on the Service/Endpoint resource.

As a consequence this does not change the overall default behavior of the chart for users out of the box: I suppose there is no need for a major/breaking change version bump.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-elasticsearch-exporter/templates/servicemonitor.yaml
@@ -33,7 +33,7 @@ spec:
     metricRelabelings:
     {{ toYaml .Values.serviceMonitor.metricRelabelings | nindent 4 }}
     {{- end }}
-  jobLabel: "{{ .Release.Name }}"
+  jobLabel: {{ default "app.kubernetes.io/name" .Values.serviceMonitor.jobLabel }}
   selector:
     matchLabels:
       app: {{ template "elasticsearch-exporter.name" . }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-elasticsearch-exporter/templates/servicemonitor.yaml
```diff
@@ -33,7 +33,7 @@ spec:
     metricRelabelings:
     {{ toYaml .Values.serviceMonitor.metricRelabelings | nindent 4 }}
     {{- end }}
-  jobLabel: "{{ .Release.Name }}"
+  jobLabel: {{ default "app.kubernetes.io/name" .Values.serviceMonitor.jobLabel }}
```
monotek, line 36:
```suggestion
  jobLabel: {{ default .Release.Name .Values.serviceMonitor.jobLabel }}
```

monotek (changes requested):

monotek (approved):

PR #2024, opened 2022-05-03
make targetDown rule namespace restricted to reduce noise from not prod namespaces if needed

vladislavPV:
Signed-off-by: Vlad Paciu <vlad.paciu@adgear.com>

This PR is making targetDown rule namespace restricted if needed.
This might help to reduce alert noise.


vladislavPV:
hello :) @bismarck @Xtigyro @gkarthiks @scottrigby @desaintmartin 
could you review this PR plz
Thank you!

monotek (changes requested):
You can't change the rule like that, as this file is generated by a script: 
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/hack/sync_prometheus_rules.py

Next time it's run your changes are lost. So you have to change the script instead.

vladislavPV:
@monotek Thank you! Will do!

PR #2033, opened 2022-05-05
[prometheus-blackbox-exporter] Add custom relabeling and quote values

epitstickpan:
Adds `additionalRelabeling` to ServiceMonitor so that custom relabeling
can be done. I wanted this because I wanted to be able to add the module
as a label but I think there are other good reasons to have it.

While I was here I added `quote` for `additionalMetricsRelabels` so that
values wouldn't be converted.

Please make note of the fact I did a concat instead of replacing default values. I think this is more productive but it works differently than `additionalMetricsRelabels`. Let me know if you want me to change one or the other so they work the same. Wanted to avoid a major bump if I could.

@desaintmartin 

Signed-off-by: Evan Pitstick <epitstick@paloaltonetworks.com>

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 5.7.0
+version: 5.7.1
 appVersion: 0.20.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 5.7.1
+version: 5.8.0
 appVersion: 0.20.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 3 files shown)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 5.7.0
+version: 5.7.1
```

monotek (changes requested):

charts/prometheus-blackbox-exporter/Chart.yaml
epitstickpan: Thanks, got it.

monotek (approved):

Gentoli:
This causes a panic from helm when `target.additionalRelabeling` is not present:
```
Error: template: prometheus-blackbox-exporter/templates/servicemonitor.yaml:43:7: executing "prometheus-blackbox-exporter/templates/servicemonitor.yaml" at <concat .additionalRelabeling $.Values.serviceMonitor.defaults.additionalRelabeling>: error calling concat: runtime error: invalid memory address or nil pointer dereference
```

This is probably a helm issues, but right now it's broken when `target.additionalRelabeling` to not set.

helm version:
`version.BuildInfo{Version:"v3.8.2", GitCommit:"6e3701edea09e5d55a8ca2aae03a68917630e91b", GitTreeState:"clean", GoVersion:"go1.18.1"}`

epitstickpan:
> This causes a panic from helm when `target.additionalRelabeling` is not present:
> 
> ```
> Error: template: prometheus-blackbox-exporter/templates/servicemonitor.yaml:43:7: executing "prometheus-blackbox-exporter/templates/servicemonitor.yaml" at <concat .additionalRelabeling $.Values.serviceMonitor.defaults.additionalRelabeling>: error calling concat: runtime error: invalid memory address or nil pointer dereference
> ```
> 
> This is probably a helm issues, but right now it's broken when `target.additionalRelabeling` to not set.
> 
> helm version: `version.BuildInfo{Version:"v3.8.2", GitCommit:"6e3701edea09e5d55a8ca2aae03a68917630e91b", GitTreeState:"clean", GoVersion:"go1.18.1"}`

I have a fix coming in for it. Thanks for the heads up.

PR #2050, opened 2022-05-12
[kube-prometheus-stack] Add grafana persistence values

QuentinBisson:

This PR adds the persistence values for the grafana dependency as using persistence is a regular use case and it makes it easier to onboard with this helm chat

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
I'm not a fan of adding disabled values to the values file. It's already bug enough.

QuentinBisson:
@monotek how about a comment explaning that there are a lot of other things to configure described in the grafana helm chart like persistence and so forth ?

PR #2059, opened 2022-05-14
[prometheus-adapter] remove default rules which are causing resource spikes

### LINKED ISSUE #2052: Adapter resource usage spikes as the cluster grows
The default rules are causing a spike in memory and CPU utilisation as the workloads on the cluster increase.

https://github.com/prometheus-community/helm-charts/blob/4009e418d532f474781d79482d6c173260d4df1d/charts/prometheus-adapter/templates/configmap.yaml#L18-L83

```shell
$ kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1 | jq -r '.resources | length'
9197
```
Screenshot 2022-05-13 at 00 24 17

By removing those rules, and keeping the `resourceRules` with some of our extra custom metrics, the consumption drastically drops, and the functionalities (top pods/nodes, hpa) are working fine:
```shell
$ kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1 | jq -r '.resources | length'
4
```
Screenshot 2022-05-13 at 00 22 48

Can anyone please help me with what those metrics are we using for? If those aren't needed I can create a PR to remove those.

Also, the configMap from [`kube-prometheus`](https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheusAdapter-configMap.yaml) doesn't have those default rules as well.

junaid-ali:

The default rules as described in this issue #2052 causing spikes in CPU and particularly for memory as the cluster grows. The metrics fetched due to those rules doesn't seem to be much related/needed for prometheus adapter. We essentially would only need resource rules, alongside any custom metrics. This PR removes those default rules from adapter chart.

  - fixes #2052 

`kubectl` top pods/nodes alongside any HPA that depends on custom metrics are working as expected.

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/configmap.yaml
+++ b/charts/prometheus-adapter/templates/configmap.yaml
@@ -12,79 +12,9 @@ metadata:
     {{- include "k8s-prometheus-adapter.labels" . | indent 4 }}
 data:
   config.yaml: |
-{{- if or .Values.rules.default .Values.rules.custom }}
-    rules:
-{{- if .Values.rules.default }}
-    - seriesQuery: '{__name__=~"^container_.*",container!="POD",namespace!="",pod!=""}'
-      seriesFilters: []
-      resources:
-        overrides:
-          namespace:
-            resource: namespace
-          pod:
-            resource: pod
-      name:
-        matches: ^container_(.*)_seconds_total$
-        as: ""
-      metricsQuery: sum(rate(<<.Series>>{<<.LabelMatchers>>,container!="POD"}[5m]))
-        by (<<.GroupBy>>)
-    - seriesQuery: '{__name__=~"^container_.*",container!="POD",namespace!="",pod!=""}'
-      seriesFilters:
-      - isNot: ^container_.*_seconds_total$
-      resources:
-        overrides:
-          namespace:
-            resource: namespace
-          pod:
-            resource: pod
-      name:
-        matches: ^container_(.*)_total$
-        as: ""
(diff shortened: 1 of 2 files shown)


monotek (changes requested):
Why not setting rules.default to false and using custom rules if the default does not fit your needs?

Like this it introduces a breaking change for no reason.

junaid-ali:
@monotek thanks for the review. I should have created the change as draft since I wanted to get feedback for those default rules first.

Currently, I'm using `rules.default: false` but IMO we should have those disabled by default (or removed if not needed) since at some point those rules will cause issues for everyone. Could you please suggest a way to do that without introducing a breaking change?

monotek:
There is no way to do it without a new major version.
As it's already configurable i don't see the need for it.

junaid-ali:
@monotek what about if we change the default value to `false`. I assume that wouldn't bring in any breaking change.

junaid-ali:
My confusion is whether those metrics are really required for prometheus-adapter, if not: at least disable those by default.

monotek:
You can't change an existing value without makin it a breaking change.
If somebody has installed it before and relies on the default it would change, which is bad.

steven-sheehy:
I agree with @monotek. You can just set it to false for your deployments.

junaid-ali:
I do agree that it's a major change (don't think it would be a breaking change since that means a failure/issues to existing deployments).

Do you think we should have it documented at least in the values.yaml or README.md?

monotek:
why change it at all if it's already configurable?

PR #2066, opened 2022-05-17
Adjust gRPC failed requests

svparamzin:
https://github.com/etcd-io/etcd/pull/13127

Let me try to summarize the issue:

The problem of the expression comes from the grpc_code!="OK" part: it treats other non-error codes (canceled, invalidArgument, notFound, and etc) as errors so the expression reports a high rate of errors thus triggers too many alerts (see GRPC standard here).
The above problem in the expression has been fixed in the upstream etcd repo: etcd-io/etcd#13127

Signed-off-by: svparamzin <60502949+svparamzin@users.noreply.github.com>

  - fixes #

#### Checklist

- [ ] Chart Version bumped


monotek (changes requested):
This would be needed to be fixed upstream as the dashboards are downloaded via the scripts in the hack directory.

PR #2067, opened 2022-05-17
[kube-prometheus-stack] Fix not working cluster selection in ETCD dashboard

TomaszSyc:
What this PR does / why we need it:

  - #2069


monotek (changes requested):
This would be needed to be fixed upstream as the dashboards are downloaded via the scripts in the hack directory.

PR #2153, opened 2022-06-13
[kube-prometheus-stack] fixed typos for additional secret configs

### LINKED ISSUE #2152: [kube-prometheus-stack] typo in additional scrape configs

The prometheus.yaml in templates has a typo in the name and that caused some headaches before I could figure out that it not being picked up was because of typo in the name. 

### What's your helm version?

version.BuildInfo{Version:"v3.8.1", GitCommit:"5cb9af4b1b271d11d7a97a71df3ac337dd94ad37", GitTreeState:"clean", GoVersion:"go1.17.5"}

Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.12", GitCommit:"696a9fdd2a58340e61e0d815c5769d266fca0802", GitTreeState:"clean", BuildDate:"2022-04-13T19:07:00Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}

### Which chart?

kube-prometheus-stack

### What's the chart version?

36.0.0

### What happened?

Nothing was failing, beacuse I assumed that the manually created secret was correctly named in templates but was not, it was not getting picked up.

fix the typo 😅   -- I will raise a PR

### How to reproduce it?

create a secret named {{ template "kube-prometheus-stack.fullname" . }}-prometheus-scrape-config and it does not get picked up by the chart

none

helm install my-release prometheus-communitykube-prometheus-stack --version 36.0.0 --values values.yaml

the command is not actually correct, we install it from via terraform 😅

comdotlinux:
Signed-off-by: Guruprasad Kulkarni <guru@linux.com>

  - fixes #2152 

* Not sure if a major version bump was required but in case someone else also created a secret externally then they would easily notice it in the Upgrading section as opposed to a minor version bump. 
* Also this makes the chart incompatible with previous version
* That was my rationale for major version bump.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.0.0
+version: 37.0.0
 appVersion: 0.57.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -80,6 +80,18 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
 
 A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.
 
+### From 36.x to 37.x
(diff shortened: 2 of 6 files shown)


monotek (commented):
Is this realy needed?

As far as is see the prometheus spec does reference the secret with the typo? So the only problem you have is, that an already exisitng secret has to have the "correct" name (with the typo)?

It looks cleaner though but i'm not sure if its worth the hastle for upgrading users?

Maybe we could change it in a way which keeps the current default but makes the secretname configurable? Like this we would also only need to raise the minor version of the chart.

comdotlinux:
True this is not required, I dont know why but I missed the key `additionalScrapeConfigsSecret` where we can specify the name of external config that you want to use, so this solves my issue without any changes to the chart 🤦🏻

PR #2167, opened 2022-06-17
[kube-state-metrics] render extraArgs as YAML to allow multi-line string arguments

mhulscher:
The newly introduced `--custom-resource-state-config` argument
(https://github.com/kubernetes/kube-state-metrics/pull/1710) accepts a multi-line string as value.
The deployment.yaml template currently loops over each .Values.extraArgs, which causes indentation
issues when passing a multi-line string. This PR fixes that by rendering .Values.extraArgs as YAML
with proper indentation.

Signed-off-by: Mitch Hulscher <mitch.hulscher@lib.io>

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.9.1
+version: 4.9.2
 appVersion: 2.5.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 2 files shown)


monotek (changes requested):
This would be a breaking change for upgrading users, which already use this variable.

mhulscher:
@monotek do you have an example of what could break? At the moment, passing multi-line arguments is not possible. Existing single-line arguments will continue to work.

monotek (commented):
Ok, seems something like `helm template . --set extraArgs[0]="--test1",extraArgs[1]="--test2"` still works.
Do you have an example for your multiline config?

monotek (approved):

Issue #1418, opened 2021-10-08
[kube-prometheus-stack] Generate GrafanaDashboard manifest files for Grafana-Operator


ErikLundJensen:
### Is your feature request related to a problem ?

Using kube-prometheus-stack together with Grafana Operator works fine except getting the default dashboards from kube-prometheus-stack into Grafana instances managed by Grafana Operator.

The dashboards are defined in ConfigMaps and not in GrafanaDashboard manifests, .e.g.
```
apiVersion: integreatly.org/v1alpha1
kind: GrafanaDashboard
```

### Describe the solution you'd like.

I suggest that we either extend the existing script [sync_grafana_dashboards.py](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py) to generate GrafanDashboard manifests or that we create a new script that just generates the GrafanaDashboard manifest files. The two scripts would be very alike and some refactoring could be applied after initial test.

Thereby the GrafanaDashboard definition could be added to the Helm Chart Grafana and only be activated when setting grafana.operator: enabled

Thereby we will have a setup where this chart is backwards compatible but also support the Grafana Operator.

### Describe alternatives you've considered.

Alternative jsonnet or python script could do post-processing of the dashboard ConfigMaps and create GrafanaDashboards from those ConfigMaps by changing the object kind.
In principal it would be replacing `kind` and `api` and change `data: *.json` to `spec: json'.

PR #2172, opened 2022-06-18
[kube-prometheus-stack] Adding support for Grafana Operator

jtyr:

This PR is adding the possibility to use Grafana Operator instead of the simple Grafana instance. The main driver is to be able to use the `GrafanaDashboard` and `GrafanaDatasource` custom resources instead of `ConfigMap`.

- related to issue #1418

The looks so big because all the auto-generated dashboards from `templates/dashboards-1.14` were moved to the top level of the chart and are now shared across the `templates/grafana` and `templates/grafana-operator` templates.

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
IMHO this not needs to be an option in the chart. You can just disable grafana and add the operator seperatly.

jtyr:
Thank you for your comment, @monotek. You could say the same about the `grafana` chart as well. The advantage of having the Grafana Operator as a part this chart is that people get automatically configured datasource as well as all the dashboards in their Grafana instance and they can keep extending it through standard `GrafanaDataSource` and `GrafanaDashboard` custom resources that goes through CRD validation during the installation. That's something that's hard to achieve through configuration pushed via `ConfigMap` without any operator or validating wekbook.

jtyr:
@andrewgkew @bismarck @desaintmartin @gianrubio @gkarthiks @scottrigby @Xtigyro Please could I ask you for your review?

dwalker-sabiogroup:
Adding this optional functionality to the be able to use the custom resources provided by Grafana Operator when using this chart would be a huge advantage.  

The current option of disabling Grafana in this chart and adding to the Grafana Operator chart means having to manage the out of the box dashboards.  These dashboards are so frequently deployed that end users expect them to be available and doing so separately and keeping them up to date is a problem

jtyr:
@andrewgkew @bismarck @desaintmartin @gianrubio @gkarthiks @scottrigby @Xtigyro Please could review this PR?

vladimir-babichev:
Having Grafana Operator bundled together with a Prometheus Operator will help to address multi-tenanted environments, where at the moment, we need to deploy a separate Grafana Helm chart together with Prometheus CR in order to create a dedicated monitoring stack for every developer team. 

Apart from the above, it also unifies the experience of managing all `kube-prom-stack` resources:
* Prometheus instances
* AlertManager instances
* Grafana instances
* Prometheus resources: rules, monitors, etc
* Grafana resources: dashboards, datasources, notification channels

In other words, you're just one CR away from your entire monitoring stack, which hugely simplifies deployment and is one of the main goals of this project.

The other huge benefit I see is the ability to define dashboards via CRD. Apart from automated validation of resources and ease of error discovery through a status field, the user will get the ability to author dashboards with jsonnet, add plugin requirements, map datasources, and manage folders.

jtyr:
@monotek it looks like you are the only person who cares about this Helm chart as the rest of the assigned reviewers don't bother to provide any feedback. Please could you reconsider this PR in regards of the feedback from the community above?

vladimir-babichev:
@monotek @andrewgkew @bismarck @desaintmartin @gianrubio @gkarthiks @scottrigby @Xtigyro Can you guys review and/or comment on this PR? I think this PR will make life easier for many of us, especially for those running heterogeneous (k8s and OpenShift) multi-tenanted environments.

dotdc:
Seems like a valid use-case, maybe we should reconsider this.

Does any grafana-operator users tested the PR chart?

@jtyr @monotek @dwalker-sabiogroup @vladimir-babichev

dotdc:
If you don't like the idea, maybe we should close this PR then? Unless there's hope for them? If yes how? 

If not, I'll guess they could make a chart with kube-prometheus-stack, grafana-operator and whatever they need on their side.

Both options LGTM

monotek:
I'm not a maintainer of the chart, so i would not be in the way if the maintainers think it's a good idea.

dotdc:
My bad... I thought you were because it seems you're the most active on this chart/repository.

jkroepke:
Adding any chart from https://charts.bitnami.com/bitnami as dependency is always a really bad idea.

Bitnami has some retention policies on the chart repository, e.g. they remove older chart version from the repo.

https://github.com/bitnami/charts/issues/10539

At some point, older kube-prometheus-stack version can not be installed anymore, if the grafana-operator version is not longer available. 

I personally would not pull the grafana-operator helm chart into this stack. 

From my point of view, if `kube-prometheus-stack` exposes GrafanaDataSource and GrafanaDashboard CR, it would be sufficient to support the Grafana Operator. Users can bundle the installation with an local umbrella chart.

vladimir-babichev:
> Adding any chart from https://charts.bitnami.com/bitnami as dependency is always a really bad idea.
> Bitnami has some retention policies on the chart repository, e.g. they remove older chart version from the repo.
> https://github.com/bitnami/charts/issues/10539
> At some point, older kube-prometheus-stack version can not be installed anymore, if the grafana-operator version is not longer available.

I believe it shouldn't be a problem, as helm packages all dependency charts into one single tgz file. It does so, to specifically resolve dependency management problem.

> From my point of view, if kube-prometheus-stack exposes GrafanaDataSource and GrafanaDashboard CR, it would be sufficient to support the Grafana Operator. Users can bundle the installation with an local umbrella chart.

vladimir-babichev:
The other problem, I personally find, is the lack of `prometheus-operator` chart that installs the operator itself with its CRDs. Having one would resolve issues like this, as users could create wrapper charts and bundle dependencies according to their needs.

rhysxevans:
Hi

Has there been any progress on this? 

I would like to help get this through.  However I would like to move away from the bitnami charts as they don't support arm as of yet.

Any thoughts

Thanks

vladimir-babichev:
@rhysxevans I don't believe there is non-bitnami chart for grafana-operator, but there is [an official image published](https://quay.io/repository/grafana-operator/grafana-operator?tab=tags) to quay.io that has arm builds.

DWebb0:
+1. I'm registering my interest here

rhysxevans:
Bitnami has just released arm support for the majority of their images apparently https://blog.bitnami.com/2023/02/bitnami-arm-containers-available-at.html?m=1  and https://hub.docker.com/r/bitnami/grafana-operator/tags

jkroepke:
After bitnami announcement 4 weeks ago to limit there free offerings, the was a very good choice to not depend against bitami.

Issue #2133, opened 2022-06-08
[kube-prometheus-stack] Reduce container metrics by default.


SuperQ:
### Is your feature request related to a problem ?

Many users complain about memory use of Prometheus when using `kube-prometheus-stack`. One major contributor to this is kubelet cAdvisor `container_...` metrics.

There are a number of overlapping, less useful, or always zero metrics returned by cAdvisor.

### Describe the solution you'd like.

In order to improve on this, we should add a default `metricRelabelings` to the `ServiceMonitor`. This would more closely align with the recommended defaults in the [upstream kube-prmetheus manifests](https://github.com/prometheus-operator/kube-prometheus/blob/bfc49f2fe984fbc8de6a2f3bf14c7bc5d2cbce57/manifests/kubernetesControlPlane-serviceMonitorKubelet.yaml#L59-L74).

### Describe alternatives you've considered.

We could document this, but it would be much better to have sane defaults built in.

SuperQ:
First proposal, this is the metricRelabeling I use in production.

```
kubelet:
  serviceMonitor:
    cAdvisorMetricRelabelings:
      # Drop less useful container CPU metrics.
      - sourceLabels: [__name__]
        action: drop
        regex: 'container_cpu_(cfs_throttled_seconds_total|load_average_10s|system_seconds_total|user_seconds_total)'
      # Drop less useful container / always zero filesystem metrics.
      - sourceLabels: [__name__]
        action: drop
        regex: 'container_fs_(io_current|io_time_seconds_total|io_time_weighted_seconds_total|reads_merged_total|sector_reads_total|sector_writes_total|writes_merged_total)'
      # Drop less useful / always zero  container memory metrics.
      - sourceLabels: [__name__]
        action: drop
        regex: 'container_memory_(mapped_file|swap)'
      # Drop less useful container process metrics.
      - sourceLabels: [__name__]
        action: drop
        regex: 'container_(file_descriptors|tasks_state|threads_max)'
      # Drop container spec metrics that overlap with kube-state-metrics.
      - sourceLabels: [__name__]
        action: drop
        regex: 'container_spec.*'
      # Drop cgroup metrics with no pod.
      - sourceLabels: [id, pod]
        action: drop
        regex: '.+;'
      # Drop cgroup metrics with no container.
      - sourceLabels: [id, container]
        action: drop
        regex: '.+;'
```

bryanasdev000:
Not stale.

[...]

PR #2197, opened 2022-06-27
[kube-prometheus-stack] Update metric relabel values

SuperQ:

Add some default metric relabel config values to cAdvisor and apiserver
configuration to reduce default cardinality. This will likely save a
couple hundred MiB of memory off Prometheus in the default
configuration.

Fixes: https://github.com/prometheus-community/helm-charts/issues/2133

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.6.2
+version: 37.0.0
 appVersion: 0.57.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
this changes the defaults and is therefore a breaking change.
not sure if necessarily needed if its already configurable.

SuperQ:
The problem is that the current defaults are very bad for most users. Users with lower resources have a hard time deploying kube-prometheu-stack due to the amount of memory consumed.

The metrics being removed are also quite useless.
* Excessive high cardinality in the API server metrics provide no value, but upstream Kubernetes is not really fixing this.
* Many of the cAdvisor metrics produce no data (all zeros) or are overlapping / low value.

monotek:
Still possible somebody is using / relying on it.
Maybe it could be implemented as option or as comment?

monotek:
If so major version needs to be raised and some notes added to the upgrading section.
I would like to see some other reviews from the cart maintainers too.

monotek (changes requested):

PR #2216, opened 2022-06-30
[kube-prometheus-stack] fix: rename thanos sidecar servicemonitor

aslafy-z:
Signed-off-by: Zadkiel Aharonian <zadkiel_aharonian@carrefour.com>

Renames the sidecar thanos ServiceMonitor from `{fullname}-thanos-discovery` to `{fullname}-thanos-sidecar`.
This would make the official Thanos Sidecar grafana dashboard work out of the box with this implementation. See https://github.com/thanos-io/thanos/blob/v0.25.2/examples/dashboards/sidecar.json#L1462. 

  - fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.2.0
+version: 36.2.1
 appVersion: 0.57.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus/servicemonitorThanosSidecar.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/servicemonitorThanosSidecar.yaml
@@ -2,15 +2,15 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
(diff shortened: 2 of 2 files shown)


charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.2.0
+version: 36.3.0
```
monotek, line 26:
```suggestion
version: 36.2.1
```

monotek (changes requested):

charts/kube-prometheus-stack/Chart.yaml
aslafy-z: You're right. Done!

monotek (approved):

rouke-broersma:
@monotek @aslafy-z maybe I'm missing something here but if the servicemonitor is changed but the thanos sidecar service is not, how is this change supposed to work? Seems to me there is now a disconnect between the thanos sidecar service and the servicemonitor since the service name is still discovery instead of sidecar.

rouke-broersma:
> @rouke-broersma The ServiceMonitor name has no need to match the Service name. If the selector is well defined, it should not cause any issues AFAIK

But the service label is still thanos-discovery, it does not match with the new selector thanos-sidecar in the servicemonitor.

aslafy-z:
@rouke-broersma I just opened https://github.com/prometheus-community/helm-charts/pull/2244 to fix the issue. Sorry for the noise!

PR #2222, opened 2022-07-04
[kube-state-metrics] Add selfMonitor.telemetryNodePort field to values

AlexHill:
Adds a `selfMonitor.telemetryNodePort` field to values.yaml so that a NodePort for the telemetry endpoint can be specified when `service.type` is set to `NodePort`.

Tagging @tariq1890 @mrueg @dotdc as per the PR template comment.

There is a slight incongruence here in that if `selfMonitor.telemetryNodePort` is specified but `service.type` is not set to `NodePort`, the deployment will fail. Options to fix, if desired:
* put the telemetry endpoint fields under `service`
* create separate Services for telemetry and core metrics
* set the Service type to `NodePort` if either `service.nodePort` or `selfMonitor.telemetryNodePort` are specified

(In the existing chart, the deployment will also fail if `services.nodePort` is non-zero while `services.type` is not `NodePort` - but that's a bit less surprising since the dependent fields are all under the `services` top-level key.)

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.10.0
+version: 4.11.0
 appVersion: 2.5.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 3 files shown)


dotdc:
Hi @AlexHill,

I think it's ok to merge as is if you add the limitation in the comment block (see my review)

AlexHill:
Hi @dotdc, thanks for the extremely quick response!

Not sure what's going on but I can't see your review. In any case though I've added a comment now.

dotdc (approved):
LGTM

dotdc:
My bad, the review was still pending. Anyway the comment is basically what I suggested.
Thank you for your contribution!

PR #2238, opened 2022-07-06
[kube-prometheus-stack] Fix/yaml error tag yaml 2002 value

Moep90:
@andrewgkew
@bismarck
@gianrubio

This will fix a pyaml issue which causes the equal sign to throw a `could not determine a constructor for the tag 'tag:yaml.org,2002:value'` error.

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.2.1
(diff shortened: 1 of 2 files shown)


monotek (changes requested):
crds are created automatically by the scripts in the hack dir.
your change would get lost when recreated.
please fix upstream.

zanhsieh:
Source PR here:

https://github.com/prometheus-operator/prometheus-operator/pull/4897

Moep90:
see: https://github.com/yaml/pyyaml/issues/89

dotdc:
Agree with the previous comments, this should be done directly in [prometheus-operator/prometheus-operator](https://github.com/prometheus-operator/prometheus-operator)
Closing this PR.

PR #2244, opened 2022-07-06
[kube-prometheus-stack] fix the thanos discovery servicemonitor selector

aslafy-z:
Signed-off-by: Zadkiel Aharonian <hello@zadkiel.fr>

Fix a bug I introduced with #2216 which prevents the Thanos Sidecar ServiceMonitor from reaching the Thanos Sidecar service.

  - fixes https://github.com/prometheus-community/helm-charts/pull/2216#issuecomment-1176262339

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 36.2.1
+version: 36.2.2
 appVersion: 0.57.0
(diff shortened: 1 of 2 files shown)


monotek (approved):

rouke-broersma:
@aslafy-z this fixes the servicediscovery, thank you!

However the dashboard is still not functional. Unfortunately I am not quite up to speed on how prometheus works exactly but the change did not change the job label which is why the dashboard still does not work:

image

I couldn't find how prometheus determines what should be used as the job label but from naming I assume that it uses the service name as the job name. This to me suggests that your original change was correct, but that the service name must also be changed. This is of course a breaking change. Perhaps it's easier to change the example dashboard in the thanos repo?

@monotek how would you like to deal with this so we can fix it properly?

PR #2245, opened 2022-07-06
[prometheus-statsd-exporter] Statsd exporter configurable liveness

rabunkosar-dd:

This makes the statsd exporter liveness probe configurable and makes the check use / by default. The current default times out after 1 second and if the `/metrics` endpoint has a lot of content 1 second can be a problem. 

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
Please keep original /metrics path so an update does not change the default value.

PR #2264, opened 2022-07-11
[kube-prometheus-stack] Update port for metric of etcd

nobiit:
Fix port for metric etcd

https://github.com/kubernetes/kubernetes/blob/dafa55be106730eaacb740e20b999b84848c6ac5/cmd/kubeadm/app/constants/constants.go#L95-L96

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -22,7 +22,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 39.4.1
+version: 39.4.2
 appVersion: 0.58.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1284,8 +1284,8 @@ kubeEtcd:
   ##
   service:
     enabled: true
-    port: 2379
-    targetPort: 2379
+    port: 2381
+    targetPort: 2381
     # selector:
     #   component: etcd
 


charts/kube-prometheus-stack/values.yaml
```diff
@@ -1284,8 +1284,8 @@ kubeEtcd:
   ##
   service:
     enabled: true
-    port: 2379
-    targetPort: 2379
+    port: 2381
+    targetPort: 2381
```
monotek, line 1288: Maybe we should have some logic here using the right port depnedeing on the k8s version?
The values of the ports could be empty and only used when actualy set.

nobiit: A pretty good idea, maybe I'll give it a try :D

nobiit: Hey @monotek , I checked kubernetes git documentation and history (via blame), it seems that etcd metrics was always running on port 2381.
https://github.com/kubernetes/kubernetes/blob/96a32be18dfbd38ff5cb1df4d68528e2cf226958/cmd/kubeadm/app/constants/constants.go#L96 

I checked kubernetes git documentation and history (via blame), it seems that etcd metrics was always running on port 2381

monotek: Not sure about that: https://etcd.io/docs/v3.1/op-guide/monitoring/

I don't have time to check further. Have only my mobile phone the next 4 weeks...

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -22,7 +22,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 39.4.1
+version: 39.4.2
```
monotek, line 25:
```suggestion
version: 40.0.1
```

charts/kube-prometheus-stack/values.yaml
monotek: just tested it on a kind cluster. it is localhost:2381/metrics

monotek (changes requested):

monotek (approved):

nobiit:
Thanks @monotek

Issue #2279, opened 2022-07-15
[prometheus-kube-stack] New cAdvisorMetricRelabelings are a bit too restrictive by default


dotdc:
### Describe the bug a clear and concise description of what the bug is.

Hi,

I think two of the new rules introduced in 37.0.0 by @SuperQ are a bit too restrictive to be enabled by default.

Rules concerned :

```
      # Drop cgroup metrics with no pod.
      - sourceLabels: [id, pod]
        action: drop
        regex: '.+;'
      # Drop cgroup metrics with no container.
      - sourceLabels: [id, container]
        action: drop
        regex: '.+;'
```

Theses 2 rules drop every metrics without a pod and/or a container.
This will make some basic queries like `container_cpu_usage_seconds_total{id="/"}` to stop working out of the box.
I would at least comment theses two to limit the number of impacted users.

What do you think?

### What's your helm version?

3.9.0

### What's your kubectl version?

1.24.2

### Which chart?

prometheus-kube-stack

### What's the chart version?

37.2.0

### Enter the command that you execute and failing/misfunctioning.

container_cpu_usage_seconds_total{id="/"}

SuperQ:
Adding my comments from the PR:

These are mostly generated by systemd cgroups, which can be pretty noisy for Kubernetes users. This is why I added them as the default.

The goal here is to monitor Kubernetes resources.

Also, I'm not sure what use of monitoring the CPU of the root cgroup is, when `node_cpu_seconds_total` covers that.

dotdc:
Good point, can't remember why I didn't used `node_cpu_seconds_total` to calculate global cluster usage, will check again and update this.

BeckYeh:
Becasue this metrics used in prometheus-adapter for nodes resource query（ kubectl top nodes ）.
So by default I hope don't add this filter.

SuperQ:
Ooof, the prometheus-adapter seems to have a bunch of incorrect queries built in. Those should be fixed.

SuperQ:
https://github.com/kubernetes-sigs/prometheus-adapter/pull/516 should fix up hte prometheus-adapter bug.

SuperQ:
I think we can close it. Other projects can find this issue for advice on how to fix their query use.

PR #2280, opened 2022-07-15
[prometheus-kube-stack] disabled restrictive cAdvisorMetricRelabelings rules by default

dotdc:
Signed-off-by: David <davidcalvertfr@gmail.com>

I think two of the new rules introduced in 37.0.0 by @SuperQ are a bit too restrictive to be enabled by default.

Rules concerned :

      # Drop cgroup metrics with no pod.
      - sourceLabels: [id, pod]
        action: drop
        regex: '.+;'
      # Drop cgroup metrics with no container.
      - sourceLabels: [id, container]
        action: drop
        regex: '.+;'
Theses 2 rules drop every metrics without a pod and/or a container.
This will make some basic queries like container_cpu_usage_seconds_total{id="/"} to stop working out of the box.
I would at least comment theses two to limit the number of impacted users.

  - fixes https://github.com/prometheus-community/helm-charts/issues/2279

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 37.2.0
+version: 37.3.0
 appVersion: 0.57.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


SuperQ:
These are mostly generated by systemd cgroups, which can be pretty noisy for Kubernetes users. This is why I added them as the default.

The goal here is to monitor Kubernetes resources.

SuperQ:
Also, I'm not sure what use of monitoring the root cgroup is, when `node_cpu_seconds_total` covers that.

dotdc:
Closing this, no longer needed because `node_cpu_seconds_total` is a better approach. Thanks!

PR #2297, opened 2022-07-21
[kube-prometheus-stack] Remove a cAdvisor metric relabelling

LambArchie:
Without this change v37 broke certain metrics which are useful, namely network metrics such as `container_network_receive_bytes_total`

While this issue was identified on k3s, I have encountered the issue on both an AKS and GKE cluster (running v1.21 and v1.23 respectively). You can see the problem if you run the following command `kubectl get --raw /api/v1/nodes/NODE_NAME/proxy/metrics/cadvisor | grep 'container_network'` (replacing NODE_NAME with a nodes name in your cluster)

  - fixes https://github.com/k3s-io/k3s/issues/5891

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 37.3.0
+version: 38.0.0
 appVersion: 0.57.0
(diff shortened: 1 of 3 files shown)


monotek (approved):

SuperQ:
This change is not correct. The correct fix would have been to narrow the matching so that it drops metrics only for things that should have a container label. A simple rollback is not enough to do what is intended by the original change.

SuperQ:
It's even mentioned in the upstream issue that this is not the correct fix: https://github.com/k3s-io/k3s/issues/5891#issuecomment-1191385292

PR #2298, opened 2022-07-21
[kube-state-metrics] - Add option to only use the release namespace

Allex1:
This pr allows kube-state-metrics to only watch the release namespace for collecting metrics. This is particularly useful for users running in multi tenant Kubernetes clusters where their access is restricted to 1 namespace.
Similar functionality is already available in Grafana/Prometheus-Operator helm charts. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.13.0
+version: 4.14.0
 appVersion: 2.5.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -74,7 +74,9 @@ spec:
         {{- if .Values.metricDenylist }}
         - --metric-denylist={{ .Values.metricDenylist | join "," }}
         {{- end }}
-        {{- if .Values.namespaces }}
+        {{- if .Values.releaseNamespace }}
+        - --namespaces={{ template "kube-state-metrics.namespace" . }}
+        {{- else if .Values.namespaces }}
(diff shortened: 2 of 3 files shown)


charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.13.0
+version: 4.13.1
```
monotek, line 10: new feature should raise minor version

monotek (changes requested):

charts/kube-state-metrics/Chart.yaml
Allex1: @monotek dome

monotek:
Please sign commits to get DCO check working.

Allex1:
@monotek done

Allex1:
@dotdc @mrueg @tariq1890 @monotek Kind reminder

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -80,6 +80,9 @@ spec:
         {{- if .Values.namespacesDenylist }}
         - --namespaces-denylist={{ tpl (.Values.namespacesDenylist | join ",") $ }}
         {{- end }}
+        {{- if .Values.releaseNamespace }}
```
mrueg, line 83: Can you move this a couple of lines up, so it is below the other item that configures namespaces.

Also: what happens if both are set?

Allex1: > Can you move this a couple of lines up, so it is below the other item that configures namespaces.

Done

> Also: what happens if both are set?

Good point. I've updated the template and comments. Have another look pls

Allex1:
@monotek @mrueg kind reminder.

monotek (approved):

mrueg (approved):

PR #2339, opened 2022-08-03
[prometheus-mysql-exporter] Add option for ipAddressTypes

agarat:
This PR adds the option to add the parameter `-ip-address-types` for the cloudsqlproxy deployment. The default value of this parameter is `PUBLIC,PRIVATE`. With this option you can change it to what you prefer, in our case we would prefer to use `PRIVATE,PUBLIC`.

No issue

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/values.yaml
+++ b/charts/prometheus-mysql-exporter/values.yaml
@@ -144,7 +144,7 @@ cloudsqlproxy:
     tag: "1.30.1-alpine"
     pullPolicy: "IfNotPresent"
   instanceConnectionName: "project:us-central1:dbname"
-  # ipAddressTypes: "PRIVATE,PUBLIC"
+  ipAddressTypes: ""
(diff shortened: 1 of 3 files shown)


blackjid (approved):

charts/prometheus-mysql-exporter/values.yaml
```diff
@@ -144,6 +144,7 @@ cloudsqlproxy:
     tag: "1.30.1-alpine"
     pullPolicy: "IfNotPresent"
   instanceConnectionName: "project:us-central1:dbname"
+  # ipAddressTypes: "PRIVATE,PUBLIC"
   port: "3306"
```
monotek, line 148: Please uncomment and use empty string here.

monotek (changes requested):

monotek (approved):

PR #2348, opened 2022-08-05
[prometheus-blackbox-exporter] Drop all capabilities in case that it is not needed

daniel-maganto:
Signed-off-by: Daniel Maganto <daniel.maganto@allianzdirect.es>

Drop all capabilities in case it is not needed to reduce the surface attack 
N/A

- fixes #

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/deployment.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/deployment.yaml
@@ -61,6 +61,8 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: "{{ .Values.priorityClassName }}"
       {{- end }}
+      securityContext:
+{{ toYaml .Values.podSecurityContext | indent 8 }}  
       {{- if .Values.extraInitContainers }}
       initContainers:
 {{ toYaml .Values.extraInitContainers | indent 8 }}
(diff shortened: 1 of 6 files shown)


charts/prometheus-blackbox-exporter/templates/deployment.yaml
```diff
@@ -86,6 +86,8 @@ spec:
               {{- if .Values.runAsGroup }}
             runAsGroup: {{ .Values.runAsGroup }}
               {{- end }}
+            capabilities:
+              drop: ["ALL"]
```
monotek, line 90: please make this configurable.

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/deployment.yaml
daniel-maganto: Done

charts/prometheus-blackbox-exporter/templates/deployment.yaml
```diff
@@ -86,6 +86,10 @@ spec:
               {{- if .Values.runAsGroup }}
             runAsGroup: {{ .Values.runAsGroup }}
               {{- end }}
+              {{- if .Values.dropAllCapabilities }}
+            capabilities:
```
monotek, line 90: is defined twice: https://github.com/prometheus-community/helm-charts/blob/65b9e4fcf1ffdeac6cc3317d49cc7874bad591f9/charts/prometheus-blackbox-exporter/templates/deployment.yaml#L79

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/deployment.yaml
daniel-maganto: I did some refactor to simplify the syntax and avoid duplication

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -88,6 +88,9 @@ runAsGroup: 1000
 readOnlyRootFilesystem: true
 runAsNonRoot: true
 allowPrivilegeEscalation: false
+capabilities:
+  dropAllCapabilities: true
+  allowIcmp: false
```
monotek, line 93: this introduces a breaking change.

imho it would be better to implement it like this, keeping the current default values: 
https://github.com/prometheus-community/helm-charts/blob/0c70d2b4855fc096716afc07b648f6230207409d/charts/prometheus-json-exporter/values.yaml#L28-L40

monotek (changes requested):

charts/prometheus-blackbox-exporter/values.yaml
daniel-maganto: You're right, we can simplify it, but it takes `allowIcmp` for other resources so that is cause I prefer to continue with this approach. So if you agree, we can remove the capabilities indentation and leave all as root level to avoid breaking changes

daniel-maganto:
hello @monotek , I already did the changes as you mentioned. Update the documentation and bump the chart version to 7.0.0 because this is a breaking change.

monotek (changes requested):
Could we please also update Blackbox exporter to the latest stable version?

daniel-maganto:
> Could we please also update Blackbox exporter to the latest stable version?

Done

monotek (approved):

PR #2355, opened 2022-08-08
[prometheus-statsd-exporter]: parameterize readiness and liveness pro…

### LINKED ISSUE #2305: [prometheus-statsd-exporter] re-point liveness/readiness probe endpoint  to '/health`

Using helm chart version `v0.22.1`.

Currently, the helm chart sets up the statsd-exporter deployment with the `liveness` / `readiness` probes configured to go to `/metrics` endpoint, however the web server does expose the `/health` api endpoint too.

This can potentially become a problem as I recently discovered. The `/metrics` endpoint _can_ get relatively unresponsive depending on the number of metrics it has to serve out at each request, which in turn depends up on the cache size at any given point in time, especially if the metrics expiration `TTL` is set relatively high.

In our case, there was a single instance deployment, with the `TTL` set to `12h` initially. With this setup, and a relatively higher volume of metrics flowing in, we noticed that the pod kept getting restarted every few hours as the readiness probe started failing. Investigating further I noticed that the chart sets up the  default  `liveness/readiness` probes path to  `/metrics`, with a timeout of `1s`, which is causing the probes to fail, leading to the pod restart. 

In my opinion, the `liveness/readiness` probe path should be set to `/health` by default as 
* it serves the intended purpose, and is extremely lightweight. 
* secondly, as the `/metrics` endpoint is production, I'd rather not overload it with probes, especially given the default timeout in the chart is only `1s`.  
* seems to be the established pattern across `k8s`

Increasing the replica count to more than 1 better distribute the load, but eventually, the pods do get restarted once the metrics cache size becomes big, and the `/metrics` endpoint becomes too overloaded to respond to health probes in a timely manner.

### Additional context.

_No response_

makeittotop:
…bes using `/health` instead of `/metrics` endpoint (#2305)

    
Signed-off-by: Abhishek Pareek <makeittotop@gmail.com>

- fixes #2305 

cc @scDisorder 

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-statsd-exporter]`)

### DIFF
--- a/charts/prometheus-statsd-exporter/Chart.yaml
+++ b/charts/prometheus-statsd-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.5.0
+version: 0.5.1
 appVersion: 0.22.7
 home: https://github.com/prometheus/statsd_exporter
 sources:
--- a/charts/prometheus-statsd-exporter/Chart.yaml
+++ b/charts/prometheus-statsd-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.5.1
+version: 0.6.0
 appVersion: 0.22.7
 home: https://github.com/prometheus/statsd_exporter
 sources:
--- a/charts/prometheus-statsd-exporter/templates/deployment.yaml
+++ b/charts/prometheus-statsd-exporter/templates/deployment.yaml
@@ -79,13 +79,9 @@ spec:
               protocol: UDP
             {{- end }}
           livenessProbe:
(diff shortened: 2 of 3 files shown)


charts/prometheus-statsd-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.5.0
+version: 0.5.1
```
monotek, line 4: New feature should raise minor version.

monotek (changes requested):

charts/prometheus-statsd-exporter/Chart.yaml
zanhsieh:
```suggestion
version: 0.6.0
```

monotek (approved):

PR #2370, opened 2022-08-15
feat[kube-state-metrics]: add image sha support for image specified

### LINKED ISSUE #2369: [kube-state-metrics] image tag and sha logic support

Currently the deployment for the chart does not support the option to add a SHA256 image sha. Other prometheus-community charts already support this. The option to use it should be available.

Add a new field under values.yaml called `sha` for the image. Add all the necessary logic withing the deployment to account for it when specified.

Tried using the tag as the SHA256, but since the '`:`' is part of the image string, it will not work.

### Additional context.

Example of other prometheus-community chart using this:
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L41

francRang:
Signed-off-by: francisco.rangel-heb <rangel.francisco@heb.com>

This PR adds support for image sha (also called digest) for the image field. This is needed when signed images are used vs regular images with only tag.

- fixes #2369

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.15.1
+version: 4.16.0
 appVersion: 2.5.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.15.0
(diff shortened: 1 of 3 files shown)


charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.15.0
+version: 4.15.1
```
monotek, line 10: New feature should raise minor version

monotek (changes requested):

francRang:
@monotek applied suggested changes, let me know if there's anything I missed.

monotek (approved):

PR #2376, opened 2022-08-16
feat[prometheus-node-exporter]: add image sha support for image specified

### LINKED ISSUE #2375: [prometheus-node-exporter] image tag and sha logic support

Currently the helper method used by the daomonset for the chart does not support the option to add a SHA256 image sha. Other prometheus-community charts already support this. The option to use it should be available.

Add a new field under values.yaml called sha for the image. Add all the necessary logic within the helper method called in the daemonset manifest to use it accordingly.

Tried using the tag as the SHA256, but since the ':' is part of the image string, it will not work.

### Additional context.

Example of other prometheus-community chart using this:
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L41

francRang:

This PR adds support for image sha (also called digest) for the image field. This is needed when signed images are used vs regular images with only tag

- fixes #2375

#### Checklist

- [x] Chart Version bumped


charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 3.3.1
+version: 3.4.1
```
monotek, line 9:
```suggestion
version: 3.4.0
```

monotek (changes requested):

charts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -58,8 +58,13 @@ Create the name of the service account to use
 The image to use
 */}}
 {{- define "prometheus-node-exporter.image" -}}
+{{- if .Values.image.sha -}}
+"{{ .Values.image.repository }}:{{ .Values.image.tag }}@sha256:{{ .Values.image.sha }}"
```
mrueg, line 62: I believe this line got added accidentally?

francRang: No, this is intentional, I want to check for a sha and add the image in that fashion if it is found.

francRang: Ohh I see, sorry, I was grabbing it as an example of the format to use and forgot to delete the line.

francRang:
I think adding the changes via Github UI messed up the signatures, I am going to recreate this PR

PR #2384, opened 2022-08-21
[prometheus-postgres-exporter] Change queries to map

amit-o:
Signed-off-by: Amit Oren <amit@coralogix.com>

This allows us to override keys in the queries field. Right now we have to copy paste the entire section if we want to change add or omit something.

- fixes #

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
This would be a breaking change and therefore major version has to be raised. Not sure if worth it.

zanhsieh:
Agreed with @monotek .

PR #2407, opened 2022-08-29
[prometheus-node-exporter] custom servicemonitor apiversion

lusson-luo:
Signed-off-by: lusson <lusson@foxmail.com>

Add values prometheus.monitor.apiVersion to modify serviceMonitor api.
To meet the needs of modifying the apiversion of serviceMonitor to access the monitoring platforms of some cloud manufacturers.

no

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-node-exporter/templates/servicemonitor.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.prometheus.monitor.enabled }}
-apiVersion: monitoring.coreos.com/v1
+apiVersion: {{ .Values.prometheus.monitor.apiVersion | default "monitoring.coreos.com/v1" }}
 kind: ServiceMonitor
 metadata:
   name: {{ template "prometheus-node-exporter.fullname" . }}
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -51,6 +51,8 @@ prometheus:
     metricRelabelings: []
     interval: ""
     scrapeTimeout: 10s
+    ## prometheus.monitor.apiVersion ApiVersion for the serviceMonitor Resource(defaults to "monitoring.coreos.com/v1")
+    apiVersion: ""
 
 ## Customize the updateStrategy if set
 updateStrategy:
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.2.1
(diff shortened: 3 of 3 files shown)


zanhsieh (dismissed):

zanhsieh:
@lusson-luo 
Could you sign DCO please? Thank you.

charts/prometheus-node-exporter/values.yaml
```diff
@@ -28,7 +28,7 @@ env: {}
 
 prometheus:
   monitor:
-    enabled: false
+    enabled: true
```
monotek, line 31: Unrelated change?
Would break update. 
Please remove.

monotek (changes requested):

lusson-luo:
> @lusson-luo Could you sign DCO please? Thank you.

ok，pass the DCO

zanhsieh (changes requested):
@lusson-luo 
Could you fix this lint error please?

lusson-luo (commented):
i have review my change

charts/prometheus-node-exporter/templates/servicemonitor.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.prometheus.monitor.enabled }}
-apiVersion: monitoring.coreos.com/v1
+apiVersion: {{ .Values.prometheus.monitor.apiVersion | default "monitoring.coreos.com/v1" }}
```
monotek, line 2: Imho this could be more generic. Instead if defining API version manually it could be checked which API version is available and the most recent one is used.

monotek (changes requested):

charts/prometheus-node-exporter/templates/servicemonitor.yaml
zanhsieh: I feel @lusson-luo intends to forcibly override apiVersion for matrix build purposes. Auto detect which apiVersion available is less concerned.

lusson-luo: > what api exactly is wanted? if its just version change it could be done automatically.
> 
> Something like:
> 
> `{{- if .Capabilities.APIVersions.Has "route.openshift.io/v1/Route" }}`

@monotek The scenario is like this, 
I encapsulate a node-exporter helm chart, which will be deployed on the k8s bases of various cloud manufacturers. The monitoring platforms of cloud manufacturers are different, but most of them will provide servicemonitors with promethues-like rules, but apiversion does not The same. When the operation and maintenance personnel deploy my charts, they can adapt only by modifying the values, without changing the helm source code.

zanhsieh (approved):

monotek (approved):

PR #2411, opened 2022-08-30
[kube-prometheus-stack] Hinting containers' image source values for offline deployment

moroviintaas:
[kube-prometheus-stack]
* Hint for replacing image sources for containers:
	- grafana
	- sidecar
	- kube-state-metrics
	- prometheus-node-exporter

Signed-off-by: moroviintaas <moroviintaas@gmail.com>

[kube-prometheus-stack]

* Bump chart version

Signed-off-by: moroviintaas <moroviintaas@gmail.com>

I was deploying kube-prometheus-stack on offline environment with internal image registry. I have created own values.yml file with changed image sources pointing to my registry. Yet some images failed to be pulled, as they had been defined in dependency charts. Maybe it is worth to note that default grafana image tag (in dependency chart) is currently set to "", which should resolve to "latest" as far as I know and potentially lead to stability issue.   

This PR does not change chart behavior, only adds comment hints on where I had to insert my registry paths in kube-prometheus-stack values.yml file in order to set up containers. It may save some time of digging for other offline deployers.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -22,7 +22,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 39.9.0
+version: 39.10.0
 appVersion: 0.58.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -703,6 +703,14 @@ grafana:
   enabled: true
   namespaceOverride: ""
 
+  ## Image and version of grafana. If not provided (left commented out) default values from grafana charts will be used.
+  ##
+  # image:
+  #   repository: docker.io/grafana/grafana
+  #   tag: 9.0.6
+  #   sha: ""
+  #   pullPolicy: IfNotPresent
+
   ## ForceDeployDatasources Create datasource configmap even if grafana deployment has been disabled
   ##
   forceDeployDatasources: false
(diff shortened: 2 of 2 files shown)


Xtigyro (changes requested):
Thanks for the contribution, @moroviintaas! 🥇

Could you please bump the minor (and not the bugfix) chart version in `Chart.yaml`?

monotek (dismissed):
I don't think that's needed as the defaults are used anyway. Values file is already big enough.

Xtigyro:
> I don't think that's needed as the defaults are used anyway. Values file is already big enough.

@monotek We do that for other components too. I think it's a good practice.

moroviintaas:
> Thanks for the contribution, @moroviintaas! 🥇
> 
> Could you please bump the minor (and not the bugfix) chart version in `Chart.yaml`?

Done in replace commit [9ceee96](https://github.com/prometheus-community/helm-charts/commit/9ceee968f9375f8f73f1b75bc68bbceda3b517d2)

moroviintaas:
> I don't think that's needed as the defaults are used anyway. Values file is already big enough.

In most cases it works well, however when I needed to change image sources (offline deployment), I firstly changed only these visible in this file and deployment failed. Components [prometheus-node-exporter](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-node-exporter) and [kube-state-metrics](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) are managed in this very repository so in online it is ok. However sidecar's and grafana's image is defaulted in external repository [grafana/helm-charts](https://github.com/grafana/helm-charts), with sidecar tagged properly and grafana not tagged (resolving to latest). However I am new to Kubernetes and Helm so I might be wrong.

Xtigyro (approved):

monotek:
@Xtigyro imho Bugfix version would have been right, as no new features have been added.

PR #2441, opened 2022-09-09
[prometheus-node-exporter] Add sidecarHostVolumeMounts

huangkevin404:
Signed-off-by: Kevin Huang <git@kevin.huang.to>

To export the disk's SMART status using a sidecar attached to node-exporter, we would like to mount the /dev from the node to the sidecar in read-only.

To do so, we add a `sidecarHostVolumeMounts` value, that works similarly to `extraHostVolumeMounts`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -146,7 +146,7 @@ spec:
             {{- end }}
 {{- if .Values.sidecars }}
 {{ toYaml .Values.sidecars | indent 8 }}
-          {{- if or (.Values.sidecarVolumeMount)(.Values.sidecarHostVolumeMounts)}}
+          {{- if or .Values.sidecarVolumeMount .Values.sidecarHostVolumeMounts }}
           volumeMounts:
             {{- range $_, $mount := .Values.sidecarVolumeMount }}
             - name: {{ $mount.name }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -146,13 +146,21 @@ spec:
             {{- end }}
 {{- if .Values.sidecars }}
 {{ toYaml .Values.sidecars | indent 8 }}
-          {{- if .Values.sidecarVolumeMount }}
+          {{- if or (.Values.sidecarVolumeMount)(.Values.sidecarHostVolumeMounts)}}
```
zanhsieh, line 149: @huangkevin404 
Could you explain in what scenario we need extra bracket `()` for each variable in order to do `or` operation? Why we could not just do this:

```suggestion
          {{- if or .Values.sidecarVolumeMount .Values.sidecarHostVolumeMounts }}
```

zanhsieh (changes requested):

charts/prometheus-node-exporter/templates/daemonset.yaml
huangkevin404: @zanhsieh 
I'm not super familiar with helm syntax. It seems that you're right, the extra brackets are not needed here, thx for the suggestion !

zanhsieh (approved):

PR #2446, opened 2022-09-11
[prometheus] fix double securityContext for server-container in prometheus-deployment

unki:
If you set `server.containerSecurityContext` (when `kind: Deployment`) helm-install fails with the following error

```
error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
        line 95: mapping key "securityContext" already defined at line 49
```

Apparently by some changes earlier this year `securityContext` gets set twice for the server-container.

This PR removes one of them and also aligns a similar section in `kind: StatefulSet` to look the same as in the deployment.

### DIFF
--- a/charts/prometheus/templates/server/sts.yaml
+++ b/charts/prometheus/templates/server/sts.yaml
@@ -192,9 +192,9 @@ spec:
           {{- if .Values.server.extraVolumeMounts }}
           {{ toYaml .Values.server.extraVolumeMounts | nindent 12 }}
           {{- end }}
-          {{- if .Values.server.containerSecurityContext }}
+          {{- with .Values.server.containerSecurityContext }}
           securityContext:
-            {{- toYaml .Values.server.containerSecurityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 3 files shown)


charts/prometheus/templates/server/sts.yaml
```diff
@@ -194,6 +192,10 @@ spec:
           {{- if .Values.server.extraVolumeMounts }}
           {{ toYaml .Values.server.extraVolumeMounts | nindent 12 }}
           {{- end }}
+          {{- if .Values.server.containerSecurityContext }}
+          securityContext:
+            {{- toYaml .Values.server.containerSecurityContext | nindent 12 }}
+          {{- end }}
```
zanhsieh, line 198: Why not using `with`?

```suggestion
          {{- with .Values.server.containerSecurityContext }}
          securityContext:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

If somebody defines `.Values.server.containerSecurityContext` as `{}` then it shall output nothing.

zanhsieh (changes requested):

zanhsieh (approved):

PR #2457, opened 2022-09-14
[kube-state-metrics] - Add option to merge namespaces

Allex1:
…itional namespaces

Signed-off-by: birca <Allex21ro@yahoo.com>

Add option to merge .Release.Namespace and add additional namespaces

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.18.0
+version: 4.19.0
 appVersion: 2.6.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -74,7 +74,9 @@ spec:
         {{- if .Values.metricDenylist }}
         - --metric-denylist={{ .Values.metricDenylist | join "," }}
         {{- end }}
-        {{- if .Values.releaseNamespace }}
+        {{- if and .Values.releaseNamespace .Values.namespaces}}
+        - --namespaces={{ template "kube-state-metrics.namespace" . }},{{ tpl (.Values.namespaces | join ",") $ }}
+        {{- else if .Values.releaseNamespace }}
         - --namespaces={{ template "kube-state-metrics.namespace" . }}
         {{- else if .Values.namespaces }}
         - --namespaces={{ tpl (.Values.namespaces | join ",") $ }}
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -211,7 +211,7 @@ kubeconfig:
   secret:
 
 # Enable only the release namespace for collecting resources. By default all namespaces are collected.
-# If releaseNamespace and namespaces are both set only releaseNamespace will be used.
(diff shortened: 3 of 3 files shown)


dotdc:
Hi @Allex1,

I don't think this patch is necessary. As I see it, you can just fill `.Values.namespaces`  with your namespaces list and include you release namespace with them. 

If not, can you give more context on your use-case?

Allex1:
@dotdc that's what we did so far and worked. The problem is when you deploy the chart in a namespace that you don't know the name of before. We have some gitops use cases where this happens and we would like to be able to use the global helm variables where possible.

Allex1:
@dotdc @mrueg @tariq1890  can you pls take a look?

dotdc:
Hi @Allex1,

Sorry but this patch does'nt look good to me.
All namespaces should be collected by default and you have the ability to either:
- set `releaseNamespace` to scrape only the release namespace
- set `namespaces` with a custom include list.
- set `namespacesDenylist` to exclude a set of  namespaces to exclude

You should be good to go with theses, if not, please give even more details on your use-case.
Especially, how can you deploy a chart to a namespace that you are not aware of?

From the values file:
```
# Enable only the release namespace for collecting resources. By default all namespaces are collected.
# If releaseNamespace and namespaces are both set only releaseNamespace will be used.
releaseNamespace: false

# Comma-separated list of namespaces to be enabled for collecting resources. By default all namespaces are collected.
namespaces: ""

# Comma-separated list of namespaces not to be enabled. If namespaces and namespaces-denylist are both set,
# only namespaces that are excluded in namespaces-denylist will be used.
namespacesDenylist: ""
```

@mrueg @monotek What do you think about this?

Allex1:
@dotdc We have a cicd pipeline with a wrapper on top that generates random ns names before deploying apps and monitoring into them.
We're running in a multi tenant k8s cluster where any user can get access to one or more ns's.
When we deploy monitoring in a separate namespace we don't know it's name but we know the app namespaces names previously deployed.
Basically we want to merge the global Release.Namespace variable with a custom include list instead of having them mutually exclusive. This is currently possible with the [Prometheus-operator helm chart](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L1658)

P.S. Updated the docs as well.

charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.18.0
+version: 4.18.1
```
dotdc, line 10: As this is a new feature, can you bump chart's version to 4.19.0 please

dotdc (changes requested):

charts/kube-state-metrics/Chart.yaml
Allex1: @dotdc done

dotdc (approved):
LGTM

dotdc:
Still looks like a corner case to me, but I guess we can merge this because it shouldn't impact anyone else.

Thank you for your contribution!

Allex1:
> Still looks like a corner case to me, but I guess we can merge this because it shouldn't impact anyone else.
> 
> Thank you for your contribution!

@dotdc I agree that is a weird use case, but the main reason is for consistency with the other charts.

Current options for the various charts are:
- all namespaces
- release namespace
- custom list
- merge between 2 and 3

The problem is that the default behavior is different across charts. 

I will submit a pr that enables this for Grafana as well for that reason.

PR #2462, opened 2022-09-18
[prometheus-pushgateway] update notes.txt notify port

lusson-luo:

prometheus-pushgateway pod port is 9091, but install notify is 8080. i update to real port 9091.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-pushgateway/templates/NOTES.txt
+++ b/charts/prometheus-pushgateway/templates/NOTES.txt
@@ -14,6 +14,6 @@
   echo http://$SERVICE_IP:{{ .Values.service.port }}
 {{- else if contains "ClusterIP" .Values.service.type }}
   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "prometheus-pushgateway.name" . }},release={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
-  echo "Visit http://127.0.0.1:8080 to use your application"
-  kubectl port-forward $POD_NAME 8080:80
+  echo "Visit http://127.0.0.1:9091 to use your application"
+  kubectl port-forward $POD_NAME 9091:80
 {{- end }}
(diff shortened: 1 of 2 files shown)


zeritti:
I allow myself a comment on the change. The proposed change would not yield the desired results - the connection would get forwarded from the local 9091/tcp to the pod's 80/tcp. There is, however, nothing listening at 80/tcp in the pod by default.
I would suggest the change as follows:
```
kubectl port-forward $POD_NAME 9091
```
so that the user can just point their browser to `http://localhost:9091`.

Alternatively, if a "known" unprivileged port number is desired, then
```
kubectl port-forward $POD_NAME 8080:9091
```
which would mean opening `http://localhost:8080` in a browser.

charts/prometheus-pushgateway/templates/NOTES.txt
```diff
@@ -14,6 +14,6 @@
   echo http://$SERVICE_IP:{{ .Values.service.port }}
 {{- else if contains "ClusterIP" .Values.service.type }}
   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "prometheus-pushgateway.name" . }},release={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
-  echo "Visit http://127.0.0.1:8080 to use your application"
-  kubectl port-forward $POD_NAME 8080:80
-{{- end }}
\ No newline at end of file
+  echo "Visit http://127.0.0.1:9091 to use your application"
+  kubectl port-forward $POD_NAME 9091:80
```
monotek, line 18:
```suggestion
  kubectl port-forward $POD_NAME 9091
```

monotek (changes requested):
I don't think port 80 is a good idea, as that would need privileged linux user to bind to it. 

So imho original port should be used.

lusson-luo:
i change as follows:

`kubectl port-forward $POD_NAME 9091`

then open browser to http://localhost:9091

monotek (approved):

Issue #1936, opened 2022-03-30
[prometheus-blackbox-exporter] ServiceMonitor for blackbox /metrics endpoint


OneCricketeer:
### Is your feature request related to a problem ?

It is not possible to use `serviceMonitor.enabled` + targets (for example, a target "loopback" to the exporter) to scrape the blackbox-exporter's `/metrics` endpoint itself; [the paths are all hard-coded as `/probe`.](https://github.com/prometheus-community/helm-charts/blob/prometheus-blackbox-exporter-5.5.1/charts/prometheus-blackbox-exporter/templates/servicemonitor.yaml#L24)

### Describe the solution you'd like.

By default, with no "external" targets, there should be a ServiceMonitor for the blackbox-exporter Service itself, scraping `/metrics`. 

e.g.

```yml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: prometheus-blackbox-exporter-monitor
  namespace: 'default' # example
  labels: {}
spec:
  endpoints:
  - path: /metrics
    interval: 30s
    port: http
    scheme: http
  namespaceSelector:
    matchNames:
      - 'default'  # example
  selector:
    matchLabels:
      app.kubernetes.io/instance: prometheus-blackbox-exporter
      app.kubernetes.io/name: prometheus-blackbox-exporter
```

### Describe alternatives you've considered.

Write my own ServiceMonitor (above)

tr3mor:
Not sure why this one is closed, I think blackbox chart definitely need option to deploy selfmonitor like Prometheus has 
https://github.com/prometheus-community/helm-charts/blob/88447d1886a982b0e26f047d9973d2f6bad7fe7e/charts/kube-prometheus-stack/values.yaml#L409
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/servicemonitor.yaml
I can bring PR if needed

PR #2471, opened 2022-09-20
[prometheus-blackbox-exporter] Add selfMonitor support

tr3mor:

Now if you deploy blackbox-exporter via this chart it will not create ServiceMonitor for itself, only for targets. This PR provides the ability to do it by overriding variables and enabling selfMonitor, similar to what kube-prometheus-stack chart does for alert-manager for example https://github.com/prometheus-community/helm-charts/blob/63a64f62d19f2c505a5139e91dc544ad1891273e/charts/kube-prometheus-stack/values.yaml#L405-L409 The problem was originally raised in #1936

Not sure about value structure and formatting, open to suggestions
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/values.yaml
+++ b/charts/prometheus-blackbox-exporter/values.yaml
@@ -213,7 +213,7 @@ extraArgs: []
 replicas: 1
 
 serviceMonitor:
-  ## If true, a ServiceMonitor CRD is created for a prometheus operator 
+  ## If true, a ServiceMonitor CRD is created for a prometheus operator
   ## https://github.com/coreos/prometheus-operator for blackbox-exporter itself
   ##
   selfMonitor:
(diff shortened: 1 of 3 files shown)


charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -213,8 +213,21 @@ extraArgs: []
 replicas: 1
 
 serviceMonitor:
-  ## If true, a ServiceMonitor CRD is created for a prometheus operator
-  ## https://github.com/coreos/prometheus-operator
+  ## If true, a ServiceMonitor CRD is created for a prometheus operator 
+  ## https://github.com/coreos/prometheus-operator for blackbox-exporter itself
+  ##
+  selfMonitor: false
```
monotek, line 219: i would prefer it like this:

```suggestion
  selfMonitor: 
    enabled: false
    additionalMetricsRelabels: {}
    additionalRelabeling: []
    labels: {}
    interval: 30s
    scrapeTimeout: 30s
```

monotek (changes requested):

charts/prometheus-blackbox-exporter/values.yaml
tr3mor: Updated

monotek (changes requested):
please fix yaml lint

monotek (approved):

PR #2478, opened 2022-09-22
Added new diagram

XDavidT:
Signed-off-by: David Tayar <38883783+XDavidT@users.noreply.github.com>

Only new diagram


kylewin (approved):

monotek (changes requested):
pleaee use issue template and raise chart version

zanhsieh:
@XDavidT 
The diagram is not accessible.

dotdc:
Don't think the diagram is a good addition.
But if we do add it, it's probably better to load it directly from GitHub rather than an external website like imgur.com. 

My two cents.

PR #2500, opened 2022-09-27
[prometheus-cloudwatch-exporter] Add extraEnv to values file to specify extra env vars being added to the deployment

AClarkie:

This PR adds `extraEnv` as an option in the values.yaml which can be used to add extra environment variables to the deployment. To simplify the `env` logic I've added `POD_IP` as an env var that will always be present.

The reason it's needed is that I want the option to set an environment variable `AWS_SHARED_CREDENTIALS_FILE` to instruct where the aws credentials file exists on disk, so that I can use the vault injector to mount the creds to a specific directory.

I'm happy to modify the logic if required, please let me know

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
@@ -44,8 +44,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
-        {{- if .Values.extraEnv }}
-        {{- .Values.extraEnv  | toYaml | nindent 12 }}
+        {{- with .Values.extraEnv }}
+        {{- toYaml . | nindent 12 }}
         {{- end }}
         {{- if not .Values.aws.role }}
         {{- if .Values.aws.secret.name }}
--- a/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
@@ -39,9 +39,16 @@ spec:
     {{- end }}
       containers:
         - name: {{ .Chart.Name }}
+          env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+        {{- if .Values.extraEnv }}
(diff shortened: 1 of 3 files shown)


AClarkie:
@asherf @gianrubio @torstenwalter any chance we can get some eyes on this today?

charts/prometheus-cloudwatch-exporter/templates/deployment.yaml
```diff
@@ -39,9 +39,16 @@ spec:
...
       containers:
         - name: {{ .Chart.Name }}
+          env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+        {{- if .Values.extraEnv }}
+        {{- .Values.extraEnv  | toYaml | nindent 12 }}
+        {{- end }}
```
zanhsieh, line 49: This should be better to deal with empty object `{}`:

```suggestion
        {{- with .Values.extraEnv }}
        {{- toYaml . | nindent 12 }}
        {{- end }}
```

zanhsieh (changes requested):

monotek (changes requested):
Why don't you define the pod ip via extraenv? 
Imho it's not needed to be hard coded.

AClarkie:
> Why don't you define the pod ip via extraenv?
Imho it's not needed to be hard coded.

It's purely to simplify the env logic otherwise you need to check if each of the other variables like `.Values.aws.secret.name`, `.Values.aws.aws_secret_access_key` etc. are set before adding in the `env:` which imo would complicate the setup. A very similar approach is taken in the other community charts, see the below examples:

https://github.com/prometheus-community/helm-charts/blob/d7a033184b896d865b76507e585366c86d3cec65/charts/prometheus/templates/alertmanager/deploy.yaml#L53

https://github.com/prometheus-community/helm-charts/blob/d7a033184b896d865b76507e585366c86d3cec65/charts/prometheus/templates/alertmanager/sts.yaml#L52

monotek (approved):

monotek (changes requested):

monotek (approved):

PR #2507, opened 2022-09-30
[kube-state-metrics] Allow namespaces to be defined as both string/list

Allex1:
Allow namespaces to be defined as both string/list.
This was inspired by the fact that Prometheus-Operator requires [namespaces list](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L1664) and not string. By supporting this a kube-prometheus-stack user can use [yaml anchors ](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) in values.yaml in order to define the watched namespaces once and reuse in both deployments.

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 4.20.1
+version: 4.20.2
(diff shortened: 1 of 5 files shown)


amuraru (approved):

dotdc (approved):
LGTM

PR #2512, opened 2022-10-01
[prometheus-json-exporter] - Allow to pass name of the module for scr…

faisal-khalique:
Signed-off-by: Faisal K <khalique.faisal@gmail.com>

The [json-exporter](https://github.com/prometheus-community/json_exporter) provides the option of [passing name of the module](https://github.com/prometheus-community/json_exporter/blob/v0.5.0/cmd/main.go#L90) which should be picked up from [config.yml](https://github.com/prometheus-community/json_exporter/blob/master/examples/config.yml) for scraping metrics from the associated target.
However, [service monitor](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-json-exporter/templates/servicemonitor.yaml#L25) template for [json-exporter](https://github.com/prometheus-community/json_exporter) [helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-json-exporter) does not provide such an option to be passed.
This change allows the name of the module to be passed (optionally) along with the target name and URL.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-json-exporter/Chart.yaml
+++ b/charts/prometheus-json-exporter/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
+version: 0.5.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/prometheus-json-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-json-exporter/templates/servicemonitor.yaml
@@ -25,6 +25,10 @@ spec:
     params:
       target:
       - {{ .url }}
+      {{- if .module }}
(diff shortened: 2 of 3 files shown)


faisal-khalique:
I wanted to clarify few points:

1. Should name of the module (to read from config.yml for a target scraping) be same as the suffix used for name of the service monitor i.e. `.Values.serviceMonitor.targets.[*].name` be used for `module` query param in `http://localhost:7979/probe?module=default&target=http://example.com/healthz` ? To me it didn't make sense as I could have multiple modules defined for the same target endpoint but again we are defining only service monitor for every combination of target (and module with this PR)
2. When name of the `module` is not provided, [json-exporter](https://github.com/prometheus-community/json_exporter/blob/master/cmd/main.go#L92) picks up `default`. I prefer to use defaults from the application rather than defining default values in the helm chart. Is this ok or it is preferred that we provide defaults from helm chart as we define the config which has this `default` section here in the helm chart itself?

monotek (approved):

zanhsieh (approved):

PR #2520, opened 2022-10-04
[prometheus-node-exporter] use recent version of PSP API in K8S 1.25

JacekLakis-TomTom:
policy/v1beta1 for PodSecurityPolicy is deprecated in K8S 1.25. This change configures template to recognize k8s version and ses appropriate api version.  

prometheus-node-exporter chart broken on 1.25

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/psp.yaml
+++ b/charts/prometheus-node-exporter/templates/psp.yaml
@@ -1,5 +1,6 @@
 {{- if .Values.rbac.create }}
 {{- if .Values.rbac.pspEnabled }}
+{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
@@ -46,5 +47,6 @@ spec:
       - min: 0
         max: 65535
   readOnlyRootFilesystem: false
+{{- end }}
 {{- end }}
 {{- end }}
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.3.0
+version: 4.3.1
 appVersion: 1.3.1
 home: https://github.com/prometheus/node_exporter/
 sources:


monotek (approved):

monotek (changes requested):
please fix lint

mrueg:
I don't think this is the correct solution here, as PodSecurityPolicy Admission Controller got replaced with PodSecurityAdmission in v1.25. See also: https://kubernetes.io/docs/concepts/security/pod-security-policy/

charts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -105,3 +105,15 @@ Create the namespace name of the service monitor
...
+Return the appropriate apiVersion for podSecurityPolicy.
+*/}}
+{{- define "podSecurityPolicy.apiVersion" -}}
+{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+{{- if semverCompare "<1.25-0" $kubeTargetVersion -}}
+{{- print "policy/v1beta1" -}}
+{{- else -}}
+{{- print "policy/v1" -}}
+{{- end -}}
+{{- end -}}
```
zanhsieh, line 119: Actually this is not needed due to PSP deprecated.

```suggestion
```

charts/prometheus-node-exporter/templates/psp.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.rbac.create }}
 {{- if .Values.rbac.pspEnabled }}
-apiVersion: policy/v1beta1
+apiVersion: {{ template "podSecurityPolicy.apiVersion" . }}
```
zanhsieh, line 3: Just disable PSP in case the api is not available. Remember append an `{{- end -}}` at the end of file.

```suggestion
{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
apiVersion: policy/v1beta1
```

zanhsieh (changes requested):

JacekLakis-TomTom:
Perhaps also remove clusterrole and clusterrolebinding from templates?

monotek (changes requested):
i don't see how this now will use v1?
if v1beta1 is not found it will not be used at all?

krisztianfekete (commented):
This currently blocks deploying the chart to K8s 1.25.x.

What about adding a helper like this?

```yaml
{{- define "kube-prometheus-stack.pdb.apiVersion" -}}
 {{ if $.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" -}}
 {{- print "policy/v1" -}}
 {{- else -}}
 {{- print "policy/v1beta1" -}}
 {{- end -}}
 {{- end -}}
```

zanhsieh:
@monotek @JacekLakis-TomTom @krisztianfekete 
In my opinion (and you probably knew), PSP is under `extensions/v1beta1` before k8s 1.16, and on `policy/v1beta1` till k8s 1.25. Given the pace of k8s upgrading trend, I really hard to image some company / people still stick on 1.16. I would say it shall be safe to just drop PSP after k8s 1.25 since PSP never reaches to policy/v1.

@krisztianfekete 
`PodDisruptionBudget` (pdb) is another beast; please do not mess up with PSP.

monotek (approved):

monotek:
@zanhsieh 
right, my fault, i had v1 as the last available api for psp in mind.

PR #2525, opened 2022-10-06
[prometheus-cloudwatch-exporter] Pass metrics configuration from separate file

iflan7744:
Currently, we have to pass Metrics configuration to configmap by defining in values.yaml file. This PR will help to call the metrics from a separate file to keep the values.yaml file clean and error-free. Therefore we can add  all the metrics define in 
https://github.com/prometheus/cloudwatch_exporter/tree/master/examples without growing the values.yaml file.

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
this would break the chart for everybody upgrading from earlier versions.
also changes default settings.

PR #2565, opened 2022-10-16
[prometheus-adapter] Fix PSP deprecation after k8s 1.25+

zanac1986:

Try to fix PodSecurityPolicy being removed after Kubernetes 1.25+.

- This PR just simply removes PSP (same as disable PSP).
- PSP before 1.25+ only reaches `policy/v1beta1`
- `policy/v1` does `NOT` contain PodSecurityPolicy. Please don't mix with PodDisruptionBudget.

#### Checklist

- [x] DCO signed
- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/psp.yaml
+++ b/charts/prometheus-adapter/templates/psp.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.psp.create -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
 ---
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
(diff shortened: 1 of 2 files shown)


charts/prometheus-adapter/templates/psp.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.psp.create -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
```
monotek, line 2: This could be a single line:

```
{{- if and .Values.psp.create .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
```

monotek (changes requested):

monotek (approved):

PR #2540, opened 2022-10-11
[kube-prometheus-stack]: Alert mgr service monitor: enableHttp2

gadisn:
Signed-off-by: gadisn <gadisn@gmail.com>

#### What this PR does / why we need it
Allow setting `enableHttp2` for alert manager service monitor

#### Which issue this PR fixes
This PR is a continuation to exposing http2 configuration on service monitor in https://github.com/prometheus/prometheus/pull/10492

#### Special notes for your reviewer
When using mtls with istio it might be required to disable http2 - this PR allows it.

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #2570, opened 2022-10-16
[kube-prometheus-stack]: Fix for alert mgr service monitor: enableHttp2

gadisn:
Signed-off-by: gadisn <gadisn@gmail.com>

A fix for the recent https://github.com/prometheus-community/helm-charts/pull/2540.
Since the value here is a boolean, and the default is true, the previous template had no way to set `false` for enableHttp2...
Instead, now we always take the value and and it to the created resource

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -22,7 +22,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 41.4.1
+version: 41.4.2
 appVersion: 0.60.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -22,7 +22,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 41.6.1
+version: 41.7.1
 appVersion: 0.60.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/alertmanager/servicemonitor.yaml
+++ b/charts/kube-prometheus-stack/templates/alertmanager/servicemonitor.yaml
@@ -18,6 +18,7 @@ spec:
       - {{ printf "%s" (include "kube-prometheus-stack.namespace" .) | quote }}
   endpoints:
   - port: {{ .Values.alertmanager.alertmanagerSpec.portName }}
+    enableHttp2: {{ .Values.alertmanager.serviceMonitor.enableHttp2 }}
     {{- if .Values.alertmanager.serviceMonitor.interval }}
     interval: {{ .Values.alertmanager.serviceMonitor.interval }}
     {{- end }}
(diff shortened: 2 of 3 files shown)


monotek (changes requested):
Don't think that will work.
See: https://github.com/prometheus-community/helm-charts/pull/2549

gadisn:
> Don't think that will work. See: #2549

@monotek #2549 means a change of default - isn't this a breaking change?
If the default change is indeed intentional, I'll close my PR here

monotek:
When the value was introduced there was an error. See: https://github.com/prometheus-community/helm-charts/commit/bd46559a54b765b6489f8205b87db913ef3e51d5#commitcomment-86546702

gadisn:
> When the value was introduced there was an error. See: [bd46559#commitcomment-86546702](https://github.com/prometheus-community/helm-charts/commit/bd46559a54b765b6489f8205b87db913ef3e51d5#commitcomment-86546702)

Thanks @monotek for the quick fix there. 

But `enableHttp2` **is** a part of `endpoints` , see [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/crds/crd-servicemonitors.yaml#L147).
Can you tell why the error was reported?
Also, again, using `false` as default in values would change the existing behavior (where http2 is being used)

gadisn:
@monotek I think that adding `enableHttp2` without a condition is required - we can't have a condition around it, since upon false in values, `enableHttp2` won't be added at all. And the default behavior is `true` (see [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)).
Currently, I can't set `enableHttp2` to false, which means I can't turn off the use of http2 in a scraping job.

Solutions I can think of:
1. Reverse the flag here to something like `disableHttp2`
2. Always add `enableHttp2` in the template (no condition, as currently exist)

What do would you suggest?

gadisn:
> please raise chart version.
> 
> Have you checked that the [error we saw before](https://github.com/prometheus-community/helm-charts/commit/bd46559a54b765b6489f8205b87db913ef3e51d5#commitcomment-86546702) does not pop up again?
> 
> ```
> Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(ServiceMonitor.spec.endpoints[0]): unknown field "enableHttp2" in com.coreos.monitoring.v1.ServiceMonitor.spec.endpoints
> ```

@monotek I couldn't reproduce the error myself. Also [asked the reporter](https://github.com/prometheus-community/helm-charts/commit/bd46559a54b765b6489f8205b87db913ef3e51d5#commitcomment-87917233) - without response so far.
Were you able to identify why he got his error?

monotek:
Yes, i can reprocduce, updating your main branch to the patch-3 branch version:

```
 helm upgrade -i kps .
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(ServiceMonitor.spec.endpoints[0]): unknown field "enableHttp2" in com.coreos.monitoring.v1.ServiceMonitor.spec.endpoints

```

gadisn:
Thanks @monotek. I run the same command and didn't get the error.

I believe the error as also seen by you is due to stale CRD for serviceMonitor.

-This new field (`enableHttp2`) was introduced [here](https://github.com/prometheus-community/helm-charts/pull/2320) (by you...).
-I believe it was released with [39.0.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-prometheus-stack-39.0.0).
-According to the [upgrade guide](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-38x-to-39x), the crds should be updated.

Can you please validate that your servicemonitors CRD was indeed updated?
Thanks.

monotek:
As mentioned above i've used your main branch for initial install in a new kind cluster.
Did not check that it's rather old :D
After applying current crds it worked without error.

monotek (approved):
set "enableHttp2" to "true" again and raised version

PR #2596, opened 2022-10-20
[prometheus-statsd-exporter] add PodMonitoring for Google Managed Prometheus

pkchuyen:
Signed-off-by: Chuyen Pham <pkchuyen@users.noreply.github.com>

It creates PodMonitoring resource which allows Google Managed Prometheus creating the scraping config and collect statsd metrics.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-statsd-exporter/Chart.yaml
+++ b/charts/prometheus-statsd-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.6.2
+version: 0.6.3
 appVersion: 0.22.7
 home: https://github.com/prometheus/statsd_exporter
 sources:
--- a/charts/prometheus-statsd-exporter/templates/podmonitor.yaml
+++ b/charts/prometheus-statsd-exporter/templates/podmonitor.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.podMonitor.enabled }}
+apiVersion: monitoring.googleapis.com/v1
+kind: PodMonitoring
+metadata:
+  name: {{ template "prometheus-statsd-exporter.fullname" . }}-metrics
+  namespace: {{ default .Release.Namespace .Values.podMonitor.namespace | quote }}
+  labels: {{- include "prometheus-statsd-exporter.labels" . | nindent 4 }}
+    app.kubernetes.io/component: metrics
+    {{- range $key, $value := .Values.podMonitor.additionalLabels }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "prometheus-statsd-exporter.selectorLabels" . | nindent 6 }}
+  endpoints:
+    - port: http
+      path: {{ .Values.podMonitor.endpoint }}
+      interval: {{ .Values.podMonitor.interval }}
+{{- end }}
--- a/charts/prometheus-statsd-exporter/values.yaml
+++ b/charts/prometheus-statsd-exporter/values.yaml
@@ -76,6 +76,15 @@ serviceMonitor:
   # More info https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
   relabelings: []
 
+podMonitor:
+  # If enabled, It creates Google Managed Prometheus scraping config
+  enabled: false
+  interval: 30s
+  # Set a custom namespace where to deploy PodMonitoring resource
+  namespace: ""
+  # This is the default endpoint of the metrics
+  endpoint: /metrics
+
 serviceAccount:
   # Specifies whether a service account should be created
   create: true


monotek (changes requested):

charts/prometheus-statsd-exporter/values.yaml
```diff
@@ -76,6 +76,15 @@ serviceMonitor:
   # More info https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
   relabelings: []
 
+podMonitor:
```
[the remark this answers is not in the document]
pkchuyen: Thanks @monotek for the suggestion.
I think it is a good idea to combile the config for both serverMonitor and podMonitor.

However, I will prefer to set [2 flags to enable serviceMonitor and podMonitor saparately](https://github.com/prometheus-community/helm-charts/pull/2596/files#diff-5e6d51ef06ed55367a5c12de02ad1caccc4bd17568b4d1ec97d90d5aa11d1c04R69). Because it it needed for the migration from self-hosted prometheus to Google Managed Prometheus, At some points we still need to maintainance 2 monitor systems at the same time until all the services fully migrated to GMP (at least this is the scenario we are facing now)

charts/prometheus-statsd-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.6.2
+version: 0.6.3
```
monotek, line 4: New feature should raise minor version

charts/prometheus-statsd-exporter/templates/podmonitor.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.prometheus.monitor.podMonitor }}
+apiVersion: {{ .Values.prometheus.monitor.apiVersion | default "monitoring.googleapis.com/v1" }}
+kind: PodMonitoring
```
monotek, line 3: Kind should be configurable too, as it's "PodMonitor" for Prometheus.

charts/prometheus-statsd-exporter/values.yaml
```diff
@@ -63,18 +63,49 @@ readinessProbe:
     path: /health
     port: http
 
-serviceMonitor:
```
monotek, line 66: Please keep the current defaults. Otherwise it breaks for everyone updating from earlier versions.

Please also keep Servicemonitor and PodMonitor config separated.

monotek (changes requested):

charts/prometheus-statsd-exporter/values.yaml
pkchuyen: I am a bit confused between your first comment and this one.
so I update the PR and hope I got your point correctly.

Otherwise, could you please provide more context of your expectation on supporting Google Managed Prometheus config for this helm chart.

monotek (commented):
The question if if we even need servicemonitor and podmonitoring at the same time, when the servicemonitor would be configurable to act as podminoitoring resource too?

pkchuyen:
> The question if if we even need servicemonitor and podmonitoring at the same time, when the servicemonitor would be configurable to act as podminoitoring resource too?

we might not need serviceMonitor and podMonitoring at the same time. But it is needed to have at the migration state.
at least, in our case, we are running self-hosted Prometheus with serviceMonitor enabled and we need to switch the moinitoring system to use Google Managed Prometheus.

at that time, we cannot switch everything in one day, so we need to keep both serviceMonitor and podMonitoring running in order to do the migration for alerting rules, dashboard, etc.

I afaird serviceMonitor cannot configure to act as podMonitoring resource. Google docs has a document to guide `Convert existing prometheus-operator resources` [here](https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#gmp-servicemonitor)

charts/prometheus-statsd-exporter/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: prometheus-statsd-exporter
 description: A Helm chart for prometheus stats-exporter
-version: 0.6.2
-appVersion: 0.22.7
+version: 0.7.0
+appVersion: "v0.22.8"
```
monotek, line 5:
```suggestion
appVersion: v0.22.8
```

monotek (changes requested):

charts/prometheus-statsd-exporter/values.yaml
```diff
@@ -63,18 +63,37 @@ readinessProbe:
...
+    endpoint: "/metrics"
+
 serviceMonitor:
   enabled: false
   interval: 30s
   scrapeTimeout: 10s
   namespace: monitoring
   honorLabels: false
   additionalLabels: {}
+  ## resource kind(defaults to "PodMonitoring")
```
monotek, line 86: this seems wrong

monotek (changes requested):

charts/prometheus-statsd-exporter/values.yaml
pkchuyen: fixed it, 
the description was wrong but[ serviceMonitor resource ](https://github.com/prometheus-community/helm-charts/pull/2596/files#diff-541ccd8e525b85e105d07f955e7b683d80319625191918e67d53e1ea79b3b103R3)is correct

charts/prometheus-statsd-exporter/templates/servicemonitor.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       {{- include "prometheus-statsd-exporter.selectorLabels" . | nindent 6 }}
   endpoints:
     - port: http
-      path: "/metrics"
+      path: {{ .Values.prometheus.monitor.endpoint }}
```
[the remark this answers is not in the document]
pkchuyen: good catch,
Thank you

monotek (approved):

PR #2601, opened 2022-10-24
[kube-prometheus-stack] When the custom label repeatedly conflicts with the general label, the general label is used first. Generic labels should be used in preference here

chenshijian73-qq:
…ts with the general label, the general label is used first. Generic labels should be used in preference here.

When the custom label repeatedly conflicts with the general label, the general label is used first. Generic labels should be used in preference here.

- fixes #

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
Charts version needs to be raised.

I also don't quite understand why the change is needed?
Normaly labels should not conflict.
If so, why configuring them twice at all?

chenshijian73-qq:
> Charts version needs to be raised.
> 
> I also don't quite understand why the change is needed? Normaly labels should not conflict. If so, why configuring them twice at all?



> Because it is normal not to change the general tag, generally by modifying the custom tag to achieve your own needs

PR #2650, opened 2022-11-06
[prometheus-node-exporter] Refactor

zanac1986:

Refactor 

- From `{{- if .Values.foo }}` to `{{- with }}` and using the context under it
- Get rid of `{{- if .Values.foo }}` when it combines with `{{ range $key, $value := .Value.foo }}`
- Proper indention

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -1,29 +1,28 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: {{ template "prometheus-node-exporter.fullname" $ }}
-  namespace: {{ template "prometheus-node-exporter.namespace" $ }}
+  name: {{ include "prometheus-node-exporter.fullname" . }}
+  namespace: {{ include "prometheus-node-exporter.namespace" . }}
   labels:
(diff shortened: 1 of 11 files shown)


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -1,41 +1,43 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: {{ template "prometheus-node-exporter.fullname" . }}
-  namespace: {{ template "prometheus-node-exporter.namespace" . }}
-  labels: {{ include "prometheus-node-exporter.labels" . | indent 4 }}
+  name: {{ template "prometheus-node-exporter.fullname" $ }}
+  namespace: {{ template "prometheus-node-exporter.namespace" $ }}
+  labels:
+    {{- include "prometheus-node-exporter.labels" $ | indent 4 }}
```
monotek, line 7: i would use the default here, like "helm create" would do it.:
```
metadata:
  name: {{ include "prometheus-node-exporter.fullname" . }}
  labels:
    {{- include "prometheus-node-exporter.labels" . | nindent 4 }}

```

monotek (changes requested):

monotek (approved):

zanhsieh (approved):

PR #2673, opened 2022-11-11
[prometheus] moving node-exporter out of prometheus

zanac1986:
#### What this PR does / why we need it

Move node-exporter the embedded chart back to its parent. As mentioned here:

https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300167738
https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300192978
https://github.com/prometheus-community/helm-charts/issues/2654#issuecomment-1306468145

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #2680, opened 2022-11-14
[prometheus-server] Add prometheus-server helm chart

### LINKED ISSUE #2654: [Prometheus] Standalone helm chart for Prometheus Server

Apologize if I missed this chart but I wasn't able to find one in this repo.

I want to use a helm chart that has solely the Prometheus Server image. The [Prometheus helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus) is the closest to work with, but it has dependency on kube-state-metrics, and has additional component of alert-manager, node_exporter, and push gateway. 

It seems like this repo contains standalone helm chart for different exporters, so having a parallel standalone helm chart for Prometheus server will make sense as well.

Have a standalone prometheus server helm chart.

Modify the [Prometheus helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus), but it's much easier to start from a cleaner state.

### Additional context.

_No response_

ivyostosh:
This PR creates a standalone `prometheus-server` helm chart. Currently, this `helm-charts` repo has standalone charts for different exporters while lacking a standalone `prometheus-server` helm chart. Having this chart enables user to use multiple prometheus helm charts like lego pieces and build observability stack with more flexibility. 

This PR uses the [Prometheus helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus) v15.18.0 as the base, and takes out dependencies on alertmanager, pushgateway, nodeExporter, kubeStateMetrics, and configmapReload. Refer to individual commit messages for details.

I'm happy to maintain this chart so that it stays parallel with the [Prometheus helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus).

- fixes #2654

#### Checklist

- [x] Chart Version bumped


zanhsieh:
@ivyostosh 
I think it might be better to add some tests, not only for ensure this chart itself running ok, but also for helping the integrity. For example,

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-statsd-exporter/templates/tests/test-connection.yaml
https://github.com/zanac1986/grafana-helm-charts/tree/main/charts/grafana/templates/tests

charts/prometheus-server/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+appVersion: 2.39.1
+version: 1.0.0
+description: Prometheus is a monitoring system and time series database.
+home: https://prometheus.io/
+icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/assets/prometheus_logo-cb55bb5c346.png
+sources:
+  - https://github.com/prometheus/prometheus-server
+maintainers:
+  - name: ivyostosh
+    email: ivyjin215@gmail.com
```
zanhsieh, line 12: More than single maintainer would be better.

```suggestion
    email: ivyjin215@gmail.com
  - name: MH
    email: zanhsieh@gmail.com
```

zanhsieh (changes requested):
Also need to add .github/CODEOWNERS:

```
# .github/CODEOWNERS
# line 34
/charts/prometheus-redis-exporter/ @acondrat @zanhsieh
+ /charts/prometheus-server/ @ivyostosh @zanhsieh 
/charts/prometheus-smartctl-exporter/ @kfox1111
/charts/prometheus-snmp-exporter/ @miouge1
```

charts/prometheus-server/Chart.yaml
ivyostosh: Updated

ivyostosh (commented):
@zanhsieh Thank you for your review!

- Regarding CODEOWNERS, I made the edits but may need to address the error about I don't have writer access to the repo separately.
- Regarding testing, I added a connection test that mimics prometheus-statsd-exporter. I would like to add more tests in the future similar to the Grafana example you shared.

zanhsieh:
@monotek 
Could you comment on this PR please?

monotek (changes requested):
Imho this is not needed.

There are already some aspirations to remove the dependencies from the exsiting prometheus chart. 

See for example: https://github.com/prometheus-community/helm-charts/pull/2673

I would prefer to move out the rest there too and keep a single chart for it.

@ivyostosh 
I guess some help would be welcome, so you might want to add a pr to replace alertmanager service with altertmanager helm chart in the prometheus chart for example?

zanac1986:
> @monotek Make sense. If the main prometheus chart can remove the internal dependencies in the near term, I agree it will be redundant to create this standalone chart. I'm happy to take on the alertmanager task as you suggested.
> 
> Also just so I get a sense on the timeline, is there an estimated target date to complete the entire refactoring work for the main prometheus chart?

@ivyostosh I used to think splitting sub-charts in parallel should be no problem, but while soon gave it up once started that idea. Have to admit I am not a long time maintainer of all sub-charts. So currently have to do it one-by-one. I would roughly expect the beginning  of Feb 2023 to get all sub-charts split if we count in all holidays.

ivyostosh:
> @ivyostosh I used to think splitting sub-charts in parallel should be no problem, but while soon gave it up once started that idea. Have to admit I am not a long time maintainer of all sub-charts. So currently have to do it one-by-one. I would roughly expect the beginning of Feb 2023 to get all sub-charts split if we count in all holidays.

@zanac1986 Got it. Thanks for the clarification! Regardless, I'm happy to contribute to the refactoring work, and I will start with the alert-manager task as @monotek suggested.

zanac1986:
> > @ivyostosh I used to think splitting sub-charts in parallel should be no problem, but while soon gave it up once started that idea. Have to admit I am not a long time maintainer of all sub-charts. So currently have to do it one-by-one. I would roughly expect the beginning of Feb 2023 to get all sub-charts split if we count in all holidays.
> 
> @zanac1986 Got it. Thanks for the clarification! Regardless, I'm happy to contribute to the refactoring work, and I will start with the alert-manager task as @monotek suggested.

@ivyostosh Following is the workflow I migrated from the embedded prometheus-node-exporter to its origin. You might already found some more efficient / better ways than mime.

1. Dump the embedded chart yaml. For example,

```bash
for y in $(ls templates/alertmanager/*.yaml); do helm template . -s $y; done > ~/Downloads/old_alertmanager.yaml
```

2. Change Charts.yaml and append dependencies
3. In values.yaml, append `old_` prefix in the old embedded chart referred section, and create new section as the new parent chart.
4. Copy the current origin chart values.yaml configuration and drop to the new section with proper indention. Make sure the new section contains an field as `enabled: true`
5. Rename the old embedded chart folder with `old_` prefix.
6. Run `helm dependency build`
7. Run 

```bash
for y in $(ls charts/alertmanager/*.yaml); do helm template . -s $y; done > ~/Downloads/new_alertmanager.yaml
```

8. Use comparison tools (e.g. `vimdiff -O ~/Downloads/{old,new}_alertmanager`) to see the difference and match the configure value in `prometheus/values.yaml` respectively.

ivyostosh:
> @ivyostosh Following is the workflow I migrated from the embedded prometheus-node-exporter to its origin. You might already found some more efficient / better ways than mime.
> 
> 1. Dump the embedded chart yaml. For example,
> 
> ```shell
> for y in $(ls templates/alertmanager/*.yaml); do helm template . -s $y; done > ~/Downloads/old_alertmanager.yaml
> ```
> 
> 2. Change Charts.yaml and append dependencies
> 3. In values.yaml, append `old_` prefix in the old embedded chart referred section, and create new section as the new parent chart.
> 4. Copy the current origin chart values.yaml configuration and drop to the new section with proper indention. Make sure the new section contains an field as `enabled: true`
> 5. Rename the old embedded chart folder with `old_` prefix.
> 6. Run `helm dependency build`
> 7. Run
> 
> ```shell
> for y in $(tar -tf ./charts/prometheus-node-exporter-x.x.x.tgz | grep templates | grep yaml); do echo helm template . -s charts/$y ; done > ~/Downloads/new_alertmanager.yaml
> ```
> 
> 8. Use comparison tools (e.g. `vimdiff -O ~/Downloads/{old,new}_alertmanager`) to see the difference and match the configure value in `prometheus/values.yaml` respectively.

@zanac1986 Thanks! I will make sure to use this as a reference.

PR #2694, opened 2022-11-16
[prometheus-elasticsearch-exporter] Refactor securityContext configuration

### LINKED ISSUE #2494: [prometheus-elasticsearch-exporter]  Container security context capabilities

When namespace is protected by [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) Pod Security the containers must drop `ALL` capabilities.
Is there a reason why not all Linux capabilities are dropped?

https://github.com/prometheus-community/helm-charts/blob/56fc02fe4dafa39f8118c28f8d2f4a55fed8930a/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml#L128-L144

Drop `ALL` Linux capabilities.
```
securityContext:
  capabilities:
    drop:
      - ALL
```
None

### Additional context.
No response

### LINKED ISSUE #2617: [prometheus-elasticsearch-exporter] Make securityContext for pod and container configurable

To be able to use the ["restricted" level of Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) it is necessary to explicitly configure a couple of settings. Basically it boils down to:

```yaml
podSecurityContext:
  runAsNonRoot: true
  seccompProfile:
    type: "RuntimeDefault"

securityContext:
  capabilities:
    drop:
    - ALL
```

Also it is necessary to start the container with a specific UID instead of a named user to ensure root is not used. This is already covered by the current implementation which by default uses UID 1000.

The issue #2494 is closely related.

My suggestion would be to move the current `securityContext` settings for the [pod](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml#L52) and [container](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml#L128) from deployment.yaml to values.yaml as podSecurityContext and securityContext respectively and just render it’s values verbatim. This mirrors the behaviour of basic charts created by `helm create` and should be mostly backwards compatible while allowing users to overwrite settings to their preference.

The only gap in this solution is `securityContext.enabled` since it also currently refers to what would then be the podSecurityContext. But it’s misleading anyway since the container-level securityContext does not respect it anyway at the moment.

Kustomize may be able to circumvent the issue but with the switch from PSPs to PSS this really should be possible to achieve just by using custom values.

### Additional context.

_No response_

manzsolutions-lpr:
To allow the use of configuration conforming to the Pod Security Standard level "restricted" it is necessary to be able to adapt the securityContext on pod and container level. Since the implementation up to now is not very flexible and a bit misleading (securityContext.enabled only affects the pod level securityContext) the approach taken is to align it with the way of charts created by `helm create` implement this. 

- fixes #2617
- fixes #2494 

Unfortunately aligning it with the behaviour of other charts implies a breaking change as "securityContext" of 4.x is renamed to "podSecurityContext" and "securityContext" is used for the container level "securityContext" going forward. "securityContext.enabled" is removed altogether. To remove the configuration from the pod users should set `podSecurityContext: null` instead. Additionally it is now possible to also remove the container level configuration by setting `securityContext: null`.

Last but not least `seccompProfile` and `allowPrivilegeEscalation` settings were added to the default configuration and all capabilities are dropped to comply with the PSS level `restricted` without further intervention.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
+++ b/charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
@@ -1,5 +1,5 @@
 ---
-# Set default security context for kubernetes
+# Set default security context for Openshift
 
-securityContext:
-  disable: true
+podSecurityContext:
+  runAsUser: null
--- a/charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
+++ b/charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
@@ -1,5 +1,4 @@
 ---
-# Set default security context for Openshift
+# Unset pod level securityContext
 
-podSecurityContext:
-  runAsUser: null
+podSecurityContext: null
--- a/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
@@ -48,10 +48,9 @@ spec:
 {{- end }}
       {{- include "elasticsearch-exporter.image.pullSecret.name" (dict "images" (list .Values.image) "context" $) | nindent 6 }}
       restartPolicy: {{ .Values.restartPolicy }}
-      {{- if .Values.securityContext.enabled }}
+      {{- with .Values.podSecurityContext }}
       securityContext:
-        runAsNonRoot: true
-        runAsUser: {{ .Values.securityContext.runAsUser }}
+        {{- toYaml . | nindent 8 }}
       {{- end }}
       {{- with .Values.dnsConfig }}
       dnsConfig:
@@ -125,24 +124,10 @@ spec:
                     {{- end }}
                     "--web.listen-address=:{{ .Values.service.httpPort }}",
                     "--web.telemetry-path={{ .Values.web.path }}"]
+          {{- with .Values.securityContext }}
           securityContext:
-            capabilities:
-              drop:
-                - SETPCAP
-                - MKNOD
-                - AUDIT_WRITE
-                - CHOWN
-                - NET_RAW
-                - DAC_OVERRIDE
-                - FOWNER
-                - FSETID
-                - KILL
-                - SETGID
-                - SETUID
-                - NET_BIND_SERVICE
-                - SYS_CHROOT
-                - SETFCAP
-            readOnlyRootFilesystem: true
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- with .Values.resources }}
           resources:
             {{- toYaml . | nindent 12 }}
--- a/charts/prometheus-elasticsearch-exporter/values.yaml
+++ b/charts/prometheus-elasticsearch-exporter/values.yaml
@@ -18,19 +18,12 @@ image:
   pullPolicy: IfNotPresent
   pullSecret: ""
 
-## Set to `null` to keep pod level securityContext unspecified
-#podSecurityContext: null
 podSecurityContext:
   runAsNonRoot: true
-  ## Openshift won't deploy with runAsUser: 1000 without additional permissions.
-  ## Remove the field by setting it to `null`:
-  #runAsUser: null
   runAsUser: 1000
   seccompProfile:
     type: "RuntimeDefault"
 
-## Set to `null` to keep container level securityContext unspecified
-#securityContext: null
 securityContext:
   allowPrivilegeEscalation: false
   capabilities:
--- a/charts/prometheus-elasticsearch-exporter/values.yaml
+++ b/charts/prometheus-elasticsearch-exporter/values.yaml
@@ -18,13 +18,25 @@ image:
   pullPolicy: IfNotPresent
   pullSecret: ""
 
-## Set enabled to false if you don't want securityContext
-## in your Deployment.
-## The below values are the default for kubernetes.
-## Openshift won't deploy with runAsUser: 1000 without additional permissions.
-securityContext:
-  enabled: true  # Should be set to false when running on OpenShift
+## Set to `null` to keep pod level securityContext unspecified
+#podSecurityContext: null
+podSecurityContext:
+  runAsNonRoot: true
+  ## Openshift won't deploy with runAsUser: 1000 without additional permissions.
+  ## Remove the field by setting it to `null`:
+  #runAsUser: null
   runAsUser: 1000
+  seccompProfile:
+    type: "RuntimeDefault"
+
+## Set to `null` to keep container level securityContext unspecified
+#securityContext: null
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
+  readOnlyRootFilesystem: true
 
 # Custom DNS configuration to be added to prometheus-elasticsearch-exporter pods
 dnsConfig: {}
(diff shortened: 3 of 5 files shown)


charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-# Set default security context for kubernetes
+# Set default security context for Openshift
```
zanhsieh, line 2: Why change to support OpenShift only?

zanhsieh (changes requested):

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -18,13 +18,25 @@ image:
...
-## in your Deployment.
-## The below values are the default for kubernetes.
-## Openshift won't deploy with runAsUser: 1000 without additional permissions.
-securityContext:
-  enabled: true  # Should be set to false when running on OpenShift
+## Set to `null` to keep pod level securityContext unspecified
+#podSecurityContext: null
+podSecurityContext:
+  runAsNonRoot: true
+  ## Openshift won't deploy with runAsUser: 1000 without additional permissions.
```
monotek, line 25: please remove openshift specific comments.

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -18,13 +18,25 @@ image:
...
+  ## Remove the field by setting it to `null`:
+  #runAsUser: null
```
monotek, line 27: why adding it at all if its disabled?
imho only the current defaults should be used again.

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -18,13 +18,25 @@ image:
...
   runAsUser: 1000
+  seccompProfile:
+    type: "RuntimeDefault"
+
+## Set to `null` to keep container level securityContext unspecified
```
monotek, line 32: please remove that comment.

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -18,13 +18,25 @@ image:
   pullPolicy: IfNotPresent
   pullSecret: ""
 
-## Set enabled to false if you don't want securityContext
-## in your Deployment.
-## The below values are the default for kubernetes.
-## Openshift won't deploy with runAsUser: 1000 without additional permissions.
-securityContext:
-  enabled: true  # Should be set to false when running on OpenShift
+## Set to `null` to keep pod level securityContext unspecified
```
monotek, line 21: please remove that comment

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -18,13 +18,25 @@ image:
...
+#podSecurityContext: null
+podSecurityContext:
```
monotek, line 23: please use it like this to use current default.
```suggestion
podSecurityContext: {}
```

monotek (changes requested):

charts/prometheus-elasticsearch-exporter/values.yaml
manzsolutions-lpr: This was meant to demonstrate how to unset the field, e.g. that there must not be quotes around the `null`.

charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
manzsolutions-lpr: Hopefully my adaption resolves the confusion and at the same time better mirrors the effect the previous version of this file had.

charts/prometheus-elasticsearch-exporter/values.yaml
monotek: to unset your normal just use it like:
```
podSecurityContext: {}
```

charts/prometheus-elasticsearch-exporter/values.yaml
monotek: but curretnly podSecurityContext is not used and so you would keep the defaults and still have it configurable.

charts/prometheus-elasticsearch-exporter/values.yaml
manzsolutions-lpr: I thought so, too, and tested it but this produces the following for me:

```
      securityContext:
        - ""
```

charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
```diff
@@ -48,10 +48,9 @@ spec:
 {{- end }}
       {{- include "elasticsearch-exporter.image.pullSecret.name" (dict "images" (list .Values.image) "context" $) | nindent 6 }}
       restartPolicy: {{ .Values.restartPolicy }}
-      {{- if .Values.securityContext.enabled }}
+      {{- with .Values.podSecurityContext }}
       securityContext:
-        runAsNonRoot: true
-        runAsUser: {{ .Values.securityContext.runAsUser }}
+        {{- toYaml . | nindent 8 }}
```
manzsolutions-lpr, line 53: @monotek: The pod level SecurityContext is set in the currently released Helm chart, I’m merely moving it from hardcoded in the deployment to "podSecurityContext" in values.yaml.

charts/prometheus-elasticsearch-exporter/values.yaml
manzsolutions-lpr: I since removed the comment but I’m unsure whether you or I should mark this discussion as resolved. I’ll go ahead and do it myself given the trivial nature of the requested change but please let me know if it is supposed to be the reviewer marking it as resolved.

monotek (approved):

monotek (changes requested):
Please fix linting errors.

manzsolutions-lpr:
> Please fix linting errors.

I did my best. Unfortunately one "natural language" one is at odds with the `helm repo` command so I restructured the sentence to hopefully circumvent the failing check.

monotek:
you can have a look at the kube-prometheus-stack readme for example and copy the needed fixes from there.

charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
```diff
@@ -1,5 +1,4 @@
 ---
-# Set default security context for kubernetes
+# Unset pod level securityContext
 
-securityContext:
-  disable: true
+podSecurityContext: null
```
monotek, line 4:
```suggestion
podSecurityContext: {}
```

monotek (changes requested):

charts/prometheus-elasticsearch-exporter/ci/security-context.yaml
manzsolutions-lpr: This does not produce the desired outcome (securityContext to be removed from the pod level), at least with Helm v3.10.2 on my laptop. Helm “merges“ the empty dict with the dict from `values.yaml` and the end result is thus the dict from values.yaml. The idea here is to completely remove the securityContext and this only works by setting it to "null" causing Helm to unset the value of "podSecurityContext" instead of merging it.

monotek (approved):

PR #2696, opened 2022-11-16
[prometheus-node-exporter] Update image to v1.4.0

lukas-unity:

Node Exporter has released v1.4.0 this PR updates the base image in the chart.  

#### Checklist

- [X] Chart Version bumped


monotek (changes requested):

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,8 +6,8 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.5.2
-appVersion: 1.3.1
```
monotek, line 10: to use the app version you would need to adjust the deamonset too.
see: https://github.com/prometheus-community/helm-charts/blob/d3bccc205efd0a4dfa7fccc4cda25cf28272c6c6/charts/prometheus-node-exporter/templates/daemonset.yaml#L40

monotek (changes requested):

charts/prometheus-node-exporter/Chart.yaml
lukas-unity: Thank you, I think I looked at the wrong chart when I changed the `appVersion` 😅

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,8 +6,8 @@ keywords:
...
+version: 4.5.3
```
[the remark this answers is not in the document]
lukas-unity: Updated

charts/prometheus-node-exporter/Chart.yaml
lukas-unity: It seems to be already defined here https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/templates/_helpers.tpl#L75-L83

zanhsieh (approved):

zanhsieh (approved):

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,8 +6,8 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.6.0
-appVersion: 1.3.1
+version: 4.6.1
```
monotek, line 9: imho chart minor version should be raised, if app minor version has changed

monotek (changes requested):

charts/prometheus-node-exporter/Chart.yaml
lukas-unity: So you think we should bump it to 4.7.0 ?

monotek: yes

lukas-unity: done

monotek (approved):

Issue #2612, opened 2022-10-26
New chart proposition: kube-prometheus-stack-crds


desaintmartin:
### Is your feature request related to a problem ?

- We have kube-prometheus-stack chart with prometheus-operator CRDs
- In the [helm documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts), it is suggested to have a separate chart for CRD management. This is especially useful to automate upgrades of CRDs.

### Describe the solution you'd like.

- I propose to create a new `kube-prometheus-stack-crds` chart + related update script which only contains the CRDs. There are already three charts doing this:
  - ours: https://artifacthub.io/packages/helm/wiremind/kube-prometheus-stack-crds
  - https://artifacthub.io/packages/helm/morremeyer/kube-prometheus-stack-crds
  - and one based on prometheus-operator: https://artifacthub.io/packages/helm/appscode/prometheus-operator-crds

Would it be acceptable?

### Describe alternatives you've considered.

Maintaining it outside of prometheus-community.

desaintmartin:
cc @morremeyer for additional discussion since you automated upgrades of the CRDs.

monotek:
Will you maintain it?

morremeyer:
@ekeih and I have created https://github.com/morremeyer/charts/tree/main/charts/kube-prometheus-stack-crds which is updated 100% automated.

This is achieved using:

* The [renovate](https://github.com/renovatebot/renovate) GitHub App do automatically trigger updates
* A workflow that gets the charts from this repository and updates the templates directory in the *kube-prometheus-stack-crds* chart
* The @morremeyer-bot to commit back to the repository (necessary in morremeyer/charts to trigger all pipelines since pushes with the GitHub actions' default token don't trigger other actions)

As I use kube-prometheus-stack in all my setups, I will maintain this for the for forseeable future.

If there is an interest in integrating that into this repository, I'm happy to discuss it - it could make maintenance easier since both charts could probably be released at the same time.

jcpunk:
I'd also be interested in this chart.  It would make it much easier to install cilium via helm.

The current workflow:

1. install cilium
2. install kube-prometheus
3. enable cilium service monitor

Since you cannot enable the serviceMonitor without the CRDs and you cannot install kube-prometheus via helm without a working CNI, this gets a bit ugly.

[...]

PR #2697, opened 2022-11-16
[prometheus-crds]: CRDs provided in separate Chart

dacamposol:

Using `kube-prometheus-stack` from GitOps tools like Argo CD poses a problem in the way that currently the CRDs are defined to be used.

Since there are some CRDs (like `prometheuses`) which have a lot of changes between versions, the annotations exceed the maximum size, ending up in synchronise issues which can only be solved manually (for example by forcing Argo CD to `replace` that resource instead of `applying` changes to it). This behaviour can be modified at resource level, but that would require to be able to modify the annotations of the proper CRDs... Which we can not when they are under the `crd` folder.

Due to that, I'm presenting this PR, where we introduce the [CRDs as a separate Chart](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts) and also I'm allowing to introduce some custom annotations for compatibility with GitOps tools which can rely on those for adding custom logic on how those resources need to be applied.

For example, I would desire to have the following annotation on my CRDs for ArgoCD:

```yaml
argocd.argoproj.io/sync-options: Replace=true
```

Also, this won't break compatibility of the already existing `kube-prometheus-stack`, since it contains the CRDs in the `crd` folder (according to Helm 3 specification), so we can just avoid to install those with the `skipCRDs` flag present in almost all the GitOps tools which rely on Helm, and include the CRDs by this new Chart.

This issue has been discussed before, but it was closed without an actual solution moreover than recommending to apply the CRDs manually through `kubectl create` or `kubectl apply --force-conflicts=true --server-side`, which is not an actual solution but more a patch.

This is just a proposal, so I acknowledge that some changes should be made before it's being accepted to be included.

#### Checklist

- [X] Title of the PR starts with chart name (new Chart)

Signed-off-by: Daniel Campos Olivares <dacamposol@gmail.com>


mrueg:
I would suggest to rename the chart to prometheus-operator-crds as prometheus can run without those.

dacamposol:
> I would suggest to rename the chart to prometheus-operator-crds as prometheus can run without those.

Fair enough 😄

monotek (changes requested):
Maintainers are missing.
@desaintmartin do you want to co maintain?

desaintmartin:
Indeed, I am interested! It would be even more awesome to have it automated somehow, discussing in https://github.com/prometheus-community/helm-charts/issues/2612.

desaintmartin:
There are still linting errors @dacamposol.

charts/prometheus-operator-crds/Chart.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v2
+type: application
+version: 0.1.0
+name: prometheus-operator-crds
+icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/assets/prometheus_logo-cb55bb5c346.png
+description: |
+  A Helm chart that collects custom resource definitions (CRDs) from the Prometheus Operator, allowing for seamless
```
jcpunk, line 7: Resolving the trailing spaces here might cleanup the linting error.

dacamposol:
> There are still linting errors @dacamposol.

Resolved. 

Could you please trigger the tests?

desaintmartin:
Looks good! Should we merge this? And my first contribution will be to update them. :)

desaintmartin (approved):

monotek (approved):

PR #2714, opened 2022-11-22
[prometheus] moving pushgateway out of prometheus 

zanac1986:
#### What this PR does / why we need it

Move node-exporter the embedded chart back to its parent. As mentioned here:

https://github.com/prometheus-community/helm-charts/pull/2673
https://github.com/prometheus-community/helm-charts/pull/2680
https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300167738
https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300192978
https://github.com/prometheus-community/helm-charts/issues/2654#issuecomment-1306468145

Signed-off-by: zanac1986 <zanhsieh@protonmail.com>

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #2627, opened 2022-11-01
[prometheus] major refactor

zanac1986:
#### What this PR does / why we need it

Major refactoring several syntax and fixing mis-alignment across chart components

#### Which issue this PR fixes

- Empty values (e.g. `[]`, `{}`) shall omit.
```
# from
{{- if .resources }}
resources:
  {{- toYaml .resource | nindent 2 }}
{{- end }}
# to
{{- with .resources }}
resources:
  {{- toYaml . | nindent 2 }}
{{- end }}
```
- Misuse truncate empty spaces + indention.
```
# from
metadata:
  ...
  name: foo
{{ include "foo.namespace" $ | indent 2 }}
...
# to
metadata:
  ...
  name: foo
  {{- include "foo.namespace" $ | nindent 2 }}
...
```
- Use scope. Note that we have to change all `include` and `template` as root based (`.` -> `$`).
```
# from
apiVersion: v1
kind: Service
metadata:
{{- if .Values.foo.a.super.super.long.miles.away.annotations }}
  annotations:
{{ toYaml .Values.foo.a.super.super.long.miles.away.annotations | indent 4 }}
{{- end }}
  labels:
    {{- include "foo.labels" . | nindent 4 }}
{{- if .Values.foo.a.super.super.long.miles.away.labels }}
{{ toYaml .Values.foo.a.super.super.long.miles.away.labels | indent 4 }}
{{- end }}
  name: {{ template "foo.fullname" . }}-headless
{{ include "foo.namespace" . | indent 2 }}
spec:
  clusterIP: None
  ports:
    - name: http
      port: {{ .Values.foo.a.super.super.long.miles.away.servicePort }}
      protocol: TCP
      targetPort: 9093
{{- if .Values.foo.a.super.super.long.miles.away.enableMeshPeer }}
    - name: meshpeer
      port: 6783
      protocol: TCP
      targetPort: 6783
{{- end }}
  selector:
    {{- include "foo.matchLabels" . | nindent 4 }}
{{- end }}

# to
{{- with .Values.foo.a.super.super.long.miles.away }}
apiVersion: v1
kind: Service
metadata:
  {{- with .annotations }}
  annotations:
    {{ toYaml . | nindent 4 }}
  {{- end }}
  labels:
    {{- include "foo.labels" $ | nindent 4 }}
    {{- with .labels }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
  name: {{ template "foo.fullname" $ }}-headless
[...]

PR #2717, opened 2022-11-23
[prometheus] move alertmanager out of prometheus

zanac1986:

Move node-exporter the embedded chart back to its parent. As mentioned here:

https://github.com/prometheus-community/helm-charts/pull/2714
https://github.com/prometheus-community/helm-charts/pull/2673
https://github.com/prometheus-community/helm-charts/pull/2680
https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300167738
https://github.com/prometheus-community/helm-charts/pull/2627#issuecomment-1300192978
https://github.com/prometheus-community/helm-charts/issues/2654#issuecomment-1306468145

#### Checklist

- [x] Chart Version bumped


charts/prometheus/values.yaml
```diff
@@ -15,7 +15,7 @@ serviceAccounts:
     name:
     annotations: {}
 
-alertmanager:
+old_alertmanager:
```
monotek: Variablenames sould be camel case.

Why is this needed anyway?

The condition in chart.yaml is "alertmanager.enabled".

monotek (changes requested):

charts/prometheus/values.yaml
zanac1986: @monotek 
There must be some glitch on Github since I can't find `old_alertmanager` in values.yaml.

zanac1986:
For the reason why required Helm 3.7+, see this issue:

https://github.com/prometheus-community/helm-charts/issues/2723

monotek (approved):
lgtm.

not sure if state / pvc of alertmanager should be migrated too though?

zanhsieh:
> not sure if state / pvc of alertmanager should be migrated too though?
@monotek 
Any tip on how to migrate pvc underneath?

monotek:
> Any tip on how to migrate pvc underneath without data loss?

Could be done via manual steps (kubectl copy from old pvc to local dir and copy from local dir to new pvc).

I started to write a chart for such stuff (with move of prometheus deployment to statefulset in mind) but i need some more time to finish and test it, as i'm occupied with other stuff at the moment.

zanac1986:
> > Any tip on how to migrate pvc underneath without data loss?
> 
> Could be done via manual steps (kubectl copy from old pvc to local dir and copy from local dir to new pvc) but this might mess up permissions.
> 
> I started to write a chart for such stuff (with move of prometheus deployment to statefulset in mind) but i need some more time to finish and test it, as i'm occupied with other stuff at the moment.

Glad to know. Please let me know when I should use it.

monotek:
Like i said.... Don't expect it soon :/

zanhsieh (approved):

ivyostosh:
@zanac1986 

Sorry I got behind on submitting my PR to move out the alertmanager! I was under the impression some work needs to be done on the alertmanager chart side before doing this change, but it seems to be a misunderstanding on my side. Either way, I would like to ask the following clarification:

I noticed the following templates exist in the old prometheus alertmanager charts, but not in the new alertmanager charts you referenced in this PR:
clusterrole
clusterrolebinding
deploy
headless-svc
netpol
psp
pvc
role
rolebinding

So my questions are twofold:
- What will happen during an upgrade for those templates that don't exist anymore?
- Is there a plan to bring those components back to the alertmanager chart?

zanac1986:
@ivyostosh 
The answer would be somewhat complicated.

Q: What will happen during an upgrade for those templates that don't exist anymore?
A: I think they should get deleted. You could give it a try. So far I didn't see any incompatibility issue yet.

```
helm install prometheus prometheus-community/prometheus --version 15.18.0
helm upgrade prometheus prometheus-community/prometheus --version 18.0.0.0
```

Q: Is there a plan to bring those components back to the alertmanager chart?
A: Would try to bring them back upon requests. For those obsolete (e.g. psp) will get drop. For PVC, I didn't find any good way for migration.

monotek:
@zanac1986 

can you please also check whats the issue after the upgrade here: https://github.com/prometheus-community/helm-charts/issues/2745

PR #2722, opened 2022-11-23
change default strategy to recreate for alertmanager and server

### LINKED ISSUE #421: [prometheus-community/prometheus] strategy: type: leads to unhealthy deployment.
We use prometheus-community/prometheus chart to deploy Prometheus server, and it works with persistent volume for the purpose of historical data to survive pod replacement.

After every small config change applied (e.g. retention time 1d->5d), Prometheus deployment renders unhealthy: There are 2 pods and latter restarts continuously. This is because the  update strategy is first launch new, then terminate old, while with persistent volume having 2 pods accessing same volume is impossible.

The solution desired:
 to be able to set in manifest
```
strategy:
  type: Recreate
```

I think that it applies as well to other components which use persistent volume.

mcallaghan-geotab:
Signed-off-by: mcallaghan-geotab <78922178+mcallaghan-geotab@users.noreply.github.com>

By default both `prometheus-alertmanager` and `prometheus-server` are using persistent volumes, therefore we would want the default deployment strategy to be `Recreate` rather than `RollingUpdate`.

See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy

The problem with PVC (persistent volumes) and using rolling update, is that during a deployment upgrade (`helm upgrade ...`) when new pods come online they cannot claim the mount because the previous/existing (older) pod is holding the claim.

To be honest, I'm not sure why the prometheus stack/chart has been this way for so long? (anyone attempting to upgrade the stack over time while using persistent volumes would run into this problem)

The WORKAROUND (fwiw), without this fix, is to run the `helm upgrade ...` command, and then scale the deployment down to `0` and back up to `1`.

FIRST INTENT:  is to propose the idea (if maintainers agree) - then we can bump chart version, update docs if needed, etc.

- fixes #421

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -157,8 +157,8 @@ alertmanager:
     #       - alertmanager.domain.com
 
   ## Alertmanager Deployment Strategy type
-  # strategy:
-  #   type: Recreate
+  strategy:
+    type: Recreate
 
   ## Node tolerations for alertmanager scheduling to nodes with taints
   ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
@@ -1011,8 +1011,8 @@ server:
     #       - prometheus.domain.com
 
   ## Server Deployment Strategy type
-  # strategy:
-  #   type: Recreate
+  strategy:
+    type: Recreate
 
   ## hostAliases allows adding entries to /etc/hosts inside the containers
   hostAliases: []


monotek (changes requested):
Chart version needs to be raised.

I rather would change the Prometheus deployment to  be a statefulset instead by default. Needs a rather complicated upgradepath though, but I already have something in mind.

The alert manager will be moved out of the chart soon too. See: https://github.com/prometheus-community/helm-charts/pull/2717

Alert manager chart uses statefulset anyway.

mcallaghan-geotab:
OK Thanks @monotek - if the intent is to change these anyway then i'll reject my own PR (not worth addressing)

Issue #2708, opened 2022-11-18
[kube-prometheus-stack] Relabelings Assuming Action


JuanPabloSGU:
### Describe the bug a clear and concise description of what the bug is.

It seems that 

```YAML
relabelings:
  - action: replace
```

is being detected by Argocd. 

### What's your helm version?

version.BuildInfo{Version:"v3.10.0", GitCommit:"ce66412a723e4d89555dc67217607c6579ffcb21", GitTreeState:"clean", GoVersion:"go1.18.6"}

### What's your kubectl version?

Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-12T10:57:26Z", GoVersio n:"go1.19.2", Compiler:"gc", Platform:"linux/amd64"} Kustomize Version: v4.5.7

### Which chart?

kube-prometheus-stack

### What's the chart version?

v41.7.4

### What happened?

I have updated the `kube-prometheus-stack` version from 39.0.1 to 41.7.4 and going from client side apply to server side apply.  When updating Argocd had syncing issues with the recent updated chart. It specified that `kube-prometheus-stack-kubelet (service monitor)` was out of sync. 

Desired Manifest 
```YAML
  relabelings:
    - sourceLabels:
      - __metrics_path__
        targetLabel: metrics_path
```

Life Manifest
```YAML
relabelings:
  - action: replace
    - sourceLabels:
      - __metrics_path__
        targetLabel: metrics_path
```

Argocd noticed the difference and would be unable to sync with the new version of the chart. 

### What you expected to happen?

I've followed all of the chart upgrade steps to a newer version of the chart and would have expected it to not run into too many issues. 

### How to reproduce it?

The solution that I had to do is to provide an ignoreDifferences to the deployment within Argocd to allow the new chart to sync. Without this ignoreDifference the application would be out of sync and Argocd would keep trying to delete the missing "replace" and "replace" would be added back.  

### Enter the changed values of values.yaml?

NONE

### Enter the command that you execute and failing/misfunctioning.

NA - Argocd handles it

cilindrox:
Running into the same issue (and workaround), here's the constant drift I'm seeing:

```diff
# monitoring.coreos.com/ServiceMonitor/monitoring/kube-prometheus-stack-kubelet
  port: https-metrics
  relabelings:
-   - action: replace
[...]

PR #2732, opened 2022-11-24
[kube-prometheus-stack] Prevent helm diff on ServiceMonitor - Kubelet

Evilong:

Prevent helm diff on apply of the chart
Issue when using tools like ArgoCD

This is what is shown on a tool like ArgoCD which declares that the installation of the chart is out of sync.
Screenshot 2022-11-24 at 17 01 45

Changes tested by diffing results with default values

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,7 +8,7 @@
 
 /charts/alertmanager/ @monotek @naseemkullah
 /charts/jiralert/ @jkroepke @zanhsieh
-/charts/kube-prometheus-stack/ @andrewgkew @gianrubio @gkarthiks @GMartinez-Sisti @scottrigby @Xtigyro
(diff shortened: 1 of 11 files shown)


monotek (changes requested):
What should that diff look like?

"action: replace" is the default so it's used anyway.

Chart version needs to be raised.

Evilong:
Added screenshot to explain the issue on the first message. Also bumped chart version.

Evilong:
Adding on the issue, if you apply a ServiceMonitor manifest like the one generated before my changes, the final resource on the cluster will be mutated and add the same as I did on the PR

jemag:
Similar PR was added around the same time https://github.com/prometheus-community/helm-charts/pull/2741. 
Both would Fix https://github.com/prometheus-community/helm-charts/issues/2708.

Xtigyro (approved):
Thank you, @Evilong!

PR #2794, opened 2022-12-07
[kube-prometheus-stack] operator tls disable -> admission client svc port 8080

joeunlog:
Signed-off-by: joeunlog <j.eunvit@gmail.com>

When Setting up prometheus-operator not using tls, admission-webhooks client port set 8080.
Because webhook configuration client service port's default is 443.

- No issue

#### Checklist

- [ ] Chart Version bumped


zeritti:
Not a reviewer myself but still expressing an opinion. When disabling TLS at the operator, the webhooks won't effectively be used as the API server requires TLS when communicating with them, i.e. it always uses the https scheme. The default service port (443/tcp) in WebhookClientConfig is not a coincidence.

Since the operator itself implements the webhooks, it must support TLS if the webhooks are to be enabled.

PR #2822, opened 2022-12-16
[kube-prometheus-stack] Fix metrics for prometheus-adapter

kwohlfahrt:

`prometheus-adapter` queries for cgroup metrics that have the ID '/' when responding to `kubectl top node`. These metrics don't have a pod set, so are dropped by the current config.

Previously, the regex would match `.+;` - i.e any non-empty ID, and no pod. Now, it matches (if I've done this right) any ID except '/', and no pod.

- fixes: kubernetes-sigs/prometheus-adapter#549

- I'm no regex expert, there may be more efficient ways to achieve this. 
- Do  I need to bump the chart major version for this? All metrics that were previously collected still are, this just adds a few more.
- @SuperQ, who originally configured the metrics to drop might be well placed to review this.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 43.1.1
+version: 43.1.2
 appVersion: 0.61.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1002,10 +1002,10 @@ kubelet:
       - sourceLabels: [__name__]
         action: drop
         regex: 'container_spec.*'
-      # Drop cgroup metrics with no pod.
+      # Drop cgroup metrics with no pod, where the id is not exactly '/' (used by prometheus-adapter for `kubectl top node`)
       - sourceLabels: [id, pod]
         action: drop
-        regex: '.+;'
+        regex: "(/.+|[^/;]+|);"
     # - sourceLabels: [__name__, image]
     #   separator: ;
     #   regex: container_([a-z_]+);


SuperQ:
I think we already determined in another issue that the fix needs to be on the adapter side. To use the node_exporter metrics and not cgroups.

dgrisonnet:
Since this is the chart used by the kube-prometheus-stack are there any reasons that prevent you to reuse the configuration from kube-prometheus which uses node-exporter metrics instead of the one from upstream prometheus-adapter ?

SuperQ:
I'll try and take a look at the prometheus-adapter queries this weekend.

kwohlfahrt:
Thanks for the replies - yes, I'm happy to use different metrics if these are more suitable! In that case, I think this is a documentation issue, [the README](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-adapter#resource-metrics) and [values.yaml](https://github.com/prometheus-community/helm-charts/blob/8d94a176a682977daee27b20444b9b9c363fe42c/charts/prometheus-adapter/values.yaml#L112-L137) for `prometheus-adapter` in this repo seems to use the cgroup metrics for `nodeQuery` (the same as upstream) - or am I missing something?

```
rules:
  resource:
    cpu:
      containerQuery: sum(rate(container_cpu_usage_seconds_total{<<.LabelMatchers>>, container!=""}[3m])) by (<<.GroupBy>>)
      nodeQuery: sum(rate(container_cpu_usage_seconds_total{<<.LabelMatchers>>, id='/'}[3m])) by (<<.GroupBy>>)
      resources: ...
    memory:
      containerQuery: sum(container_memory_working_set_bytes{<<.LabelMatchers>>, container!=""}) by (<<.GroupBy>>)
      nodeQuery: sum(container_memory_working_set_bytes{<<.LabelMatchers>>,id='/'}) by (<<.GroupBy>>)
      resources: ...
    window: 3m
```

kwohlfahrt:
This PR is probably not the right approach, and can be closed then.

SuperQ:
Thanks. It seems like that is a "how many CPUs are in use" query. I would probably use something like this:

```
nodeQuery: sum by (<<.GroupBy>>) (rate(node_cpu_seconds_total{mode!="idle",mode!="iowait",mode!="steal",<<.LabelMatchers>>}[3m]))
```

PR #2845, opened 2022-12-23
[prometheus-adapter] Support environment variables (#2842)

### LINKED ISSUE #2842: [prometheus-adapter] Add option to add environment variables to the prometheus-adapter container

We use a hosted prometheus in grafana cloud and want to connect to it with basic authentication.
This is possible through passing the[ --prometheus-header](https://github.com/kubernetes-sigs/prometheus-adapter/issues/439#issuecomment-896730036) flag to the adapter,  which can be achieved by setting the extraArguments variable of the values.yaml 

The problem is that there is no way to inject an environment variable to the container

This means that i cannot use gitops to deploy this chart as i need to have the password hardcoded in the values.yaml

Have an option to inject environment variables from secrets to this container and use them as extra arguments

I considered to create a kubeconfig file which contains the username and password

```
apiVersion: v1
clusters:
- name: prom
  cluster:
    server: http://hosted-prometheus.com:9090
contexts:
- name: prom-context
  context:
    cluster: prom
    user: prom-user
current-context: prom-context
users:
- name: prom-user
  user:
    username: admin
    password: password123
```

then i could add it as secret and mount this to a file which can be referenced by the prometheus-auth-config flag in the adapter

Evrything would be in place for this, because i can use extraVolumes and extraVolumeMounts

I have two problems:

1. it is complex
2. Per https://github.com/kubernetes-sigs/prometheus-adapter/issues/439#issuecomment-896718145 this was not possible

### Additional context.

_No response_

zeritti:

This PR adds support for setting environment variables loaded from a secret or a config map.

- fixes #2842

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.4.3
+version: 3.5.0
 appVersion: v0.10.0
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.4.2
+version: 3.4.3
 appVersion: v0.10.0
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/templates/deployment.yaml
+++ b/charts/prometheus-adapter/templates/deployment.yaml
@@ -44,9 +44,9 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        {{- if .Values.envFrom }}
-        envFrom:
-        {{- toYaml .Values.envFrom | nindent 8 }}
+        {{- if .Values.env }}
+        env:
+        {{- toYaml .Values.env | nindent 8 }}
         {{- end }}
         args:
         - /adapter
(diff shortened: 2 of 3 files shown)


charts/prometheus-adapter/values.yaml
```diff
@@ -151,6 +151,14 @@ tls:
   certificate: |-
     # Public key of the APIService
 
+# Set environment variables from given secrets or configmaps
+envFrom: []
```
monotek, line 155: can we just use "env" please.
it's mor flexible and can refetence configmaps and secrets too.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-container-environment-variables-using-configmap-data
https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

charts/prometheus-adapter/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 3.4.2
+version: 3.4.3
```
monotek, line 3: new feature should raise minor version

monotek (changes requested):

charts/prometheus-adapter/values.yaml
zeritti: Yes, changed.

charts/prometheus-adapter/Chart.yaml
zeritti: Minor number updated.

charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -44,6 +44,10 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        {{- if .Values.env }}
```
monotek, line 47: please use `{{- with .Values.env }}`

monotek (changes requested):

charts/prometheus-adapter/templates/deployment.yaml
zeritti: Changed. I preferred to stick with the conditional as it is quite common in this chart.

monotek (approved):

Issue #2874, opened 2022-12-30
[prometheus] VPA - Old apiVersion(bump v1beta1 to v1)


arukiidou:
### Is your feature request related to a problem ?

- VPA - Old apiVersion(v1beta1)

### Comparison with node-exporter(Before sorting)

![image](https://user-images.githubusercontent.com/10738333/210078965-296081c4-7ee1-4272-ad43-68f5f81b3aaf.png)

### Comparison with node-exporter(After sorting)

![image](https://user-images.githubusercontent.com/10738333/210080399-bc621aec-be5c-40d7-b50d-eac4c119caae.png)

### Describe the solution you'd like.

bump APIVersion

### Describe alternatives you've considered.

None.

PR #2875, opened 2023-01-01
[prometheus] vpa.yaml - bump apiVersion(v1beta1 to v1)

arukiidou:
### What this PR does / why we need it

- vpa.yaml - bump apiVersion(v1beta1 to v1)

- closes https://github.com/prometheus-community/helm-charts/issues/2874

- refactoring namespace is separated.
  - see: https://github.com/prometheus-community/helm-charts/pull/2883

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/templates/clusterrolebinding.yaml
+++ b/charts/prometheus/templates/clusterrolebinding.yaml
@@ -8,7 +8,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: {{ template "prometheus.serviceAccountName.server" . }}
-{{ include "prometheus.namespace" . | indent 4 }}
+    namespace: {{ template "prometheus.namespace" . }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/charts/prometheus/templates/clusterrolebinding.yaml
+++ b/charts/prometheus/templates/clusterrolebinding.yaml
@@ -8,7 +8,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: {{ template "prometheus.serviceAccountName.server" . }}
-    namespace: {{ template "prometheus.namespace" . }}
+{{ include "prometheus.namespace" . | indent 4 }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/charts/prometheus/templates/clusterrolebinding.yaml
+++ b/charts/prometheus/templates/clusterrolebinding.yaml
@@ -8,7 +8,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: {{ template "prometheus.serviceAccountName.server" . }}
-{{ include "prometheus.namespace" . | indent 4 }}
+{{ printf "namespace: %s" (include "prometheus.namespace" .) | nindent 4  }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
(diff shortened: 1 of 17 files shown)


charts/prometheus/templates/clusterrolebinding.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: {{ template "prometheus.serviceAccountName.server" . }}
-{{ include "prometheus.namespace" . | indent 4 }}
+    namespace: {{ template "prometheus.namespace" . }}
```
zanhsieh, line 11: This is not relevant. Please keep it as-is. Plus, the official document recommend to use `{{ include ... }}` not `{{ template ... }}`.  See:

https://helm.sh/docs/chart_template_guide/named_templates/#the-include-function
https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function

charts/prometheus/templates/vpa.yaml
```diff
@@ -1,11 +1,11 @@
-{{- if .Values.server.verticalAutoscaler.enabled -}}
-apiVersion: autoscaling.k8s.io/v1beta2
+{{- if and (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") (.Values.server.verticalAutoscaler.enabled) }}
+apiVersion: autoscaling.k8s.io/v1
```
zanhsieh, line 2: Just curious if any document (e.g. in README.md) relates to upgrade (e.g. which versions K8S drop support).

zanhsieh (changes requested):

charts/prometheus/templates/clusterrolebinding.yaml
arukiidou: Ok, revert and keep it as-is.

charts/prometheus/templates/vpa.yaml
arukiidou: I think these are out of our scope,

VPA is disable by default and requires pre-installation of CRD.

https://pkg.go.dev/k8s.io/autoscaler/vertical-pod-autoscaler#section-readme

charts/prometheus/templates/clusterrolebinding.yaml
arukiidou: Done, but a slightly clumsy solution.

Split the namespace fix if you want, do you?

zanhsieh:
@arukiidou 
I would keep non-related change away. Keep it simple and straightforward, which means not to find good excuses for maintainers not approve your PR.

arukiidou:
> @arukiidou I would keep non-related change away. Keep it simple and straightforward, which means not to find good excuses for maintainers not approve your PR.

ok, namespace refactoring will come later

zanhsieh (approved):

monotek (commented):
Do I miss something or will that break for all users still use the old API?

If so major version change and upgrading docs should have been upgraded.

arukiidou:
@monotek 
Very sorry.
I will work on a compatible fix soon

arukiidou:
@monotek

To prevent 19.3.0 from being used, needs 19.3.1 which is compatible with 19.2.2. Which fix do you want?

1. Make a patch to restore compatibility between 19.2.2 and 19.3.1 (#2896)
   - still a draft, because needs additional testing
2. Revert this, and create 19.3.1 which is exactly the same as 19.2.2

monotek:
19.3.1 should add a fix that both api versions are supported, by using "apabilities.APIVersions.Has" function.

arukiidou:
> 19.3.1 should add a fix that both api versions are supported, by using "apabilities.APIVersions.Has" function.

#2896 is both supported v1/v1beta2.

```
{{- if .Values.server.verticalAutoscaler.enabled -}}
{{- if .Capabilities.APIVersions.Has "autoscaling.k8s.io/v1/VerticalPodAutoscaler" }}
apiVersion: autoscaling.k8s.io/v1
{{- else }}
apiVersion: autoscaling.k8s.io/v1beta2
{{- end }}
```

PR #2897, opened 2023-01-07
[prometheus-adapter] Make securityContext user-settable

zeritti:

This PR makes setting container's `securityContext` possible through the values file and enables thus setting `seccompProfile` as well. The value `runAsUser` is being merged with the new `securityContext`.

Changes to the securityContext values themselves as they have been set so far have not been made.
  
None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/values.yaml
+++ b/charts/prometheus-adapter/values.yaml
@@ -13,9 +13,6 @@ metricsRelistInterval: 1m
 
 listenPort: 6443
 
-# User to run adapter container as
-runAsUser: 10001
-
 nodeSelector: {}
 
 priorityClassName: ""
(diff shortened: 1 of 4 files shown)


monotek (approved):

charts/prometheus-adapter/values.yaml
```diff
@@ -13,9 +13,6 @@ metricsRelistInterval: 1m
 
 listenPort: 6443
 
-# User to run adapter container as
-runAsUser: 10001
```
monotek, line 17: even that the value does not change this might be not working for updating users, which change d that values.
therfore major version  update is needed with upgrade instructions in readme.

monotek (changes requested):

charts/prometheus-adapter/values.yaml
zeritti: Understood. Upgrade note added, major version number bumped.

charts/prometheus-adapter/values.yaml
```diff
@@ -45,6 +42,18 @@ replicas: 1
...
+# ref. https://kubernetes.io/docs/tasks/configure-pod-container/security-context
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop: ["all"]
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 10001
+  # seccompProfile:
+  #   type: RuntimeDefault
```
steven-sheehy, line 55: I think we can consider including this as the default since it does provide a security improvement and we're already doing a major version bump.

steven-sheehy (dismissed):

charts/prometheus-adapter/values.yaml
zeritti: I second that and will update it. Thank you.

steven-sheehy (approved):

monotek (approved):

PR #2905, opened 2023-01-09
[prometheus-mysql-exporter] add imagePullSecrets and fullname override

yaskinny:
let user set registry pull secrets if they want to use their private registry.

add imagePullSecrets to mysql-exporter chart  
add fullname override(You had it in helpers templates but not in values file)  
set default image tag from .Values.image.tag to .Chart.AppVersion  

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/Chart.yaml
+++ b/charts/prometheus-mysql-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for prometheus mysql exporter with cloudsqlproxy
 name: prometheus-mysql-exporter
-version: 1.11.1
+version: 1.11.2
 home: https://github.com/prometheus/mysqld_exporter
 appVersion: v0.14.0
 sources:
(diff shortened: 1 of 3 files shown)


charts/prometheus-mysql-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for prometheus mysql exporter with cloudsqlproxy
 name: prometheus-mysql-exporter
-version: 1.11.1
+version: 1.11.2
```
monotek, line 4: New feature should raise minor version.

charts/prometheus-mysql-exporter/values.yaml
```diff
@@ -1,14 +1,22 @@
...
+
+## override release name
+fullnameOverride: ""
 
 replicaCount: 1
 
 image:
   repository: "prom/mysqld-exporter"
-  tag: "v0.14.0"
+  ## default on chart appVersion
```
monotek, line 12:
```suggestion
  ## if not set charts appVersion var is used
```

monotek (changes requested):

yaskinny:
I made your requested changes,  
if there's no more changes that I need to make, merge this PR.

monotek (approved):

PR #2907, opened 2023-01-10
kube-prometheus-stack: change admission webhook image registry defaults

akamac:
Right now, even if `prometheusOperator.admissionWebhooks.patch.image.registry` is defined, `global.imageRegistry` is preferred for webhook image.


monotek (changes requested):
chart version needs to be updated

PR #2910, opened 2023-01-11
[prometheus-blackbox-exporter] Fix logging options can be set individually

naohiroo:
Signed-off-by: naohiro <naohiro.dev@gmail.com>

Suggest that the logging option, like other charts, should be able to be specified from the each values. ( Currently, it can only be set via extraArgs. )

#### Checklist

- [x] Chart Version bumped


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.2.0
+version: 7.2.1
```

```suggestion
version: 7.3.0
```

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -15,6 +15,11 @@ podDisruptionBudget: {}
 ## Allow automount the serviceaccount token for sidecar container (eg: oauthproxy)
 automountServiceAccountToken: false
 
+## Only log messages with the given severity or above. One of: [debug, info, warn, error]
+# logLevel: info
```
monotek, line 19: what are the defaults for these settings (when not set)?
i would prefer to have them used instead of commented out?

monotek (changes requested):

charts/prometheus-blackbox-exporter/values.yaml
naohiroo: Thanks for review.

It will be info.
>what are the defaults for these settings (when not set)?

Fixed. ( Also  fixed log format part same way. )
>i would prefer to have them used instead of commented out?

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -18,6 +18,11 @@ podDisruptionBudget: {}
 ## Allow automount the serviceaccount token for sidecar container (eg: oauthproxy)
 automountServiceAccountToken: false
 
+## Only log messages with the given severity or above. One of: [debug, info, warn, error]
+logLevel: info
```
monotek, line 22: can we please use it like this:

```
log:
  format: logfmt
  level: info
```

monotek (changes requested):

monotek:
@zanhsieh 
please stop mergeing for no reason.
this creates notification noise.

nlamirault:
any news on this feature ?

Issue #1041, opened 2021-06-04
[kube-prometheus-stack] Helm-upgrade fails due to admission-webhooks being unable to reach service


Jaskaranbir:
**Describe the bug**
It seems like admission-webhooks arent patched to ignore failures during Helm-upgrades. This causes admission-webhooks to fail while the pod is being patched in single-replica deployments.

**Version of Helm and Kubernetes**:
Helm Version:

```console
$ helm version
version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}
```

Kubernetes Version:

```console
$ kubectl version
v1.19.9-gke.1900
```

**Which chart**: kube-prometheus-stack

**Which version of the chart**: 16.1.2

**What happened**:

[...]

PR #2911, opened 2023-01-12
[kube-prometheus-stack] More secure admissionWebhooks failurePolicy

sathieu:

I need this to avoid kube-prometheus-stack being OutOfSync in ArgoCD because of failurePolicy. Also, this could be a security issue, because the webhook is not enforced during upgrades.

NB: The default behavior is preserved

Related issue https://github.com/prometheus-community/helm-charts/issues/1041

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 43.3.1
+version: 44.0.0
 appVersion: 0.61.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -22,7 +22,7 @@ sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
 version: 44.0.0
-appVersion: 0.62.0
+appVersion: v0.62.0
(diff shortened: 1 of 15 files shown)


monotek (changes requested):
If its a feature update it would be 43.4.0 but this seems to be a braking change.
Therofre major update with update instructions would be needed.
I would prefer to keep the current value.

I also don't understand why this change is needed at all, as it's allready configurable?
You could set "Ignore" or "Fail"  already now?

sathieu:
Thanks for your review @monotek :pray:.

> If its a feature update it would be 43.4.0 but this seems to be a braking change. Therofre major update with update instructions would be needed.

I've updated the PR with major version bump and upgrade notes. This breaking change will only affect people setting `prometheusOperator.admissionWebhooks.failurePolicy` when it had no effect.

> I would prefer to keep the current value.
> I also don't understand why this change is needed at all, as it's allready configurable? You could set "Ignore" or "Fail" already now?

No, and this is the problem: when `prometheusOperator.admissionWebhooks.patch.enabled`
is `true` (the default), `prometheusOperator.admissionWebhooks.failurePolicy` is ignored.

Back to you for another review :badminton: !

monotek (dismissed):
I've included the prom operator update, so we don't have to bump the major version again.

sathieu:
Thanks @monotek!

neiser:
@monotek @sathieu I think the proposed changes break upgrading `kube-prometheus-stack` again, as outlined in #1041. The problem lies in the fact that when `failurePolicy` is not set (default to empty), then the job template `kube-prometheus-stack-admission-patch` renders the argument `--patch-failure-policy=`  instead of the desired `--patch-failure-policy=Fail`. Having `Ignore` during upgrades is mandatory as otherwise a missing `caBundle` breaks deploying any resources.

So, I propose this change #3051

sathieu:
Thanks @neiser.

PR #2912, opened 2023-01-12
[prometheus] add extra scrape config with secrets

nyambati:
This PR adds the option to have extra scrape configurations with secrets which allows us to split scrape configs that need authentication to one file for encryption especially when using sops and helm-secrets

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: v2.41.0
-version: 19.3.1
+version: 19.3.2
 kubeVersion: ">=1.16.0-0"
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
Why is this needed?
This can already be configrued in extraScrapeConfigs?

nyambati:
our team manages a lot of Prometheus instances for scalability we have shared scrape configurations that are passed via server files and then use extra configurations via ` extraScrapeConfigs` to cater to the unique needs of each cluster. When adding scrape configuration with sensitive credentials, we are encrypting the entire file even though it is only a small change. This approach creates extra steps for us when troubleshooting, and PR reviews as we have to decrypt the file again and again. Separating sensitive scrape configuration offers us the visibility we need and also reduce unnecessary encryption and decryption process.

monotek:
I don't think we should update the chart because of some exotic setup doubling functionality.

PR #2615, opened 2022-10-26
[kube-prometheus-stack] add servicemonitor/podmonitor scrape limits

mblaschke:
#### What this PR does / why we need it
adds support for servicemonitor/podmonitor scrape limits (eg. sampleLimit) to kube-prometheus-stack (replacement for #2608)

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [ ] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #2913, opened 2023-01-12
[kube-prometheus-stack] modify prometheusOperator.serviceMonitor.sampleLimit default value 0 (mean no limit)

BobDu:
Signed-off-by: BobDu <i@bobdu.cc>

- fixes https://github.com/prometheus-community/helm-charts/pull/2615#issuecomment-1379866223

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -2024,7 +2024,7 @@ prometheusOperator:
 
     ## SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
     ##
-    sampleLimit: 1000
+    sampleLimit: 0
 
     ## TargetLimit defines a limit on the number of scraped targets that will be accepted.
     ##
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -2024,7 +2024,7 @@ prometheusOperator:
 
     ## SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
     ##
-    sampleLimit: 100
+    sampleLimit: 1000
```
monotek, line 2027: i would go for "0" so the old behaviour is restored.

```suggestion
    sampleLimit: 0
```

monotek (changes requested):

monotek (approved):

charts/kube-prometheus-stack/values.yaml
BobDu: good idea, There really isn't enough need to limit

PR #2936, opened 2023-01-19
[kube-state-metrics] Use appVersion to define container image version

mrueg:
Adds logic so it will use appVersion to set the container image tag and we don't need to repeat ourselves.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -3,7 +3,7 @@ prometheusScrape: true
 image:
   repository: registry.k8s.io/kube-state-metrics/kube-state-metrics
   # If unset use v + .Charts.appVersion
-  #tag:
+  # tag:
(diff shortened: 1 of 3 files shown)


charts/kube-state-metrics/values.yaml
```diff
@@ -2,7 +2,8 @@
 prometheusScrape: true
 image:
   repository: registry.k8s.io/kube-state-metrics/kube-state-metrics
-  tag: v2.7.0
+  # If unset use v + .Charts.appVersion
+  #tag:
```
monotek, line 6:
```suggestion
  # tag:
```

monotek (changes requested):

dotdc (approved):
LGTM

monotek (approved):

PR #2937, opened 2023-01-19
[prometheus-cloudwatch-exporter] Support extra manifests in cloudwatch exporter.

dtrejod:
A simple change that adds support for rendering additional manifests in cloudwatch-exporter.

cc @gianrubio, @torstenwalter, @asherf

- fixes #

Similar functionality already supported by the prometheus helm chart. This PR copies that functionality. 
https://github.com/prometheus-community/helm-charts/blob/f034c7cb0ce9f1275f697292dfe8f2e1db0431c3/charts/prometheus/values.yaml#L1047-L1055

#### Checklist

- [x] Chart Version bumped


asherf:
FWIW, I don't this is a good change as there are better way to achieve the same goal.
for example, you could have your own chart that reference the CW exporter chart and then in that chart templates directory put whatever manifests you need.

this will look something like this:
Chart.yaml:
```yaml
apiVersion: v2
name: my-chart
version: 0.1.0
dependencies:
- name: prometheus-cloudwatch-exporter
  # prometheus-cloudwatch-exporter app version: 0.15.1 (https://github.com/prometheus/cloudwatch_exporter)
  version: 0.24.0
  repository: https://prometheus-community.github.io/helm-charts
```

in your values file you will configure values for the CW chart like so:
values yaml:
```yaml
prometheus-cloudwatch-exporter:
  serviceAccount:
    name: cloudwatch-exporter
  resources:
    requests:
      cpu: 30m
      memory: 256Mi
    limits:
        cpu: 250m
        memory: 512Mi
  # https://github.com/prometheus/cloudwatch_exporter#configuration
  config: |-
    region: us-east-1
    metrics:
      # Lambda metrics
      - aws_dimensions: [FunctionName, Resource]
        aws_metric_name: Duration
        aws_namespace: AWS/Lambda
      - aws_dimensions: [FunctionName, Resource]
        aws_metric_name: Invocations
...... more stuff....
```

and then, as I mentioned the template directory for your chart can have whatever k8s objects/manifests you want.
and the advantage is, that those are templates that can be configured via your values.yaml if you need to do so.
I hope this helps.

dtrejod:
> FWIW, I don't this is a good change as there are better way to achieve the same goal.
for example, you could have your own chart that reference the CW exporter chart and then in that chart templates directory put whatever manifests you need.


Thanks for the in-depth reply! Yes this is the approach we ended up going with. We have our own chart that imports this one and templates out the additional manifests we need.

PR #2938, opened 2023-01-20
[prometheus-operator-crds] Keep upstream annotations

zeritti:

In this PR I am proposing
1. change in the update utility and 
3. refresh of the manifests. 

At the moment, any annotations that might be present in the CRD manifests are not kept (see annotation `controller-gen.kubebuilder.io/version`). I propose that these be preserved.

Through a small patch to the utility, present annotations from upstream will be kept and user-defined annotations will be supported as is currently the case.

I raised the minor version number only due to the previously ommitted annotations.

#### Checklist

- [x] Chart Version bumped


desaintmartin (approved):
Thanks, indeed it is closer from upstream!

PR #2944, opened 2023-01-24
[kube-prometheus-stack] Remove false alarm on a volume with infinite inodes

garo:
On some storage systems, such as a CephFS filesystem, the underlying storage doesn't have any limit on inodes.

If the kubelet_volume_stats_inodes_free metric is always zero, then the KubePersistentVolumeInodesFillingUp alert will always fire.

This patch adds a simple conditional to ensure that the kubelet_volume_stats_inodes_free is meaningful for the volume and thus fixing the false alert on this edge case.

Signed-off-by: Juho Mäkinen <juho.makinen@gmail.com>

#### Checklist

- [ ] Chart Version bumped


monotek (changes requested):
Needs to be fixed upstream. See: 

https://github.com/prometheus-community/helm-charts/blob/48115dea3fef920546c5cf69c8df1a5c80789744/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kubernetes-storage.yaml#L1-L5

Also chart version would need a raise.

garo:
Thank you, I'll check the process on how to get this into the upstream

PR #2948, opened 2023-01-24
prometheus-operator: add EndpointSlices for operator to support K8s 1.22+

prudnitskiy:

This MR adds EndpointSlices permissions for Kube-prometheus-operator to support K8s version 1.22+

There's no open MR for this repo, but there's an ongoing discussion in Prometheus operator [repository](https://github.com/prometheus-operator/prometheus-operator/issues/3862)

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
Why should we add it now if it's not even implemented upstream?

prudnitskiy:
As far as I see it from my own experience - it is already implemented somehow (?). Without this change prometheus-operator constantly restarts without any log messages.

Xtigyro (changes requested):
Hey @prudnitskiy - thanks for the contribution!

Could you please bump the minor version (instead of the bugfix version)?

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
```diff
@@ -78,4 +78,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
Xtigyro, line 81: Can we put also an `if` clause around this and have this deployed only when the K8s version is `>=1.22`?

Xtigyro (changes requested):

TsengSR:
> As far as I see it from my own experience - it is already implemented somehow (?). Without this change prometheus-operator constantly restarts without any log messages.

I'm seeing the same issue on the operator after upgrading to 45.0 helm chart. Running against k8s v1.21.7 (can't upgrade right now for several reasons). I tried adding the permissions to the clusterrole myself, but the service is still crashing

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
TsengSR: According to the [k8s docs](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) endpoint slices are a 1.21 feature

monotek: But the chart claims to be k8s 1.16 compatbile: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prerequisites

prudnitskiy: Case added, please check

monotek (approved):

istheway:
For reference, this is problematic for k8s 1.26.2 that I just attempted deploying the latest operator with the kube-prometheus-stack helm chart.  The kube-prometheus-stack-operator CrashLoopBackoff on the following logs:

`level=warn ts=2023-04-18T23:37:00.703117184Z caller=operator.go:332 component=prometheusoperator msg="failed to check if the API supports the endpointslice resources" err="converting (v1.APIGroup) to (v1.APIResourceList): unknown conversion"`

What's interesting is the bind address tcp 10250 is not in use at the moment on any service.

PR #2951, opened 2023-01-26
[prometheus-adapter]: allow setting annotations on the deployment

weikinhuang:

This PR adds the ability to set annotations on the Deployment resource itself, which is currently not possible. Tooling like `stakater/Reloader` requires that the annotation be set on the deployment/statefulset workload and not as a pod annotation.

I also fixed an issue with a duplicate key in `spec.metadata.annotations` for the deployment resource that kubeconform found when setting the `customAnnotations` key.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/values.yaml
+++ b/charts/prometheus-adapter/values.yaml
@@ -217,6 +217,9 @@ podLabels: {}
 # Annotations added to the pod
 podAnnotations: {}
 
+# Annotations added to the deployment
+deploymentAnnotations: {}
+
 hostNetwork:
   # Specifies if prometheus-adapter should be started in hostNetwork mode.
   #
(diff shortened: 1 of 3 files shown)


charts/prometheus-adapter/values.yaml
```diff
@@ -217,6 +217,9 @@ podLabels: {}
 # Annotations added to the pod
 podAnnotations: {}
 
+# Annotations added to the deployment
+deploymentAnnotations: {}
```
weikinhuang, line 221: I've personally been naming this key as `workloadAnnotations` in my personal charts.

monotek (approved):

weikinhuang:
thanks!

PR #2957, opened 2023-01-26
[Prometheus-elasticsearch-exporter] Overide exporter tag with docker …

mvnarsing:
…instead of main repo

This PR overrides elasticsearch-exporter tag with latest releases available in docker, whereas main repo is not updated in few months

Using latest tag addresses some of the CVE issues we are seeing and staying on top of latest one helps

#### Checklist

- [ ] Chart Version bumped


monotek (changes requested):
i don't think it's a good idea to change to the bitnami image by default.
if you want an update, add it upsrtream.

besides there are current builds in our repo: 
https://hub.docker.com/r/prometheuscommunity/elasticsearch-exporter/tags
https://quay.io/repository/prometheuscommunity/elasticsearch-exporter?tab=tags

you can use them in a fixed manor via using the digest.

monotek (changes requested):
.

PR #2969, opened 2023-01-30
[kube-prometheus-stack] Added namespace selector for webhook configuration

### LINKED ISSUE #2915: [kube-prometheus-stack] webhook configuration should respect operators namespace filter

When configuring the operator to (not) interact with resources for specific namespaces only by using the `prometheusOperator.namespaces` or `prometheusOperator. denyNamespaces` I would expect that also the webhooks are not getting triggered for the excluded namespaces. Indeed, the webhooks are ignoring that options and are getting triggered for all namespaces.

I would expect that the webhooks are respecting the namespaces configuration id present, so that there is also no interaction from apiserver to the operator instance for specific namespaces.

So translating the mentioned options above to an addition to the webhook configuration like:
```yaml
namespaceSelector:
  matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: NotIn
      values:
      - "myDeniedNamespace"
```

Alternatively the webhook namespaceSelector could be configurable via values.yaml independently, but a re-usage of the mentioned operator values sounds more reasonable.

### Additional context.

If that request sounds reasonable, I could provide a PR for the change.

Teneroy:
Added an ability to override namespace selector for webhook configuration

Signed-off-by: Aleksei Chernyshov <aleksei.chernyshov@sap.com>

Currently, it is not possible to configure namespaceSelector for ValidatingWebhookConfiguration and MutatingWebhookConfiguration, because there is no namespaceSelector in corresponding template files. This PR adds this functionality. The person can either deny namespaces by setting denyNamespaces, or to allow specific namespaces by setting namespaces.

- fixes #2915

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 44.3.0
+version: 44.3.1
 appVersion: v0.62.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
@@ -41,4 +41,26 @@ webhooks:
     timeoutSeconds: {{ .Values.prometheusOperator.admissionWebhooks.timeoutSeconds }}
     admissionReviewVersions: ["v1", "v1beta1"]
     sideEffects: None
+    {{- if or .Values.prometheusOperator.denyNamespaces .Values.prometheusOperator.namespaces }}
+    namespaceSelector:
+      matchExpressions:
+      {{- if .Values.prometheusOperator.denyNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        {{- range $namespace := .Values.prometheusOperator.denyNamespaces }} 
+        - {{ $namespace }}
+        {{- end }}
+      {{- else if and .Values.prometheusOperator.namespaces .Values.prometheusOperator.namespaces.additional }}
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values:
+        {{- if and .Values.prometheusOperator.namespaces.releaseNamespace (default .Values.prometheusOperator.namespaces.releaseNamespace true) }}
+        - {{ .Release.Namespace }}
+        {{- end }}
+        {{- range $namespace := .Values.prometheusOperator.namespaces.additional }} 
+        - {{ $namespace }}
+        {{- end }}
+      {{- end }}
+    {{- end }}
 {{- end }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/validatingWebhookConfiguration.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/validatingWebhookConfiguration.yaml
@@ -41,4 +41,26 @@ webhooks:
     timeoutSeconds: {{ .Values.prometheusOperator.admissionWebhooks.timeoutSeconds }}
     admissionReviewVersions: ["v1", "v1beta1"]
     sideEffects: None
+    {{- if or .Values.prometheusOperator.denyNamespaces .Values.prometheusOperator.namespaces }}
+    namespaceSelector:
+      matchExpressions:
+      {{- if .Values.prometheusOperator.denyNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        {{- range $namespace := .Values.prometheusOperator.denyNamespaces }} 
+        - {{ $namespace }}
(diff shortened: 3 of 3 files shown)


charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -41,4 +41,26 @@ webhooks:
...
+    namespaceSelector:
+      matchExpressions:
+      {{- if .Values.prometheusOperator.denyNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        {{- range $namespace := .Values.prometheusOperator.denyNamespaces }} 
+        - {{ $namespace }}
+        {{- end }}
+      {{- else if and .Values.prometheusOperator.namespaces .Values.prometheusOperator.namespaces.additional }}
```
QuentinBisson, line 54: Why would those 2 properties be mutually exclusive?

charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -41,4 +41,26 @@ webhooks:
...
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values:
+        {{- if and .Values.prometheusOperator.namespaces.releaseNamespace (default .Values.prometheusOperator.namespaces.releaseNamespace true) }}
```
QuentinBisson, line 58: I would use the same syntax as the one used here https://github.com/prometheus-community/helm-charts/blob/22b98bae93a67016082ec883d4d846c1282b5bf7/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L75

Also it is best to use https://github.com/prometheus-community/helm-charts/blob/22b98bae93a67016082ec883d4d846c1282b5bf7/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L1 for the release.namespace for cases when the operator is used as a subchart.

Could you address those and bump the version?

Teneroy: @QuentinBisson  Yes, working on it right now

charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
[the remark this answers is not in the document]
Teneroy: @QuentinBisson Those are mutually exclusive, but I still need to handle those two cases differently. For instance, whenever I have denyNamespaces I should use NotIn, while for namespaces I should use In. Should I maybe use ternary operator here?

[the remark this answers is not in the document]
QuentinBisson: If they are mutually exclusive then I think the way you set it is fine

charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/mutatingWebhookConfiguration.yaml
Teneroy: @QuentinBisson I don't really understand how the line 58 can be transform in order to use syntax from the line 75 in deployment.yaml. Can you elaborate on it a little bit more, please? 

Regarding your second comment, am I understanding it right, that it is related to the line 59, and you want me to change `- {{ .Release.Namespace }}` to something like `{{- $namespace := printf "%s" (include "kube-prometheus-stack.release.namespace" .) }} ` ?

QuentinBisson: Yes you are right for the second comment, I think you added a line but the comment stayed on this line number for some reason 

For the second part, I would maybe just use mustUnique everywhere and that would be fine

Teneroy: @QuentinBisson I addressed an issue regarding `release.namespace`. However, I still don't understand how `mustUnique` is related to the if statement on the line 58. Maybe you want me to exclude any potentially similar namespaces in `.Values.prometheusOperator.namespaces.additional` ?

QuentinBisson: Yes I would avoid duplicates in either denyNamespaces or .Values.prometheusOperator.namespaces.additional :)

Teneroy: @QuentinBisson  I addressed issues you mentioned. Can you take a look again, please?

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 45.21.0
+version: 45.24.1
```
QuentinBisson, line 24:
```suggestion
version: 45.25.0
```

QuentinBisson (approved):

PR #2974, opened 2023-02-01
[prometheus-blackbox-exporter] adding optional additional hostname parameter to servicemonitor 

smbambling:

Allows adding an optional hostname parameter for serviceMonitor targets that will set the Host header and TLS SNI for HTTP probes. This avoids the need to created individual modules with specific tlsconfigs when probing against targets that are vhosts

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.2.0
+version: 7.2.1
 appVersion: 0.23.0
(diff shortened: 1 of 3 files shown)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.2.0
+version: 7.2.1
```

monotek (changes requested):

charts/prometheus-blackbox-exporter/Chart.yaml
smbambling: Do'h, yep that makes sense. I'll push up a change momentarily

monotek (approved):

PR #2997, opened 2023-02-07
[kube-prometheus-stack] Enable network policy for prometheus

QuentinBisson:

This PR adds a Network Policy to prometheus in the kube-prometheus-stack so it can scrape components behind a deny all network policy

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 45.5.0
+version: 45.6.0
 appVersion: v0.63.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus/networkpolicy.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/networkpolicy.yaml
@@ -4,29 +4,29 @@ kind: NetworkPolicy
 metadata:
   labels:
     app: {{ template "kube-prometheus-stack.name" . }}-prometheus
-{{- include "kube-prometheus-stack.labels" . | indent 4 }}   
+    {{- include "kube-prometheus-stack.labels" . | nindent 4 }}
   name: {{ template "kube-prometheus-stack.fullname" . }}-prometheus
   namespace: {{ template "kube-prometheus-stack.namespace" . }}
 spec:
+  {{- if .Values.prometheus.networkPolicy.egress }}
+  ## Deny all egress by default
   egress:
-    {{- if .Values.prometheus.networkPolicy.egress }}
     {{- toYaml .Values.prometheus.networkPolicy.egress | nindent 4 }}
-    {{- else }}
-    ## Allows all by default
-    - {}
-    {{- end }}
+  {{- end }}
+  {{- if .Values.prometheus.networkPolicy.ingress }}
+  # Deny all ingress by default (prometheus scrapes itself using localhost)
   ingress:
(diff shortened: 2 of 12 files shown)


charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 44.3.1
+version: 44.4.0
```
monotek, line 24:
```suggestion
version: 44.5.0
```

charts/kube-prometheus-stack/templates/prometheus/networkpolicy.yaml
```diff
@@ -0,0 +1,32 @@
...
+  labels:
+    app: {{ template "kube-prometheus-stack.name" . }}-prometheus
+{{- include "kube-prometheus-stack.labels" . | indent 4 }}   
+  name: {{ template "kube-prometheus-stack.fullname" . }}-prometheus
+  namespace: {{ template "kube-prometheus-stack.namespace" . }}
+spec:
+  egress:
+    {{- if .Values.prometheus.networkPolicy.egress }}
+    {{- toYaml .Values.prometheus.networkPolicy.egress | nindent 4 }}
+    {{- else }}
```
monotek, line 14: default can be defined in the values yaml.
i would likely not use "allow all" by default but give a set of reasonable policies.

monotek (changes requested):

charts/kube-prometheus-stack/templates/prometheus/networkpolicy.yaml
QuentinBisson: Pushed a new version with limited default ingress

monotek: We might want to make the whole .spec part configuralbe by using "{{ toYaml ...}}"?
Just in case somebody likes to change selectors too?

egress default for the operator would be kube api. 
prometheus should be able to reach 9090 for the scrapes. 
Not sure if i miss something?

We could also use deny all as default so everybody has to care individually if networks policy is enabled.

QuentinBisson: I would think that someone wanting to enable the netpol would have to change the egress anyway unless we set it to allow all. As enablement is a conscious choice, let's limits it to 9090, the kube api and alertmanager then. 
I'll try to figure out how :D 

Regarding the selectors, maybe we allow them to override them for now but not the whole spec? Otherwise we would need to do the same change for prometheus-operator network policies :)

monotek: Just wanted to go the easy path instead defining overrides for every available option.
In the end i don't much of an opinion here.
Maybe wait for some feedback of the charts maintainers.

QuentinBisson: @scottrigby @Xtigyro does any one of you have ideas?

Otherwise let's make it deny  or allow all and users will override it to fit their needs

QuentinBisson: @monotek I made the egress deny all by default, letting the users override it for now. This could be revisited later

monotek (approved):

QuentinBisson:
Bumping chart version @monotek can you review? I'm still not approved to merge this repo :(

QuentinBisson:
@monotek I think you forgot to sign your commit :D

monotek:
yes, damn web editor :/
can you please rebase / squash my commit?

QuentinBisson:
Sure :)

QuentinBisson:
@monotek should be good now

monotek (approved):

PR #2999, opened 2023-02-07
[prometheus-node-exporter] Update VPA containerPolicies container name to match DaemonSet

hizeph:

Changes the DaemonSet container name to match the name used by the VerticalPodAutoscaler containerPolicies

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -41,7 +41,7 @@ spec:
       serviceAccountName: {{ include "prometheus-node-exporter.serviceAccountName" . }}
       containers:
         {{- $servicePort := ternary 8100 .Values.service.port .Values.kubeRBACProxy.enabled }}
-        - name: {{ include "prometheus-node-exporter.name" . }}
+        - name: node-exporter
           image: {{ include "prometheus-node-exporter.image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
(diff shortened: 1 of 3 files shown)


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -41,7 +41,7 @@ spec:
       serviceAccountName: {{ include "prometheus-node-exporter.serviceAccountName" . }}
       containers:
         {{- $servicePort := ternary 8100 .Values.service.port .Values.kubeRBACProxy.enabled }}
-        - name: node-exporter
+        - name: {{ include "prometheus-node-exporter.name" . }}
```
monotek, line 44: Imho this makes no sense. 
The vpa does'nt care for the container name.
It looks for " kind: DaemonSet".

monotek (changes requested):

charts/prometheus-node-exporter/templates/daemonset.yaml
monotek: I still don't think a dynamic containername is a good idea.

hizeph: That's fair, I'll update the VPA object instead, to match the container name.

zanhsieh (dismissed):

zanhsieh:
@hizeph 
Could you sign DCO plz?

hizeph:
> @hizeph Could you sign DCO plz?

Done, sorry about that

zanhsieh (approved):

monotek (approved):

PR #3040, opened 2023-02-17
[kube-prometheus-stack] Fix incorrect values.yaml descriptions

mwipplinger-ftapi:
The descriptions for the `ruleNamespaceSelector`, `podMonitorNamespaceSelector`, `probeNamespaceSelector` were wrong which I found out the hard way by following the current description when trying to configure them and failing.
I have checked how they actually should be described and followed the lead on the description of `serviceMonitorNamespaceSelector` which seems to be correct. 

The old reference for https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#namespaceselector did not apply for these values as we are configuring prometheus spec at this point and instead refers to namespace selectors for other resources. I assume the fields + description were copy pasted at one point without checking if the description still applies. No matter how it came to this I just wanted to fix the description so no one else will run into the same problem as me. :)

#### Checklist

- [X] Chart Version bumped


GMartinez-Sisti:
Hi @mwipplinger-ftapi, thank you for the PR. Please rebase from main (or update branch) and bump the patch on the helm chart version.

GMartinez-Sisti (approved):

PR #3044, opened 2023-02-19
[prometheus-node-exporter] Allow parent charts to override registry hostname

johnswarbrick-napier:
Add support for optionally overriding the registry hostname from a parent chart.

This is important for air-gapped installations which do not have direct access to the public internet and must use a private, internal container registry.

It also avoids having to modify the Prometheus Node Exporter Helm chart directly.

FAO: @gianrubio @zanhsieh

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.13.1
+version: 4.13.2
 appVersion: 1.5.0
 home: https://github.com/prometheus/node_exporter/
 sources:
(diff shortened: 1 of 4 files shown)


johnswarbrick-napier:
Hi @zanhsieh, is there anything else you need from me re: this PR? Thanks.

zanhsieh (approved):

markmsmith:
This is kind of a tricky one, since there's 2 different scenarios to support, with opposite goals:  
1. In a controlled environment, you want to say "I don't care what anything else says, all images should come from this globally defined one for security reasons"
2. You have a default one you want used for most images, but you also want to have the option to override for individual ones.

Scenario 2 needs "most specific/nearest config wins", while 1 needs the global setting to dominate.  You can almost get what you want if the values were applied in the order:
* sub-chart defaults
* chart defaults
* overrides of global from user-supplied values.yaml
* overrides of individual images from user-supplied values.yaml

but that still doesn't support scenario 1 perfectly, and I'm not sure how easy helm makes that to do.

TLDR; we may want to think about having a way to control whether the global image registry is a default or an edict.  Maybe the current setting is the "edict" version, and we add a new one that's the `defaultImageRegistry`?

PR #3053, opened 2023-02-21
[kube-prometheus-stack] allow httpMethod selection on prometheus datasource

sgewirtz1:
This PR allows us to set the datasource http method for querying data from prometheus to POST, this supports larger queries instead of failing on large GET requests. As described in the attached grafana issue this is the default behavior in more updated versions of grafana and as such should have no negative impact on existing users.

From the linked PR:
POST allows for much larger query bodies that GET, which is needed when sending queries from graphs with a lot of targets (e.g., many hosts in a dashboard variable).

When queries fail due to the GET limitation on size of query body, it can be difficult for the Grafana user to troubleshoot, as the UI (and grafana.log) do not provide any info to indicate why the request has failed.

POST also makes the Query Inspector data easier to read, because the query is plaintext. (With GET, the query is URL encoded.)

In the past, it made sense to default to GET, because POST is only available in Prometheus >= v2.1.0 . But Prometheus 2.1.0 release is 3+ years old, so I think it makes sense to default to POST now.

https://github.com/grafana/grafana/issues/30837

- fixes #

This is difficult to test since the query where we saw failures was over 100,000 characters long.
#### Checklist

- [x] Chart Version bumped

Thank you for maintaining this tool!
@QuentinBisson @Xtigyro @scottrigby @GMartinez-Sisti @gkarthiks @gianrubio @andrewgkew


sgewirtz1:
@Xtigyro is there an issue with the PR?

sgewirtz1:
?

GMartinez-Sisti:
Hi @sgewirtz1, unfortunately I don't have enough context to know  if this would affect existing users. Could you update the description and possibly include some links explaining why this is a safe change?

Also, please bump minor instead since it's a feature.

Thank you!

GMartinez-Sisti (approved):

sgewirtz1:
Thanks!

PR #3064, opened 2023-02-23
[prometheus-kube-stack] Adds environment variables to prometheus

michbeck100:

Adds environment variables for prometheus. This can be everything, but specifically I would use it to set the timezone with `TZ`.

None

None

#### Checklist

- [x] Chart Version bumped


michbeck100:
Then I would extend the CRD. What do you think? I find your workaround not really intuitive.

PR #3087, opened 2023-03-02
[prometheus-blackbox-exporter] Make it possible to unset the container "securityContext“

Kirchen99:

When this chart is used as a subchart, this PR allows the securityContext in the container to be unset from the parent chart without change in previous behaviour.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
Imho this is a bit complicated.
I would likely rather use something like:

```
{{- with .Values.securityContext }}
          securityContext:
            {{- toYaml . | nindent 12 }}
{{- end }} 
```

Setting the current default values.

If needed update upgrading section and raise major version of the chart.

Kirchen99:
I don't think it's necessary to upgrade the major version number. Because there is no breaking change.

monotek (approved):

PR #3097, opened 2023-03-06
[kube-prometheus-stack] Add Vertical Pod Autoscaler to Prometheus

### LINKED ISSUE #3095: [kube-prometheus-stack] Enable vertical pod autoscaler support for prometheus

Currently, the chart enables `VPA` support for `prometheus-operator`,and I want `prometheus` to support `VPA` too.

There are two reasons:
1. In many cases, we will not or have not noticed that the resources of the default `prometheus` need to be adjusted. 
2.The current VPA already allows to change container resource requests & limits without necessarily restarting the container.

Support vertical pod autoscaler just like `prometheus-operator`.

/kube-prometheus-stack/templates/prometheus/verticalpodautoscaler.yaml:
```yaml
{{- if and (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") (.Values.prometheus.verticalPodAutoscaler.enabled) }}
apiVersion: autoscaling.k8s.io/v1
kind: VerticalPodAutoscaler
metadata:
  name: {{ template "kube-prometheus-stack.fullname" . }}-prometheus
  namespace: {{ template "kube-prometheus-stack.namespace" . }}
  labels:
    app: {{ template "kube-prometheus-stack.name" . }}-prometheus
spec:
  resourcePolicy:
    containerPolicies:
      - containerName: "prometheus"
      {{- if .Values.prometheus.verticalPodAutoscaler.controlledResources }}
        controlledResources: {{ .Values.prometheus.verticalPodAutoscaler.controlledResources }}
      {{- end }}
      {{- if .Values.prometheus.verticalPodAutoscaler.maxAllowed }}
        maxAllowed:
        {{- toYaml .Values.prometheus.verticalPodAutoscaler.maxAllowed | nindent 8 }}
      {{- end }}
      {{- if .Values.prometheus.verticalPodAutoscaler.minAllowed }}
        minAllowed:
        {{- toYaml .Values.prometheus.verticalPodAutoscaler.minAllowed | nindent 8 }}
      {{- end }}
  targetRef:
    apiVersion: apps/v1
    kind: StatefulSet
    name: prometheus-{{ include "kube-prometheus-stack.prometheus.crname" $ }}
  {{- if .Values.prometheus.verticalPodAutoscaler.updatePolicy }}
  updatePolicy:
    {{- if .Values.prometheus.verticalPodAutoscaler.updatePolicy.updateMode }}
    updateMode: {{ .Values.prometheus.verticalPodAutoscaler.updatePolicy.updateMode  }}
    {{- end }}
  {{- end }}
{{- end }}
```

`NONE`

### Additional context.

_No response_

z-anshun:
This PR adds support for `VPA` to `prometeus` so users do not have to care about the resource usage of the prometheus and let the `VPA` do it's job.

- fixes #3095 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 45.7.1
+version: 45.8.0
 appVersion: v0.63.0
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus/verticalpodautoscaler.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/verticalpodautoscaler.yaml
@@ -1,0 +1,35 @@
+{{- if and (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") (.Values.prometheus.verticalPodAutoscaler.enabled) }}
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: {{ template "kube-prometheus-stack.fullname" . }}-prometheus
+  namespace: {{ template "kube-prometheus-stack.namespace" . }}
+  labels:
+    app: {{ template "kube-prometheus-stack.name" . }}-prometheus
+{{ include "kube-prometheus-stack.labels" . | indent 4 }}
+spec:
+  resourcePolicy:
+    containerPolicies:
+      - containerName: "prometheus"
+      {{- if .Values.prometheus.verticalPodAutoscaler.controlledResources }}
+        controlledResources: {{ .Values.prometheus.verticalPodAutoscaler.controlledResources }}
+      {{- end }}
+      {{- if .Values.prometheus.verticalPodAutoscaler.maxAllowed }}
+        maxAllowed:
+        {{- toYaml .Values.prometheus.verticalPodAutoscaler.maxAllowed | nindent 8 }}
+      {{- end }}
+      {{- if .Values.prometheus.verticalPodAutoscaler.minAllowed }}
+        minAllowed:
+        {{- toYaml .Values.prometheus.verticalPodAutoscaler.minAllowed | nindent 8 }}
+      {{- end }}
+  targetRef:
+    apiVersion: apps/v1
+    kind: StatefulSet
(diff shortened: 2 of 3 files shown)


atsai1220:
Does this work? I imagine the Prometheus operator will revert any changes made by VPA.

I've tried to omit the resources field in my Prometheus object then manually edit the resources field (emulating VPA) but the operator will revert those fields.

QuentinBisson:
@atsai1220 yes this works with VPA as 1, VPA would use the statefulset and not the prometheus CR and 2, the vpa admission controller would set the recommended value on the pod directly.

atsai1220:
> @atsai1220 yes this works with VPA as 1, VPA would use the statefulset and not the prometheus CR and 2, the vpa admission controller would set the recommended value on the pod directly so even if the statefulset tries to change the value, the mutatingwebhook would set the recommended value anyway

Sounds good. I was under the assumption the `kube-prometheus-stack` chart deploys using the `Prometheus` CR here https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml

And there is another chart called `prometheus` that can deploy a regular statefulset and that already has vpa as an option. https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/vpa.yaml

QuentinBisson:
So it deploys prometheus as a Prometheus CR but the operator generates a statefulset behind it :) The only think we're missing is that this also works with autosharding

atsai1220:
> So it deploys prometheus as a Prometheus CR but the operator generates a statefulset behind it :) The only think we're missing is that this also works with autosharding

But if VPA is updating the statefulset then the prometheus operator will change it back, no?

QuentinBisson:
The controller will always to the calls anyway, that's how the reconciliation works :)

QuentinBisson:
Also, enabling it is not mandatory :)

atsai1220:
I’m having a hard time wrapping my head around this :(

Would you mind setting up a sandbox with Prometheus Operator and VPA?

atsai1220:
I think I get what you’re saying now! Ty.

The VPA webhook will mutate the pod on admission and the replicaset’s resource requests and not updated (only the pod) so any operator (e.g. Prometheus-operator) will not observe the difference and reconcile.

QuentinBisson:
Yes that's it :)

Haleygo:
Hello, any update on this one)

>you're not taking into accounts sharded Prometheus

By the way, why does this have effect on sharded Prometheus. I assume vpa will only scale those overload prometheus pod. Can you explain a little, thanks! @QuentinBisson

jdheyburn:
It's a shame this went to stale, how can we get it over the line?

PR #3099, opened 2023-03-06
[prometheus-mysql-exporter] Allow overriding of image registry and add quay.io default

jnoordsij:

Allow setting `image.registry` value to specify the registy the image should be pulled from.
Also makes the default to be `quay.io`, which seems to be more consistent with various other charts in this repository.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/values.yaml
+++ b/charts/prometheus-mysql-exporter/values.yaml
@@ -8,7 +8,8 @@ fullnameOverride: ""
 replicaCount: 1
 
 image:
-  repository: "prom/mysqld-exporter"
+  registry: quay.io
+  repository: prometheus/mysqld-exporter
   ## if not set charts appVersion var is used
   tag: ""
   pullPolicy: "IfNotPresent"
--- a/charts/prometheus-mysql-exporter/Chart.yaml
+++ b/charts/prometheus-mysql-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for prometheus mysql exporter with cloudsqlproxy
 name: prometheus-mysql-exporter
-version: 1.12.1
+version: 1.13.0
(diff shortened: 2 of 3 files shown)


charts/prometheus-mysql-exporter/values.yaml
```diff
@@ -8,7 +8,8 @@ fullnameOverride: ""
 replicaCount: 1
 
 image:
-  repository: "prom/mysqld-exporter"
+  registry: quay.io
+  repository: prometheus/mysqld-exporter
```
monotek, line 12: when we change default of an existing variable we should raise major version and add some info in the upgrading section of the chart.

monotek (changes requested):

charts/prometheus-mysql-exporter/values.yaml
jnoordsij: I modelled partially after #928 which did basically the same for the `alertmanager` chart in a minor bump. The adding of the `registry` key does make this slightly different though. You really think a major version bump is the way to go here? Then I'll add some notes to the `README`.

monotek (approved):

Jeroen0494:
This PR breaks my current installation:
```
  Warning  Failed     5m54s (x4 over 7m37s)   kubelet            Failed to pull image "quay.io/prom/mysqld-exporter:v0.14.0": rpc error: code = Unknown desc = failed to pull and unpack image "quay.io/prom/mysqld-exporter:v0.14.0": failed to resolve reference "quay.io/prom/mysqld-exporter:v0.14.0": pulling from host quay.io failed with status code [manifests v0.14.0]: 401 UNAUTHORIZED
```

Are you sure the path to the image is correct?

Jeroen0494:
Seems this PR required fixes on the user side. My repository was by default set to `prom/mysqld-exporter` instead of `prometheus/mysqld-exporter`.

PR #3123, opened 2023-03-15
[prometheus-node-exporter] remove prometheus scrape annotation

mhowell24:

By default the prometheus-node-exporter chart has the `prometheus.io/scrape` annotaiton set to true on the service. If you enable the serviceMonitor or podMonitor that annotation is not removed and you end up with double scraped metrics unless you overwrite the annotation to false. This PR mainains the default behaviour of having the scrape annotation however, it conditionally removes the scrape annotation if you enabled the serviceMonitor or podMonitor

- fixes #

#### Checklist

- [x] Chart Version bumped


charts/prometheus-node-exporter/templates/service.yaml
```diff
@@ -5,10 +5,18 @@ metadata:
...
   {{- with .Values.service.annotations }}
   annotations:
+    {{- unset . "prometheus.io/scrape" | toYaml | nindent 4 }}
+  {{- end }}
+  {{- else }}
+  annotations:
+    prometheus.io/scrape: "true"
+  {{- with .Values.service.annotations }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
```
zanhsieh, line 18: Is this really necessary?

zanhsieh (changes requested):

charts/prometheus-node-exporter/templates/service.yaml
mhowell24: That with block is required since the default value for `service.annotations` is now an empty map which could cause issues if the user does not set any annotations

zanhsieh: @mhowell24 
I mean, if `service.annotations` is empty, e.g. `{}`, then that will render like this:

```yaml
  annotations:
    prometheus.io/scrape: "true"
    {}                          # <== ???
```

I really don't think that is necessary.

zanhsieh (changes requested):

zanhsieh (changes requested):
.

charts/prometheus-node-exporter/templates/service.yaml
mhowell24: We need it for when `.Values.service.annotations` is empty. If we just use the value then when it is empty we try templating an empty map and you end up with invalid yaml/json.

 For example removing the with and updating to this

Screenshot 2023-03-17 at 09 21 17

results in this error when there are no annotations specified

Screenshot 2023-03-17 at 09 22 18


Using the with allows us to only template in the user specific annotations if there are any.

mhowell24: @zanhsieh what do you think or is there some other way you would prefer I deal with that?

zanhsieh: @mhowell24 I mean like this:

```yaml
  ...
  labels:
    {{- include "prometheus-node-exporter.labels" $ | nindent 4 }}
  annotations:
  {{- if or .Values.prometheus.monitor.enabled .Values.prometheus.podMonitor.enabled }}
  {{- with .Values.service.annotations }}
    {{- unset . "prometheus.io/scrape" | toYaml | nindent 4 }}
  {{- end }}
  {{- else }}
    prometheus.io/scrape: "true"
  {{- end }}
  ...
```

zanhsieh (changes requested):

charts/prometheus-node-exporter/templates/service.yaml
mhowell24: That stops the user from being able to add additional annotations to the service if they are not using the `podMonitor` or `serviceMonitor`, which is a valid use case. We also need the with block to contain the `annotations` heading so it only gets added if there are annotaions to be added.

I am using the with block on the else branch of that if statement to only render user speficied annotations if there are any. If there are not any specified then we end up passing an empty map to `toYaml` and helm fails to template, as I showed in my above comment. We need with blocks in both branches of the if statement to cover the case of the user specifying and not specifying annotions.

zanhsieh (approved):

zeritti (approved):

mpostument:
Can this be reverted by any chance, without ability to set prometheus.io/scrape: "false" we are getting metrics duplicates 

https://github.com/prometheus-community/helm-charts/issues/3517

PR #3147, opened 2023-03-22
[alertmanager] Disable automountServiceAccountToken

### LINKED ISSUE #3146: [alertmanager] Disable automountServiceAccountToken

By default service account tokens are automatically mounted. Some services don't need a service account token and should therefore not mount a service account token. This can be disable on a service account and pod.

Disable automountServiceAccountToken by default

NONE

### Additional context.

_No response_

stgrace:

By default service account tokens are automatically mounted. Some services don't need a service account token and should therefore not mount a service account token. This can be disable on a service account and pod.

- fixes #3146

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -25,6 +25,8 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+automountServiceAccountToken: false
+
 serviceAccount:
   # Specifies whether a service account should be created
   create: true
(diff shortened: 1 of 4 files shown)


charts/alertmanager/values.yaml
```diff
@@ -25,6 +25,8 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+automountServiceAccountToken: false
```
monotek, line 28: i guess to not change the current default this should be kept "true".

monotek (changes requested):

monotek (approved):

PR #3151, opened 2023-03-24
[kube-prometheus-stack] adding tpl upstream for secretFieldSelector

z-anshun:
Currently, the "secretFieldSelector" option only supports setting constants. But in more cases we want to be able to set variables (eg `metadata.namespace={{ .Release.Namespace }}`). 

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -108,7 +108,7 @@ spec:
             - --thanos-ruler-instance-namespaces={{ .Values.prometheusOperator.thanosRulerInstanceNamespaces | join "," }}
             {{- end }}
             {{- if .Values.prometheusOperator.secretFieldSelector }}
-            - --secret-field-selector={{ .Values.prometheusOperator.secretFieldSelector }}
+            - --secret-field-selector={{ tpl (.Values.prometheusOperator.secretFieldSelector) $ }}
             {{- end }}
             {{- if .Values.prometheusOperator.clusterDomain }}
             - --cluster-domain={{ .Values.prometheusOperator.clusterDomain }}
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -108,7 +108,7 @@ spec:
             - --thanos-ruler-instance-namespaces={{ .Values.prometheusOperator.thanosRulerInstanceNamespaces | join "," }}
             {{- end }}
             {{- if .Values.prometheusOperator.secretFieldSelector }}
-            - --secret-field-selector={{ .Values.prometheusOperator.secretFieldSelector }}
+            - --secret-field-selector={{ tpl (.Values.prometheusOperator.secretFieldSelector) . }}
```
QuentinBisson, line 111:
```suggestion
            - --secret-field-selector={{ tpl (.Values.prometheusOperator.secretFieldSelector) $ }}
```

QuentinBisson:
This needs a version bump

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
z-anshun: Can you tell me why '$' is recommended in this way? I found the '.' used in the helm documentation.
https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function

QuentinBisson: Sure, `.` is the current scope where `$` is the root scope so if you are in a range or a with, the current scope changes so .Values does not exist in there while `$` allows you to access the root scope

cf. https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

QuentinBisson: Also the other tpl are using the $ in this file so for homogeneity, let's use the same scheme

QuentinBisson (approved):

PR #3165, opened 2023-03-28
fix: correct the readiness and liveness probe addr

eminaktas:

The address for liveness and readiness was wrong. This PR fixes it. You can check the correct endpoint [here](https://github.com/prometheus/blackbox_exporter/blob/master/main.go#L173).

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/values.yaml
+++ b/charts/prometheus-blackbox-exporter/values.yaml
@@ -105,12 +105,12 @@ securityContext:
 
 livenessProbe:
   httpGet:
-    path: /health
+    path: /-/healthy
     port: http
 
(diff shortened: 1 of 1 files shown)


monotek (changes requested):
please raise chart version.
imho we should aslo use an higher failureThreshold for the livenessprobe so the probes are not the same.

eminaktas:
> please raise chart version. imho we should aslo use an higher failureThreshold for the livenessprobe so the probes are not the same.

Thanks for the review. I just added as you mentioned. is 3 sec ok? wdyt?

monotek (approved):

PR #3173, opened 2023-03-30
[prometheus-postgres-exporter] Add comments regarding excludeDatabases and includeDatabases

crisbal:

I have spent some time trying to figure out why these variables were not being recognized, so I think a comment was necessary. Especially because in the HELM chart this is not as clear as one would like: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-postgres-exporter/templates/deployment.yaml#L66

- clarifies the usage of `excludeDatabases` and `includeDatabases` variables

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.11.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.4.0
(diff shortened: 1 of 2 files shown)


zanhsieh:
@crisbal 
Thanks for the contribution. The version in Chart.yaml is still required to bump to latest, no matter just adding comment. Without it, CI won't let it pass and cannot merge.

crisbal:
Hi @zanhsieh I have now bumped the version of the chart

zanhsieh (approved):

PR #3176, opened 2023-03-31
[prometheus-node-exporter] add attachMetadata to ServiceMonitor

xanecs:
This PR allows setting the "attachMetadata > node" property of the ServiceMonitor created by the prometheus-node-exporter chart, the same way it is already possible for the PodMonitor.

We want to query metrics based on a label attached to the node but keep using the ServiceMonitor (as is the default when using kube-prometheus-stack).

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.14.0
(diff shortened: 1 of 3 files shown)


zanhsieh (changes requested):
@xanecs 
Thanks for the contribution. Could you bump the version in Chart.yaml to latest please? Without version bump, the CI won't allow this PR to merge.

PR #3178, opened 2023-03-31
[kube-state-metrics] Add annotations to service-monitor object

spaghettifunk:
This PR introduces the `annotations` metadata in the `ServiceMonitor` object. It is needed in case users want to add extra annotations to the object. For example, when using ArgoCD, you need to add extra [annotations ](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#skip-dry-run-for-new-custom-resources-types) for more `SyncOptions`

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -54,8 +54,7 @@ service:
   annotations: {}
 
 ## Additional labels to add to all resources
-customLabels:
-  {}
+customLabels: {}
   # app: kube-state-metrics
 
 ## Override selector labels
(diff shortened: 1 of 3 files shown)


charts/kube-state-metrics/values.yaml
```diff
@@ -54,7 +54,8 @@ service:
   annotations: {}
 
 ## Additional labels to add to all resources
-customLabels: {}
+customLabels:
```
mrueg, line 57: can you remove the formatting changes that are not necessary for your change?

spaghettifunk: done 👍

PR #3190, opened 2023-04-04
[prometheus-node-exporter] Patch nodeSelector on node-exporter daemonset

marciogmorales:

On a heterogenous Kubernetes cluster with Windows nodes, the daemonset pods try to be scheduled on all nodes as it doesn't have nodeSelector by OS type, and Linux is the only one supported for node-exporter.

This proposed change adds the kubernetes.io/os: Linux as nodeSelector in order to avoid failed loop on pod scheduling on Windows nodes.

- fixes #

I'm starting addressing the node-exporter daemonset, however the same fix will be needed on other prometheus deployments, which I'll work on once this PR is merged to the main.

#### Checklist

- [X] Chart Version bumped


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -39,6 +39,8 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "prometheus-node-exporter.serviceAccountName" . }}
+      nodeSelector:
+        kubernetes.io/os: {{ .Values.nodeselector.os }}
```
zanhsieh, line 43: I would let the users picks their own generic purpose spec here:

```suggestion
      {{- with .Values.nodeselector }}
      nodeSelector:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

charts/prometheus-node-exporter/values.yaml
```diff
@@ -434,3 +434,8 @@ verticalPodAutoscaler:
     # Specifies whether recommended updates are applied when a Pod is started and whether recommended updates
     # are applied during the life of a Pod. Possible values are "Off", "Initial", "Recreate", and "Auto".
     # updateMode: Auto
+    
+ ## NodeSelector
+ # It forces prometheus-node-exporter to be only scheduled on Linux-based worker nodes as it isn't compatible with Windows OS.
+nodeselector:
+  os: linux
```
zanhsieh, line 441: Set to `{}` as backward compatible.
```suggestion
nodeselector: {}
#   kubernetes.io/os: linux
```

zanhsieh (changes requested):
@marciogmorales 
Thanks for the contribution. Beside the change I suggested, would you mind to bump the version in Chart.yaml so the CI can merge this PR please?

marciogmorales:
@zanhsieh I think we will need to force the nodeSelector to Linux OS and not let it as optional value, because the Chart deployment fails on an heterogeneous cluster. Thoughts?

zanhsieh:
> @zanhsieh I think we will need to force the nodeSelector to Linux OS and not let it as optional value, because the Chart deployment fails on an heterogeneous cluster. Thoughts?

As long as that's backward compatible, I am fine with it.

marciogmorales:
> 

Should I re-write the code as I did in the first modification?

marciogmorales (commented):
Fixed several inconsistent that parameters were referenced to wrong values. Added nodeSelector to be mapped to Linux instead of an optional value.

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -220,18 +225,11 @@ spec:
...
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.nodeSelector }}
-      nodeSelector:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
-      {{- with .Values.tolerations }}
-      tolerations:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
```
zanhsieh, line 234: @marciogmorales 
Could you put back the `tolerations` please?

zanhsieh (changes requested):

marciogmorales (commented):
Removing empty 278 line

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -39,6 +39,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "prometheus-node-exporter.serviceAccountName" . }}
+      {{- with .Values.nodeselector }}
```
zanhsieh, line 42: @marciogmorales 
You miss an `{{- end }}` in line 278 if you'd like to keep `{{- with .Values.nodeselector }}`.

zanhsieh (changes requested):

marciogmorales:
Thank you @zanhsieh . Fixed the issue.

jcpunk:
I'll confess some interest in this.

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -228,15 +229,15 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.dnsConfig }}
+      {{- with .Values.affinity }}
```
jcpunk, line 232: These `with` blocks look wrong to me...

PR #3191, opened 2023-04-05
[kube-prometheus-stack] Remove deprecated sha in prometheusSpec

sathieu:
See https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.AlertmanagerSpec

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 45.26.0
+version: 45.27.0
 appVersion: v0.65.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


GMartinez-Sisti:
Was late to this PR. This should also be removed: https://github.com/prometheus-community/helm-charts/blob/ec2327466d6be8826da6ff1d813c3d657aa4aeae/charts/kube-prometheus-stack/values.yaml#L2749

sathieu:
@GMartinez-Sisti I don't think so. It's still used.

https://github.com/prometheus-community/helm-charts/blob/ec2327466d6be8826da6ff1d813c3d657aa4aeae/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L34-L46

However, I plan to remove all `sha` usage (like https://github.com/prometheus-community/helm-charts/pull/3248, https://github.com/prometheus-community/helm-charts/pull/3130).

GMartinez-Sisti:
Got it. It's still being used indeed.

I understand your concern on using the right terms (sha vs digest), however bumping a major version and forcing users to change their configs for something that actually doesn't change the chart doesn't feel productive.

sathieu:
@GMartinez-Sisti I understand your concern. In #3130, I only updated revision, which was not correct according to semver, but only users using `sha` were affected (only a few I think). In #3248, I bumped major.

In both cases, the Helm chart fails when `sha` is used. This prevent silent ignore.

PR #3194, opened 2023-04-05
[kube-state-metrics] set parameters for podsecurity restricted

jcpunk:

This permits installing this change in a namespace with PodSecurity set to `restricted`.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 5.3.0
+version: 5.4.0
 appVersion: 2.8.2
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -214,7 +214,14 @@ securityContext:
   enabled: true
   runAsGroup: 65534
   runAsUser: 65534
+  runAsNonRoot: true
   fsGroup: 65534
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+    - ALL
+  seccompProfile:
+    type: RuntimeDefault
 
 ## Specify security settings for a Container
 ## Allows overrides and additional options compared to (Pod) securityContext


dotdc:
@jcpunk, does this change require https://github.com/kubernetes/kube-state-metrics/pull/2042 to be merged ? 

@mrueg, do you know if this is a breaking change?

Not sure if the minor or major version should be bumped, but probably not patch.

jcpunk:
This patch does not require the upstream one to be merged.  It would be nice to have up there, but not strictly required.

mrueg:
I wouldn't consider this change breaking, but would prefer to raise the minor version as it changes some defaults.

dotdc:
@jcpunk Can you bump the minor version instead of the patch?

mrueg (approved):

georgekaz:
For me this merge causes: 
```
error validating data: [ValidationError(Deployment.spec.template.spec.securityContext): unknown field "allowPrivilegeEscalation" in io.k8s.api.core.v1.PodSecurityContext, ValidationError(Deployment.spec.template.spec.securityContext): unknown field "capabilities" in io.k8s.api.core.v1.PodSecurityContext]
```

I believe capabilities is only on container level securitycontext

atarax:
Yes, definitely broken after the merge, seeing the same as @georgekaz

dotdc:
Can you share your respective Kubernetes versions @jcpunk, @georgekaz and @atarax?

Will try to look at this when I can, in the meantime you can use the previous version.

georgekaz:
1.26 for me

ramdaspotale:
it fails on v1.24.9 as well.

atarax:
it's not related to the version, like @georgekaz mentioned, capabilities is a container-level setting

ramdaspotale:
@atarax agreed. and @dotdc i believe chart can be improved to add allowPrivilegeEscalation and capabilities fields to container level specs instead of rolling back changes?

jcpunk:
In theory I've got another patch that does this the right way now.  In practice, they both seem to install on my test cluster, so I suspect my test cluster is super messed up....

https://github.com/prometheus-community/helm-charts/pull/3235

PR #3203, opened 2023-04-07
[alertmanager] Add annotations for ConfigMap

jidckii:

I added annotations to configmap to be able to mutate secrets from vault. The config often contains private tokens or links and you don’t want to publish them to git when using, for example, argocd.
An example of use described.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -168,6 +168,17 @@ persistence:
     - ReadWriteOnce
   size: 50Mi
 
+configAnnotations: {}
+  ## For example if you want to provide private data from a secret vault
+  ## https://github.com/banzaicloud/bank-vaults/tree/main/charts/vault-secrets-webhook
+  ## P.s.: Add option `configMapMutation: true` for vault-secrets-webhook
+  # vault.security.banzaicloud.io/vault-role: "admin"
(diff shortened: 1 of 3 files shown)


naseemkullah (dismissed):

charts/alertmanager/values.yaml
```diff
@@ -168,9 +168,18 @@ persistence:
...
+  ## P.s.: Add option `configMapMutation: true` for vault-secrets-webhook
+  # vault.security.banzaicloud.io/vault-role: "admin"
+  # vault.security.banzaicloud.io/vault-addr: "https://vault.vault.svc.cluster.local:8200"
+  # vault.security.banzaicloud.io/vault-skip-verify: "true"
+  # vault.security.banzaicloud.io/vault-path: "kubernetes"
+
 config:
   global: {}
-    # slack_api_url: ''
+    # slack_api_url: '${vault:secret/data/slack-hook-alerts#URL}'
```
monotek, line 182: This seems unintended

monotek (changes requested):

charts/alertmanager/values.yaml
jidckii: No, it's intentional, use case.

monotek: most people will not use this and therefore i woulld not add this as default example.

monotek (approved):

PR #3206, opened 2023-04-08
Broken Link Fix

onurkanbakirci:
  Broken link fix.

- https://kubernetes.io/docs/concepts/services-networking/service/#external-ips should be used instead https://kubernetes.io/docs/user-guide/services/#external-ips

### DIFF
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -566,7 +566,7 @@ server:
     clusterIP: ""
 
     ## List of IP addresses at which the Prometheus server service is available
-    ## Ref: https://kubernetes.io/docs/user-guide/services/#external-ips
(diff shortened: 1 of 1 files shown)


zanhsieh (changes requested):
@onurkanbakirci 
Thanks for fixing the broken link. Would you mind to bump the version in Chart.yaml so the CI would let this pass please? Also please sign DCO.

onurkanbakirci:
Hi @zanhsieh . Can you check it?

zanhsieh:
> Hi @zanhsieh . Can you check it?

@onurkanbakirci 
DCO still has problem:

https://github.com/prometheus-community/helm-charts/pull/3206/checks?check_run_id=12620804174

PR #3208, opened 2023-04-09
[prometheus-operator-admission-webhook] Add chart prometheus-operator…

zeritti:
…-admission-webhook

I would like to make chart *prometheus-operator-admission-webhook* available through the Prometheus community to users of the community's charts.

Currently, the *kube-prometheus-stack* chart brings the functionality with Prometheus operator as the admission webhook runs within its code. This is good for development and testing. In a production environment, though, a standalone webhook is recommended.

The Prometheus operator project has made the admission webhook available as a standalone application ([#4494](https://github.com/prometheus-operator/prometheus-operator/pull/4494)) since release 0.55.0 which can thus be deployed independently. This brings benefits such as making it highly available through multiple replicas, making it independent of changes that might be affecting the operator's runtime, deploying it in a different namespace, etc.

This chart draws on the *kube-prometheus-stack* chart and how the admission webhook's resources have been implemented. 

In this chart:
- With the default values, a single webhook replica will be launched with TLS resources produced by the hooks (a secret and patched webhook configurations) 
- Beside mutating and validating `Prometheusrules`, validating `Alertmanagerconfigs` has also been enabled
- Mandatory TLS resources for the webhook are made available through hooks (jobs),
  [cert-manager](https://cert-manager.io/docs/getting-started/) and a custom TLS keypair from a secret through `tlsSecret` with or without `caBundle`
- Security context of all the containers is set to satisfy the Restricted profile of the [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- Webhook parameters underneath `webhooks` are currently set globally on the created webhooks
- Support for Prometheus rules and service monitor is present
- RBAC is enabled by default
- Image registry and repository are set via separate values
- A values schema has been added for clarity

None

With respect to versioning, I would suggest usual versioning of the chart following semver providing thus means to indicate e.g. breaking changes (beginning with 0.1.0). Mapping the admission webhook's version to the Prometheus operator's version (both should be in sync, indeed) won't be as practical as one would probably desire, though.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
+++ b/charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
@@ -62,17 +62,6 @@ Create the name of the service account to use
 {{- end -}}
 
 {{/*
-Return the appropriate apiVersion for rbac.
-*/}}
-{{- define "rbac.apiVersion" -}}
-{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" -}}
-{{- print "rbac.authorization.k8s.io/v1" -}}
-{{- else -}}
-{{- print "rbac.authorization.k8s.io/v1beta1" -}}
-{{- end -}}
-{{- end -}}
-
-{{/*
 Define overriding namespace
 */}}
 {{- define "prometheus-operator-admission-webhook.namespace" -}}
--- a/charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
+++ b/charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
@@ -84,17 +84,6 @@ Define overriding namespace
 {{- end -}}
 
 {{/*
-Define Pdb apiVersion
-*/}}
-{{- define "pdb.apiVersion" -}}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
-{{- printf "policy/v1" }}
-{{- else }}
-{{- printf "policy/v1beta1" }}
-{{- end -}}
-{{- end -}}
-
-{{/*
 Define image tag by attaching "v" to the application version
 preferring to keep the application version according to its release
 */}}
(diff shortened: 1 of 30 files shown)


charts/prometheus-operator-admission-webhook/values.yaml
```diff
@@ -0,0 +1,349 @@
...
+    digest: ""
+    pullPolicy: Always
+    pullSecrets: []
+
+  ## priorityClassName for pods
+  priorityClassName: ""
+
+  ## resources for job containers
+  ## Ref. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+  resources:
```
monotek, line 75: normally we don't set resources by default.

charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
```diff
@@ -0,0 +1,107 @@
...
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for rbac.
+*/}}
+{{- define "rbac.apiVersion" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" -}}
+{{- print "rbac.authorization.k8s.io/v1" -}}
+{{- else -}}
+{{- print "rbac.authorization.k8s.io/v1beta1" -}}
```
monotek, line 71: imho the old api version is not needed anymore as the k8s version supporting it is eol anyway

charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
```diff
@@ -0,0 +1,107 @@
...
+    {{- .Values.namespaceOverride -}}
+  {{- else -}}
+    {{- .Release.Namespace -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+Define Pdb apiVersion
+*/}}
+{{- define "pdb.apiVersion" -}}
```
monotek, line 89: imho the old api version is not needed anymore as the k8s version supporting it is eol anyway

monotek (changes requested):

charts/prometheus-operator-admission-webhook/templates/_helpers.tpl
zeritti: The defined template is no longer used.

monotek (approved):

PR #3216, opened 2023-04-11
[prometheus-blackbox-exporter] Add ability to define metricRelabelings action

mkilchhofer:

Recent Prometheus-Operator versions implicit set the `action` field inside `.spec.endpoints.metricRelabelings` to `replace`. This isn't convenient when doing GitOps eg. via Argo CD. The GitOps tool then always shows a diff to apply:
image

The issue is easy to fix by explicitly setting the same value inside the Git repo so that the GitOps tools doesn't detect a diff. Altough I assume that we cannot "hardcode" it as it would require Prometheus instance >= 2.36.

We did not open an issue yet.

`---`

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.7.0
+version: 7.8.0
 appVersion: 0.23.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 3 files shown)


monotek (changes requested):
action replace is the default and it's quite common to not set defaults explicitly.
I don't think we should change this, just to get ArgoCD to work.
I i were you, i would even consider to switch to another GitOps tool, like FluxCD, as this will not be the last shortcoming you'll experience with ArgoCD and Helm charts.

mkilchhofer:
IMHO this has nothing to do to with Argo vs. Flux. I know that this can be workarounded by enabling SSA (Server side apply) in Argo CD which is default in Flux. But there are also reasons why SSA is not enabled.

I think it doesn't hurt to allow users to experiment with the `action` field.

PR #3217, opened 2023-04-11
[prometheus-blackbox-exporter] use well known label as namespace label selector

PG2000:

Kubernetes uses a well known label for namespaces. (kubernetes.io/metadata.name=)
Would be great if the network policy can match this label then.

none
#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
i don't see why this should be the default, breaking the chart for everyone not having this label.

PG2000:
> i don't see why this should be the default, breaking the chart for everyone not having this label.

Sorry i missed to give some context.
I did this based on https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name

QuentinBisson:
@monotek this label is set by default on all namespaces since at least kubernetes 1.22 (c.f. https://v1-22.docs.kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name)

@PG2000 I would suggest you add a kubernetes version flag so the chart uses the old way prior to 1.22 and the new way after 1.22. 

Would that work for the both of you?

monotek:
I would just make the labels generical configurable, using the old labels as default.

PG2000:
> I would just make the labels generical configurable, using the old labels as default.

I did the mentioned label selector configurable.

Just from my point of view:
The new label makes this work out of the box for everyone with a version > 1.22 

I didn't get why this should be backwards compatible for end of life k8s versions.


From my point of view: 
For this you have some semantic versioning of the helm chart and more effort to maintain this with all the stuff just to be backwards compatible.

QuentinBisson (approved):

QuentinBisson:
@PG2000 can you bump the version?

PR #3245, opened 2023-04-17
[alertmanager] Fixing "cannot overwrite table with non table" warning

DandyDeveloper:

When overriding either of these fields we get the following warning:

```
coalesce.go:220: warning: cannot overwrite table with non table for monitoring.alertmanager.hostAliases (map[])
```

It's a non-issue but a bit frustrating to encounter. 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -41,7 +41,7 @@ priorityClassName: ""
 
 podSecurityContext:
   fsGroup: 65534
-dnsConfig: {}
+dnsConfig: []
   # nameservers:
   #   - 1.2.3.4
   # searches:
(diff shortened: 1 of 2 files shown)


charts/alertmanager/values.yaml
```diff
@@ -41,7 +41,7 @@ priorityClassName: ""
 
 podSecurityContext:
   fsGroup: 65534
-dnsConfig: {}
+dnsConfig: []
```
monotek, line 44: this does not look like a lst, as the first comment starts with: "# nameservers:"

monotek (changes requested):

charts/alertmanager/values.yaml
DandyDeveloper: Yeah, not sure why I did this. Sorry, hostAliases is the only impacted service. I'll amend.

QuentinBisson (approved):

QuentinBisson:
@monotek do you mind merging?

QuentinBisson (approved):

monotek (approved):

Issue #2833, opened 2022-12-20
[kube-prometheus-stack] grafana dashboards in seperate chart


jcpunk:
### Is your feature request related to a problem ?

Tracking the dashboard states is currently buried within the operator chart.  Having the dashboards in their own chart would also give non-operator charts a way to pull in the dashboards cleanly.

### Describe the solution you'd like.

Separate the `kube-prometheus-stack/templates/grafana/dashboards-1.14` charts into a separate chart that is pulled in optionally dependent on which flags are set.

### Describe alternatives you've considered.

Maintaining my own smaller chart with just the dashboards

### Additional context.

Having the dashboards mixed in with the operator isn't as clean as having a chart focused on just the operator and another on just the dashboards.

IDerr:
I'm having the same issue, I don't really know how to manage my monitoring dashboards with this chart

zeritti:
> I'm having the same issue, I don't really know how to manage my monitoring dashboards with this chart,
> 
> Perhaps, we can add a value to add dashboard in the values.yml to make it simple ?

The Grafana chart supports [importing](https://github.com/grafana/helm-charts/blob/3c1ca3991c46c90fada061d67bbab26095b49081/charts/grafana/values.yaml#L636) dashboards which can then be set in your values file through `grafana.dashboards`. I used such an import successfully, although, I admit, for numerous  dashboards I ended up setting up a custom Helm chart solely with dashboards.

jcpunk:
This is still something I'd like to see

stoupance:
This would be very handy in lots of situation.

jcpunk:
I'd still like to see this

jcpunk:
I'd still like to see this resolved.

bsgrigorov:
any update on this?

nicolasdonato:
Hello,

I'd like to see the capability to define grafana dashboard/datasource in others chart.

I saw the [grafana operator](https://grafana.github.io/grafana-operator/) which can help to define a dashboard by using CRD (https://grafana.github.io/grafana-operator/docs/examples/dashboard_from_grafana_com/readme/).

Can we use it into the kube-prometheus-stack?

Thank you in advance.

PR #3267, opened 2023-04-20
[kube-prometheus-stack] update kps to v0.65.1

### LINKED ISSUE #3214: [kube-prometheus-stack] update to 0.64

The prometheus operator is now on version 0.64, it would be nice if this chart followed.

update CRDs to 0.64

Using another chart

### Additional context.

https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.64.0

QuentinBisson:

Upgrades prometheus-operator to v0.65.1

- fixes #3214

#### Checklist

- [x] Chart Version bumped

@monotek and @GMartinez-Sisti as it's my first major release, I'd like if you could take a look, especially as this one adds a new crd


GMartinez-Sisti (approved):

QuentinBisson:
Well this will not be fun to fix:

```
Error: INSTALLATION FAILED: create: failed to create: Secret "sh.helm.release.v1.kube-prometheus-stack-l5sr2p6xtu.v1" is invalid: data: Too long: must have at most 1048576 bytes
```

QuentinBisson:
So the issue is coming from the newer PrometheusAgent CRD which is causing the helm secret to be too big.

The only solution I can think of is to have the crds outside the chart but maybe you can think of a better solution?

If that is fine with you i was thinking of doing something similar to https://github.com/cowboysysop/charts/tree/master/charts/vertical-pod-autoscaler/templates/crds a job that pulls and would then also update the crds

zeritti:
> The only solution I can think of is to have the crds outside the chart but maybe you can think of a better solution?

An immediate/short-term solution might be using stripped-down crds as released by the Prometheus operator team. These do not include description fields, the manifests are considerably smaller.

QuentinBisson:
That could work but this would reduce functionality as users would not be able to use kubectl explain but also, as you mention, this is only short-term. With the upcoming ScrapeConfig and maybe Silence CRD in prometheus operator, I know this issue will happen again "soon" so I would rather we have a proper fix :)

monotek (approved):

monotek (changes requested):
Can you please check for the failing ci?

QuentinBisson:
@monotek i'm trying to fix it but the issue is that the New prometheus agent crd make us go over the 1MB secret limit of Kubernetes

QuentinBisson:
@GMartinez-Sisti and @monotek I tested this on existing clusters but this is quite a big change so I won't merge with newer approvals

charts/kube-prometheus-stack/templates/crds/job.yaml
```diff
@@ -0,0 +1,73 @@
+{{- if .Values.crds.enabled }}
+{{- $appVersion := .Chart.AppVersion }}
+{{- $crds := list "alertmanagerconfigs" "alertmanagers" "podmonitors" "probes" "prometheusagents" "prometheuses" "prometheusrules" "servicemonitors" "thanosrulers" -}}
```
GMartinez-Sisti, line 3: Can we convert this to a config value inside the new crds key? This way users can be granular on what CRDs to manage.

charts/kube-prometheus-stack/templates/crds/job.yaml
```diff
@@ -0,0 +1,73 @@
...
+          {{- else }}
+          image: {{ $registry }}/{{ $.Values.crds.image.repository }}:{{ $.Values.crds.image.tag }}
+          {{- end }}
+          imagePullPolicy: {{ $.Values.crds.image.pullPolicy }}
+          args:
+            - apply
+            - --server-side
+            - --force-conflicts={{ $.Values.crds.forceConflicts }}
+            - -f
+            - https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/{{ $appVersion }}/example/prometheus-operator-crd/monitoring.coreos.com_{{ . }}.yaml
```
GMartinez-Sisti, line 49: I would also allow this URL to be customisable as some environments might not have external internet access and this can be easily checked into an internal artefact repository.

GMartinez-Sisti:
> @GMartinez-Sisti and @monotek I tested this on existing clusters but this is quite a big change so I won't merge with newer approvals

Ideally we would have CI automation to deploy the version on main and then the PR one over it. I can test this but it would take a couple of days. The logic looks good though, I like this approach 💪 .

Left some comments.

hoerup:
Looks like they are working on a 0.64.1 bugfix release
https://github.com/prometheus-operator/prometheus-operator/pull/5524

monotek (changes requested):
Please remove the change of crd logic. 

If that needs a change we probably would do it by using the existing crd chart.

QuentinBisson:
@monotek I can remove the job for sure but then the chart become unusable with the crds included. What would you do here instead?

monotek:
> @monotek I can remove the job for sure but then the chart become unusable with the crds included. What would you do here instead?

use https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds as dependency.

QuentinBisson:
@monotek I've tried your suggestion and it is not working because we cannot change the helm resource installation order

As the CRDs are not installed from the `crds` directory they are applied last so even after Custom Resources like ServiceMonitors.

I see 2 possible solutions but maybe you have more that would be nice:
1 - We have the job to install CRDs
2 - We remove the CRDs without the job and add a big message that users need to install the crds on their own

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "kube-prometheus-stack-alertmanager" namespace: "default" from "": no matches for kind "Alertmanager" in version "monitoring.coreos.com/v1"
ensure CRDs are installed first, resource mapping not found for name: "kube-prometheus-stack-prometheus" namespace: "default" from "": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1"
```

zeritti:
> > @monotek I can remove the job for sure but then the chart become unusable with the crds included. What would you do here instead?
> 
> use https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds as dependency.

That would be the simplest way forward, indeed. However, aren't subcharts' manifests included in the release secret? As far as I can see, they are. 
There's also a danger in ordering - there is no way to order deployments of subcharts through dependencies which might result in creating a CR without the corresponding CRD being in place. Please, correct me if I've got it wrong.

QuentinBisson:
@zeritti you're are right about ordering as I tested it :D

monotek (changes requested):
If it does not work with the dependency chart i would keep crd handlin as it is and just update prometheus(operator) components.

QuentinBisson:
So you would not add the prometheus-agent crd? Because if we add it, helm cannot install the release but I do not think this is a good solution

monotek:
> So you would not add the prometheus-agent crd? Because if we add it, helm cannot install the release but I do not think this is a good solution

No, i would not change the way the CRDs are handled.

QuentinBisson:
But then you are missing out on the newer CRDs

monotek:
i'm talking about the job

QuentinBisson:
Yes but then, do I still remove the CRDs from the app and we mention in the readme that users need to install it on their own? Because the CRDs being in the app go over the kubernetes secret limit

jcpunk:
How far over the 1Mb limit is it?  If it is a small amount, could folks work with the prometheus-operator team to shrink the CRD a bit upstream?

If it is over by 100 bytes, some rewording of the descriptions may do the trick.  If by 10k, well that probably wont work....

QuentinBisson:
They could probably shrink them, but there are a few CRDs coming (ScrapeConfig CRD, Silence CRD) that would just make this a temporary workaround because the rest of the chart (with dashboards, rules and so on) is actually quite big

monotek:
Non of these files is over 1 mb: https://github.com/prometheus-operator/prometheus-operator/tree/main/example/prometheus-operator-crd

1MB is the limit of etcd. 
If your resource would be bigger it could not be applied at all to any k8s cluster.

The limit you're speaking of is the size of the kubectl.kubernetes.io/last-applied-configuration annotation.
Annotation limit is 256kb / 253 characters in total, followed by a slash: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
But this annotation is only used by client side apply.
Thats the reasone we have "--server-side" added to the kubectl command in the docs, as this does not set this annotation.

rouke-broersma:
@monotek I don't think this is related to the last-applied-configuration limit. The issue seems to be that helm uses one single secret to manage release information. Helm puts the complete (compressed) manifests of a release inside the secret. This secret cannot be larger than 1mb. This new CRD causes the total compressed size of all manifests of kube-prometheus-stack to be larger than 1mb of text. 

See other project with the same problem: https://azure.github.io/azure-service-operator/design/adr-2023-02-helm-chart-size-limitations/

Helm suggested solution is to use sql as release storage backend: https://helm.sh/docs/topics/advanced/#sql-storage-backend
But that would force all kube prometheus stack helm users to switch helm storage backend.

monotek:
Ah, ok. sorry, did not get that before.
But in this case i'm out and won't review this.
I'm not a user of this chart or prometheus anymore so the maintianers should come up with a solution.
Personally i don't think it's a good idea to load stuff from external sources from inside a container.

GMartinez-Sisti:
> Personally i don't think it's a good idea to load stuff from external sources from inside a container.

^^ Agree with this part.

Some ideas to minimise this issue:
- make the source for the files configurable, users of the chart can set their origin for the files if desired
- allow providing checksum for the files

This should increase the confidence of loading the desired files.

rouke-broersma:
> > Personally i don't think it's a good idea to load stuff from external sources from inside a container.
> 
> ^^ Agree with this part.
> 
> Some ideas to minimise this issue:
> - make the source for the files configurable, users of the chart can set their origin for the files if desired
> - allow providing checksum for the files
> 
> This should increase the confidence of loading the desired files.

That still sounds like a terrible workaround for a stupid problem tbh. At some point you'll just run into this problem again and at that point you might have run out of files you can move out of the Helm chart.

Imo either the Helm project needs to be convinced to split config in multiple secrets or something, or this chart needs to become smaller.

jcpunk:
Would moving the dashboards into another chart help at all? (https://github.com/prometheus-community/helm-charts/issues/2833)

rouke-broersma:
> Would moving the dashboards into another chart help at all? (https://github.com/prometheus-community/helm-charts/issues/2833)

If the dashboard chart is not made a subchart, probably yes. But that would be a huge breaking change for any user depending on those dashboards.

jcpunk:
```shell
templates $ du -hs *
60K	alertmanager
80K	exporters
1.6M	grafana
12K	_helpers.tpl
4.0K	NOTES.txt
396K	prometheus
120K	prometheus-operator
36K	thanos-ruler
```

The bulk seems to be inside https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/grafana/dashboards-1.14

I don't like breaking things for existing users, but moving those dashboards out to a non-subchart, incrementing the breaking change int, and telling folks to add a second chart might be a way forward.

QuentinBisson:
After the many discussions here, I agree that we need to split the helm chart.

As the crds chart already exists, I would start there (not using the job but asking people to install the crd chart for this) as this would help with crd upgrades and so on

And as a follow up, move the dashboards outside. @GMartinez-Sisti wdyt?

jcpunk:
As a happy consumer of this chart, I see it as the only current helm source of:

* the deployment of the operator container that matches the CRDs
* the example alert rules
* the service monitors for built in services
* the example dashboards

I like the integration with the other charts quite a bit, but, so long as a maintained helm source remains for the parts above, I can adjust my workflows.

I'm not sure how to keep the operator image itself in sync with the CRDs, but if there is doc somewhere I'll read it.

QuentinBisson:
@jcpunk for me, as long as the prometheus-operator-crds app is kept up to date frequently, it's quite easy to do as you would update first the crd chart and then the kube-prometheus-stack chart in your clusters.

The only issue I see is when users will actually migrate to the version without the CRD so we will have to document the process which consists in "hacking" the crds by overwritting labels and annotations on the crds that were set by helm to explain that the new owner is another chart than kube-prometheus-stack but this is something I've sadly had to do multiple times :D 

It can either be done via a job or via a kubectl command

QuentinBisson:
Once we have a community approved solution then we can go ahead :)

QuentinBisson:
@GMartinez-Sisti I added the source override ;)

charts/kube-prometheus-stack/README.md
```diff
@@ -80,9 +82,30 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
 
 A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.
 
+### From 45.x to 46.x
+
+This version upgrades Prometheus-Operator to v0.65.1 with new CRDs (PrometheusAgent and ScrapeConfig), Prometheus to v2.43.1 and Thanos to v0.31.0.
+
+As of this release, CRD installation and upgrade is now managed by the chart if the property `crds.enabled` is set to true (by default).
```
rouke-broersma, line 89: This is confusing because CRD installation and upgrade is already managed by this chart before the change.

charts/kube-prometheus-stack/README.md
```diff
@@ -80,9 +82,30 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
...
+
+If you want to upgrade the CRDs the old fashion way, set `crds.enabled` to false and run these commands to update the CRDs before applying the upgrade.
```
rouke-broersma, line 91: Perhaps link to the CRDs chart instead?

charts/kube-prometheus-stack/README.md
rouke-broersma, line 89: Do you also check for the existing helm property `skipCrds` (can you even do that?). If you can't do that you should mention that users of `skipCRD` need to migrate to `.Values.crds.enabled`. Otherwise users that already don't install the CRDs using this chart might suddenly have their CRDs managed by the chart by accident.

rouke-broersma: Now I realize that changelogs need to be read carefully especially when it concerns a breaking change however the breaking change churn on this chart is so large that I can imagine that users might be a bit changelog-fatigued when it comes to this chart.

QuentinBisson:
@GMartinez-Sisti do you mind taking a look? I'm actually ok with also just moving to the prometheus-operator-crds chart but that needs to be maintained actively

onedr0p:
FWIW I am generally fine with the method described here to install and manage CRDs, however I would like to be sure that this method doesn't apply helm ownership labels and annotations to the CRDs.

My method of bootstrapping new clusters is to apply the prometheus operator CRDs before anything else so that any chart I install doesn't fail with something like `servicemonitor cr not found`. I do however rely on this chart to upgrade the CRDs after everything is bootstrapped into my cluster.

QuentinBisson:
@onedr0p how do you rely on this chart to upgrade your CRDs considering helm never upgrades CRDs if they are set in the crds/ folder? Do you upgrade your CRDs by hand?

Regarding the job method, no it does not apply the help ownership labels as it's basically applying them from the prometheus-operator repository.

Also know that you can not use this feature if you wish :)

onedr0p:
@QuentinBisson Flux does have the ability to upgrade CRDs in the special `crds` directory.

See here: https://fluxcd.io/flux/components/helm/helmreleases/

```
...
	// CRDs upgrade CRDs from the Helm Chart's crds directory according
	// to the CRD upgrade policy provided here. Valid values are `Skip`,
	// `Create` or `CreateReplace`. Default is `Create` and if omitted
	// CRDs are installed but not updated.
	//
	// Skip: do neither install nor replace (update) any CRDs.
	//
	// Create: new CRDs are created, existing CRDs are neither updated nor deleted.
	//
	// CreateReplace: new CRDs are created, existing CRDs are updated (replaced)
	// but not deleted.
	//
	// By default, CRDs are applied (installed) during Helm install action.
	// With this option users can opt-in to CRD replace existing CRDs on Helm
	// install actions, which is not (yet) natively supported by Helm.
	// https://helm.sh/docs/chart_best_practices/custom_resource_definitions.
	//
	// +kubebuilder:validation:Enum=Skip;Create;CreateReplace
	// +optional
	CRDs CRDsPolicy `json:"crds,omitempty"`
...
```

rouke-broersma:
@onedr0p 
> Also I never really understood the purpose of the prometheus-operator-crds chart...

The purpose is that the crds in this chart are not templated, so we could not add custom annotations (enable server-side apply in argocd for gitops for example).

> With GitOps/Flux, my method of bootstrapping new clusters is to apply the prometheus operator CRDs manually before anything else so that any chart I install doesn't fail with something like servicemonitor cr not found. 

Also your manual step here can easily be solved by using the CRDs chart, just make sure Flux sync's the CRDs chart first and you would not have to manually install the CRDs at all.

onedr0p:
@rouke-broersma 

> Also your manual step here can easily be solved by using the CRDs chart, just make sure Flux sync's the CRDs chart first and you would not have to manually install the CRDs at all.

The `prometheus-operator-crds` chart applies helm ownership so when you want to use Flux's `crds: CreateReplace` on the `kube-prometheus-stack` chart it will remove the helm ownership from the crds. Now when the `prometheus-operator-crds` Flux Helm Release tries to reconcile it will fail because the helm ownership labels and annotations were wiped by Flux.

Another things to mention is that you would have to use Flux `dependOns` for every chart that requires the use of any Prometheus Operator CR. This means that literally everything will need to depend on the `kube-prometheus-stack` chart OR the `prometheus-operator-crds` chart for something as simple as just having CRDs installed into your cluster prior to bootstrapping the rest of your cluster.

I can't express enough how painful all this is rather than just `k apply --server-side -f https://<prometheus-operator-cr>.yaml` at initialization of your cluster and then having `kube-prometheus-stack` chart just take over managing the CRDs in a non-helm ownership way.

onedr0p:
Argo has it's plenty issues of its own like [not supporting OCI](https://github.com/argoproj/argo-cd/discussions/8799) and the downsides to installing Helm charts via `helm template thing | kubectl apply -f` (RIP Helm Hooks or any other Helm native methods) and others but there's no need to get into a argument about what is better. Both Flux and Argo are great tools it just depends on your scope on what to use.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1853,6 +1853,99 @@ prometheus-node-exporter:
...
+    #         operator: In
+    #         values:
+    #         - e2e-az1
+    #         - e2e-az2
+
+  ## SecurityContext holds pod-level security attributes and common container settings.
+  ## This defaults to non root user with uid 2000 and gid 2000. *v1.PodSecurityContext  false
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  ##
+  securityContext:
```
jcpunk, line 1941: If this could be made compatible with the restricted profile by default that would be great.

QuentinBisson:
We really need to reach a concensus here because we need to bump the major version to avoid issues with prometheus-operator that was upgraded in another PR. 

Either we:
- Keep the CRD folders but do not add the newer ones as they are not causing the prometheus-operator to fail
- Delete the CRD folder (fix the test) and explain to users that they need to install it themselves
- Add the crd job present in this branch
- Move something else in another chart (dashboards, rules, ...)? 
- Find another way forward.


@monotek I know you do not like that approach but would you have another one we can go with in your opinion? I would really like some help to move this forwards :(

rouke-broersma:
> We really need to reach a concensus here because we need to bump the major version to avoid issues with prometheus-operator that was upgraded in another PR.

Personally I'm fine with any solution because I already don't use this chart for CRDs. I would say that moving the dashboards out of this chart is probably not an alternative but rather a necessary next step instead because I imagine that sooner or later this limit will be hit again if the dashboards remain in this chart.

zeritti:
I'd suggest adding `thanosrulers/status` in resources in the operator's clusterrole template, the subresource is new in 0.65.0.

rouke-broersma:
> One thing to note is that Argo doesn't implement helm properly meaning the crds option present currently in this PR is something that Argo folks cannot use because they are implemented with a helm hook. I'm not using Argo so I'm not sure how important that is or if Argo users have a different means of managing crds that can keep up to date with this helm chart.

Argo converts helm hooks to Argo hooks, pre-install and preupgrade become 'presync' which should be fine in this case I believe.

xiu:
> We really need to reach a concensus here because we need to bump the major version to avoid issues with prometheus-operator that was upgraded in another PR.

> Keep the CRD folders but do not add the newer ones as they are not causing the prometheus-operator to fail

This could be an option as `prometheus-operator` doesn't need either `ScrapeConfig` or `PrometheusAgent` to start. It just pushes the problem further in the future though.

> Delete the CRD folder (fix the test) and explain to users that they need to install it themselves

This would be my preferred option to keep things simple. People can use `prometheus-operator-crds` to install the CRDs.

jkroepke:
I'm for for deleting the CRDs from kube-prometheus-stack and shift the maintenance to prometheus-operator-crd helm chart. This also resolve issue that kubectl needs manually applied on major updates.

However, please keep in mind, removing the CRDs from crds/ folder does not remove the CRD from cluster. 

This can cause can conflict that prometheus-operator-crd can't installed on existing clusters. We have to add a documentation to "migrate" the CRDs from kube-prometheus-stack to prometheus-operator-crd, for example that the CRDs needs to be annotated with `meta.helm.sh/release-name: <RELEASE_NAME>` and `meta.helm.sh/release-namespace: <RELEASE_NAMESPACE>`

onedr0p:
@jkroepke I am not happy with that solution because of what I mentioned here and my proceeding 3 comments. Flux is able to manage the CRDs just fine in the special `crds` folders.

https://github.com/prometheus-community/helm-charts/pull/3267#issuecomment-1540089732

onedr0p:
> And what about non-flux users?

Documentation still stays the same, `kubectl apply` the crds on upgrade. I think we've found the major issue here, being the grafana dashboards being the largest in size[1].

[1] https://github.com/prometheus-community/helm-charts/pull/3267#issuecomment-1528349719

jkroepke:
> I think we've found the major issue here, being the grafana dashboards having the largest size[1].

Yeah. I created #3416 as POC which minified all the dashboard json files.

From
```
1.6M    ../templates/grafana/dashboards-1.14
```

down to 

```
688K    ../templates/grafana/dashboards-1.14
```

which can have a potential save of 1MB data (or 1MB of pure whitespaces)...

rouke-broersma:
> > I think we've found the major issue here, being the grafana dashboards having the largest size[1].
> 
> Yeah. I created #3416 as POC which minified all the dashboard json files.
> 
> From
> ```
> 1.6M    ../templates/grafana/dashboards-1.14
> ```
> 
> down to 
> 
> ```
> 688K    ../templates/grafana/dashboards-1.14
> ```
> 
> which can have a potential save of 1MB data (or 1MB of pure whitespaces)...

https://github.com/prometheus-community/helm-charts/pull/3267#issuecomment-1527226808

Helm already compresses all manifests inside the secret so not sure if minifying the json helps anything. Needs to be properly tested.

jkroepke:
Sure, there is branch + PR https://github.com/prometheus-community/helm-charts/pull/3416 which an Green CI. Everyone is invited to test that.

jkroepke:
> Helm already compresses all manifests inside the secret so not sure if minifying the json helps anything. Needs to be properly tested.

Everyone is invited to test #3416, the CI is at least green.

rouke-broersma:
> > Helm already compresses all manifests inside the secret so not sure if minifying the json helps anything. Needs to be properly tested.
> 
> Everyone is invited to test #3416, the CI is at least green.

Ah I missed that you posted the secret sizes sorry, looks good! Tho `979K` is still dangerously close to the limit, a couple more crds (which some have commented are already planned) and we're back there.

jkroepke:
I agree, that this is not a long term solution. We may need to discuss how to proceed there. However there a short-term fix for now to proceed as usual.

I do not really understand, why CRDs are stored in a helm release while helm does not managed them. The helm guys are crazy.

QuentinBisson:
This will be closed in favor of https://github.com/prometheus-community/helm-charts/pull/3416/ but I will keep it open to make this into an issue to define the way forward

monotek (changes requested):
set to "request changes" to block accidental merging it as changes are already implemented in https://github.com/prometheus-community/helm-charts/pull/3416

jkroepke:
Hi there, would it make sense to create an issue instead discuss in a PR? It feels like this topic is in sleep mode for know. I'm interrest into finding a solution before having the issues on new CRDs again.

QuentinBisson:
@jkroepke yes it would but I did not have the bandwith to write the issue down yet, you can do it if you want :)

jkroepke:
> @jkroepke yes it would but I did not have the bandwith to write the issue down yet, you can do it if you want :)

#3548 created. @QuentinBisson

QuentinBisson:
Thank you @jkroepke :)

PR #3289, opened 2023-04-26
[prometheus] Add support for renaming generated persistent volume

naphta:
ping: @gianrubio @zanhsieh @Xtigyro @naseemkullah

We're currently in the throws of migrating from an old forked version of this code base (or similar) from many years ago and we're trying to get the old volumes to marry up with the ones being generated by the new chart to make migrating a little bit simpler. I figured allowing the volume name to be overridden (combined with `fullNameOverride`) covers a pretty broad use case, certainly mine.

N/A

 It's a simple enough change with little fallout, I decided to follow a pattern similar to `.Values.server.service.statefulsetReplica` as this functionality is pretty easily covered via `existingClaim` if not using a statefulset. 

#### Checklist

- [x] Chart Version bumped


charts/prometheus/templates/sts.yaml
```diff
@@ -318,9 +318,13 @@ spec:
 {{ toYaml .Values.server.extraVolumes | indent 8}}
 {{- end }}
 {{- if .Values.server.persistentVolume.enabled }}
+  {{- $volumeName := "storage-volume" }}
+  {{- if ne .Values.server.persistentVolume.statefulSetNameOverride "" }}
+  {{- $volumeName = .Values.server.persistentVolume.statefulSetNameOverride }}
+  {{- end }}
```
zanhsieh, line 324: Why not use `default`? Or you might consider merge with line 323 / 327 below
```suggestion
  {{- $volumeName := default "storage-volume" .Values.server.persistentVolume.statefulSetNameOverride "" }}
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #3300, opened 2023-04-27
[prometheus-pushgateway] fix apiVersion for PDB

akamac:


zanhsieh (changes requested):
@akamac 
Thank you for the contribution. Would you mind to bump the version in Chart.yaml so the CI would get it merged once approved?

akamac:
> @akamac Thank you for the contribution. Would you mind to bump the version in Chart.yaml so the CI would get it merged once approved?

done

zanhsieh (approved):

zeritti (approved):

PR #3309, opened 2023-04-30
[prometheus]Remove 'app.kubernetes.io/version' label from selector

ryodocx:
prometheus chart is in violation of the guidelines for this section:
https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md#deployments-statefulsets-daemonsets-selectors
>spec.selector.matchLabels defined in Deployments/StatefulSets/DaemonSets >=v1/beta2 must not contain helm.sh/chart label or any label containing a version of the chart, because the selector is immutable. The chart label string contains the version, so if it is specified, whenever the Chart.yaml version changes, Helm's attempt to change this immutable field would cause the upgrade to fail.

This PR resolves failures that may occur when updating to future Chart versions.

none

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: v2.43.0
-version: 21.1.2
+version: 22.0.0
 kubeVersion: ">=1.16.0-0"
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
--- a/charts/prometheus/README.md
+++ b/charts/prometheus/README.md
@@ -67,7 +67,9 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
 
 ### To 22.0
 
-The `app.kubernetes.io/version` label has been removed from the selector. Therefore, you must delete the previous StatefulSet or Deployment before performing the upgrade.
+The `app.kubernetes.io/version` label has been removed from the pod selector.
+
+Therefore, you must delete the previous StatefulSet or Deployment before upgrading. Performing this operation will cause **Prometheus to stop functioning** until the upgrade is complete.
 
 ```console
 kubectl delete deploy,sts -l app.kubernetes.io/name=prometheus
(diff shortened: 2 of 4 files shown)


zanhsieh (changes requested):
@ryodocx 
Thanks for the contribution. Since this PR bumps the major version, would you mind to write some procedure and / or the user should know while upgrading from 21.x.x to 22.x.x in README.md upgrade section? For example, would previous data persist? If not, what should the user do to save them somewhere and import back?

ryodocx:
> @ryodocx Thanks for the contribution. Since this PR bumps the major version, would you mind to write some procedure and / or the user should know while upgrading from 21.x.x to 22.x.x in README.md upgrade section? For example, would previous data persist? If not, what should the user do to save them somewhere and import back?

@zanhsieh I've added the upgrade steps to the README.md file. It's pretty much the same procedure as updating to v21.x.x, but if you need more details, just let me know.

zeritti:
Would it be possible to correct `pdb.yaml`, please? It appears it is using wrong template `prometheus.server.labels` in its pod selector and would suffer from the same problem being corrected in this PR once the PR gets merged. The correct selector label set should be like elsewhere, i.e. `prometheus.server.matchLabels`.

Wrt README, this change is also affecting a service.

ryodocx:
> Would it be possible to correct `pdb.yaml`, please? It appears it is using wrong template `prometheus.server.labels` in its pod selector and would suffer from the same problem being corrected in this PR once the PR gets merged. The correct selector label set should be like elsewhere, i.e. `prometheus.server.matchLabels`.

@zeritti Thank you for your feedback. I have fixed bdb.yaml and re-checked that all pod selectors below are using `prometheus.server.matchLabels`:
* deploy.yaml `spec.selector`
* headless-svc.yaml `spec.selector`
* network-policy.yaml `spec.podSelector`
* pdb.yaml `spec.selector`
* service.yaml `spec.selector`
* sts.yaml `spec.selector`

> Wrt README, this change is also affecting a service.

I have updated the README. I'm not sure if the intention is clear enough. Is it sufficient?

zeritti:
Thank you, @ryodocx, looks good to me.

zanhsieh (approved):

PR #3333, opened 2023-05-05
[prometheus-rabbitmq-exporter] allow prometheus rule to be controlled through values.yaml

denny60004:
The original Prometheus rule cannot be added or deleted through values.yaml.

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-rabbitmq-exporter]`)


zanhsieh (changes requested):
@denny60004 
Thank you for the contribution. Since this PR bumped the minor version, would you take some time to write what the user should know for upgrading from previous version in README.md? E.g. `.Values.prometheus.rules` -> `.Values.prometheusRule`, what rules by default exist in previous version but got removed from this upgrade (and reasons?), what things expect to break and how to prevent them?

denny60004:
@zanhsieh PTAL!!

charts/prometheus-rabbitmq-exporter/values.yaml
```diff
@@ -71,9 +71,69 @@ prometheus:
...
+    #          Memory available for Rabbmitmq node {{ "{{ $labels.node }}" }} is lower than 10%
+    #  - alert: RabbitmqTooManyConnections
+    #    expr: rabbitmq_connectionsTotal{service="{{ template "prometheus-rabbitmq-exporter.fullname" . }}"} > 1000
+    #    for: 5m
+    #    labels:
+    #      severity: warning
+    #    annotations:
+    #      summary: Too many connections to the Rabbitmq cluster
+    #      description: |
+    #          The Rabbitmq cluster tracked by {{ "{{ $labels.service }}" }} has too many connections {{ "{{ $value }}" }} (> 1000)
```
zanhsieh, line 136: @denny60004 
Line 83 to 136 has to be uncommented; otherwise it's not the same as before upgrade as stated in README.md.

zanhsieh (changes requested):

zanhsieh (dismissed):

denny60004:
@desaintmartin @Juanchimienti @monotek 
This PR needs yours approve. 
PTAL

charts/prometheus-rabbitmq-exporter/values.yaml
```diff
@@ -71,9 +71,69 @@ prometheus:
...
+
+## Custom PrometheusRules to be defined
+## The value is evaluated as a template, so, for example, the value can depend on .Release or .Chart
+## ref: https://github.com/coreos/prometheus-operator#customresourcedefinitions
+prometheusRule:
+  enabled: false
+  additionalLabels: {}
+  namespace: ""
   rules:
-    enabled: false
```
monotek, line 75: this looks like as we're changing this to have the rules enabled by default?
i would keep it disabled, to not change default behaviour of the chart.
this and the other changed variables would be major version change too.

monotek (changes requested):

charts/prometheus-rabbitmq-exporter/values.yaml
denny60004: I have made changes so that the enabling of rules is still controlled through `values.prometheus.rules`. Then, the original rules have been moved to `values.prometheus.rules.additionalRules`.

monotek (changes requested):

monotek (approved):

Issue #3031, opened 2023-02-14
[node-exporter] node-exporter port exposed even when kubeRBACProxy is enabled


r0bj:
### Describe the bug a clear and concise description of what the bug is.

When `kubeRBACProxy.enabled: true` node-exporter metrics should be protected by TLS and RBAC. So this is the purpose of `kube-rbac-proxy` which listens on default node-exporter port 9100 and proxying traffic to node-exporter.
But for some reason node-exporter is listening on 8100 port bound to 0.0.0.0 (with `hostNetwork: true`) which means that node-exporter raw port is still available for anybody in the cluster.
Even `service.listenOnAllInterfaces: false` option is not helpful, node-exporter is then bound to the node IP which doesn't change much in this context.

A quick fix would be to bind node-exporter port 8100 to 127.0.0.1. `kube-rbac-proxy` container is using address `--upstream=http://127.0.0.1:8100/` as a upstream so it should work.

### What's your helm version?

v3.11.1

### What's your kubectl version?

v1.26.1

### Which chart?

prometheus-community/prometheus-node-exporter

### What's the chart version?

4.13.0

### What happened?

node-exporter port 8100 is bound to address 0.0.0.0 (or to node IP with `service.listenOnAllInterfaces: false`) instead of to address 127.0.0.1 with kubeRBACProxy enabled. So anybody in the cluster can easily bypass kube-rbac-proxy by using port 8100.

### What you expected to happen?

node-exporter port 8100 is bound to address 127.0.0.1 with kubeRBACProxy enabled. So communication with node-exporter is possible only via kube-rbac-proxy (which contacts with node-exporter via 127.0.0.1 inside the pod)

### How to reproduce it?

```
helm template node-exporter prometheus-community/prometheus-node-exporter --set kubeRBACProxy.enabled=true
```
or 
```
helm template node-exporter prometheus-community/prometheus-node-exporter --set kubeRBACProxy.enabled=true --set service.listenOnAllInterfaces=false
```

### Enter the command that you execute and failing/misfunctioning.

helm template node-exporter prometheus-community/prometheus-node-exporter --set kubeRBACProxy.enabled=true

r0bj:
Ping

cccsss01:
ping

r0bj:
Ping

PR #3336, opened 2023-05-06
[prometheus-node-exporter] fix node-exporter listen address with kubeRBACProxy enabled

r0bj:
With `kubeRBACProxy` enabled node-exporter port should be protected by TLS and RBAC (via kube-rbac-proxy) so it shouldn't be exposed and available for anybody to scrape. In order to do this change node-exporter listen address to 127.0.0.1.

- fixes https://github.com/prometheus-community/helm-charts/issues/3031

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.17.1
+version: 4.17.2
 appVersion: 1.5.0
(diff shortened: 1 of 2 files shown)


zanhsieh (approved):

zeritti:
Thank you, @r0bj, for the fix. If `hostNetwork` is not set, node exporter will enter a restart loop as kubelet won't be able to reach it when executing probes. From my point of view, though, since `hostNetwork` is set in the typical use case of this chart, the limitation need not be addressed at this moment.

PR #3361, opened 2023-05-10
[kube-prometheus-stack] feat: allow override of service name for alertmanager default ingress

ekristen:

The servicePort was already able to be overwritten, this allows the serviceName as well.

Use cases vary but I need to be able to point the default ingress for the alertmanager to a specific instance/service of the alertmanager from time to time and using the ingress per replica isn't useful as it prefixes/suffixes the ingress. By allowing the default to be overwritten, just brings parity since the port can be already.

* did not open an issue, just implemented. 
* allow override of the serviceName like the port is already able to be.

Keeps backwards compatibility. Just allows it to be overwritten.

#### Checklist

- [x] Chart Version bumped


ekristen:
Can I get a review?

ekristen:
> Hi @ekristen, this is a very specific use case 😄 but I guess there's no harm in having options.

It is, but thank you for having this approach, options is good and what makes OSS and community things like this viable.

GMartinez-Sisti (approved):

ekristen:
Thank you!

PR #3363, opened 2023-05-10
[kube-state-metrics] Fix body must be of type string bug for cilium network policy

kadirtaskiran:
Ensures cilium network policy configured correctly templated for kube-state-metrics

none

```
$kubectl explain CiliumNetworkPolicy.spec.ingress.toPorts.ports
KIND:     CiliumNetworkPolicy
VERSION:  cilium.io/v2

RESOURCE: ports <[]Object>

DESCRIPTION:
     Ports is a list of L4 port/protocol

     PortProtocol specifies an L4 port with an optional transport protocol

FIELDS:
   port	<string> -required-
     Port is an L4 port number. For now the string will be strictly parsed as a
     single uint16. In the future, this field may support ranges in the form
     "1024-2048 Port can also be a port name, which must contain at least one
     [a-z], and may also contain [0-9] and '-' anywhere except adjacent to
     another '-' or in the beginning or the end.

   protocol	<string>
     Protocol is the L4 protocol. If omitted or empty, any protocol matches.
     Accepted values: "TCP", "UDP", "SCTP", "ANY" Matching on ICMP is not
     supported. Named port specified for a container may narrow this down, but
     may not contradict this.
```

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 5.6.1
+version: 5.6.2
 appVersion: 2.8.2
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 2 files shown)


dotdc:
Thank you for your contribution @kadirtaskiran !

PR #3377, opened 2023-05-15
[kube-prometheus-stack] Do not set resources as default values for configmap-reloader container

nlamirault:

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -2234,13 +2234,13 @@ prometheusOperator:
     enableProbe: false
 
     # resource config for prometheusConfigReloader
-    resources:
-      requests:
-        cpu: 200m
-        memory: 50Mi
-      limits:
-        cpu: 200m
-        memory: 50Mi
+    resources: {}
+      # requests:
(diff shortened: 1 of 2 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -2234,13 +2234,13 @@ prometheusOperator:
...
 
     # resource config for prometheusConfigReloader
-    resources:
-      requests:
-        cpu: 200m
-        memory: 50Mi
-      limits:
-        cpu: 200m
-        memory: 50Mi
+    resources: {}
```
QuentinBisson, line 2237: Is there a reason you cannot override the value when you deploy the chart? This value works for most of the users

nlamirault: I see that a lot of charts do not set resources with default values.
And i would like to set resources without `cpu` limit.

zeritti: In this particular case, setting a resource to "0" will disable that limit/request. If support for setting `resources` empty is desired, further changes will be required in the operator's deployment template.

PR #3395, opened 2023-05-17
[prometheus-postgres-exporter] Correct role

zeritti:

If PodSecurityPolicy API is not available in a cluster, a `use` permission on resources `podsecuritypolicies` should not be created through the role template.

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/role.yaml
+++ b/charts/prometheus-postgres-exporter/templates/role.yaml
@@ -8,7 +8,7 @@ metadata:
     chart: {{ template "prometheus-postgres-exporter.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-{{- if .Values.rbac.pspEnabled }}
+{{- if (and .Values.rbac.pspEnabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy")) }}
 rules:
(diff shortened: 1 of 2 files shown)


charts/prometheus-postgres-exporter/templates/role.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     chart: {{ template "prometheus-postgres-exporter.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-{{- if .Values.rbac.pspEnabled }}
+{{- if (and .Values.rbac.pspEnabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy")) }}
```
zanhsieh, line 11: Extra `(` `)` should not be necessary.
```suggestion
{{- if and .Values.rbac.pspEnabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
```

zanhsieh (changes requested):

charts/prometheus-postgres-exporter/templates/role.yaml
zeritti: Corrected, thank you.

zanhsieh (approved):

PR #3415, opened 2023-05-22
[prometheus-operator] Set secretFieldSelector to exclude not necessary secret types 

gproxyz:
This PR makes the correct work for prometheus-operator in huge cluster.  In my case prometheus-operator used more than 2Gb RAM and kube-api more then 5Gb RAM answering the requests, with not set `secretFieldSelector`. 

I really don't understand why prometheus-operator need to watch secret types which are excluded in this PR.

No issue 

#### Checklist

- [x] Chart Version bumped


QuentinBisson (approved):
This makes a lot of sense that we set sane defaults

PR #3423, opened 2023-05-24
[kube-state-metrics] add missing bearerToken config for ServiceMonitor self endpoint

r0bj:
ServiceMonitor contains two endpoints: `http` and self monitor `metrics`. Self monitor endpoint is missing bearerToken config so it not possible to monitor it with `kubeRBACProxy` enabled. Let's fix by adding missing optional fields.

#### Checklist

- [x] Chart Version bumped


dotdc (approved):
Not sure if it's a feature or a fix, but I'll accept as is.

PR #3461, opened 2023-06-05
[kube-prometheus-stack] Create job-patch ServiceAccount even when global.rbac.create=false

philandstuff:

If you set global.rbac.create to false, this service account won't be created but it will still be referenced from the webhook admission patch jobs.  This will cause the jobs to fail because the service account is missing.

If global.rbac.create is false, the correct thing to do is to still create the service account (as we still do with all other service accounts), but to allow the user to define (Cluster)Roles and (Cluster)RoleBindings against that service account.

This brings the template in line with [helm charts RBAC best practises](https://helm.sh/docs/chart_best_practices/rbac/):

> RBAC and ServiceAccount configuration should happen under separate keys. They are separate things. Splitting these two concepts out in the YAML disambiguates them and makes this clearer.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 46.6.0
+version: 46.7.0
 appVersion: v0.65.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


QuentinBisson:
I'm not really sure what would be your use case to disable rbac though. Can you explain?

philandstuff:
If the feature already exists on the helm chart, it should work. Currently, it is broken.

We wanted disable rbac in the chart so we can manage it ourselves.

QuentinBisson:
Can you sign your commit?

philandstuff:
I don't have a GPG key and I try to avoid it if at all possible.

GMartinez-Sisti:
Hi @philandstuff, your contribution is appreciated, but we kindly ask that you adhere to the [Contribution Guidelines](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md) in place. Signing your work is a requirement to contribute. If you're still not comfortable creating a GPG key, maybe you can ask someone else to own the commit and do that for you.

PR #3468, opened 2023-06-07
[prometheus-blackbox-exporter] update values to use default from app-version of chart

sstoffregen:

Update to AppVersion 0.24.0 was announced.

But I have noticed that the image-tag was not updated.
So, I have updated the values.yaml to use v0.24.0 as default application version.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.9.0
+version: 7.9.1
 appVersion: 0.24.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.9.1
+version: 7.10.0
 appVersion: 0.24.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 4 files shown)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 7.9.0
+version: 7.9.1
```
rsotnychenko, line 4:
```suggestion
version: 7.10.0
```

0.24 release [introduces](https://github.com/prometheus/blackbox_exporter/releases/tag/v0.24.0) some behavior changes, so I'd suggest to bump minor version instead.

rsotnychenko (changes requested):
@sstoffregen thanks for the contribution!

All looks good, but one small nit related to chart versioning.

charts/prometheus-blackbox-exporter/Chart.yaml
sstoffregen: There was a minor change from 7.8.0 to 7.9.0.
Within PR #3457 everything was updated to use v0.24.0, but somehow the image-tag was not updated.
You could see my change as a small bugfix as "patch release".

monotek (changes requested):
Could you change it so that the image version is taken from Chart.yaml app labbel like in: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-conntrack-stats-exporter/templates/daemonset.yaml#L35

charts/prometheus-blackbox-exporter/Chart.yaml
sstoffregen: OK, I agree.
Thanks for clarifying this!

I will update my PR this afternoon, or feel free to fix it on your own if you like.

jkroepke:
> Could you change it so that the image version is taken from Chart.yaml app labbel like in: [`main`/charts/prometheus-conntrack-stats-exporter/templates/daemonset.yaml#L35](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-conntrack-stats-exporter/templates/daemonset.yaml?rgh-link-date=2023-06-07T09%3A57%3A06Z#L35)

tbh, while I update the appVersion to 0.24 (https://github.com/prometheus-community/helm-charts/pull/3457), I would not except that the tag is not overridden. I would also see that appVersion is used and leave the tag value unset by default.

monotek (approved):

PR #3484, opened 2023-06-11
[prometheus] Added node label to pod scrape jobs

romich:
This PR makes Prometheus preserve `__meta_kubernetes_pod_node_name` as `node` label for `kubernetes-pods` jobs, the same as it's already done for `service-endpoints`.

Shouldn't increase cardinality as node stays the same for the duration of pod lifetime.

#### Checklist

- [x] Chart Version bumped


romich:
I don't think this change needs anything in README, as it's purely additive — no upgrading procedures, incompatibilities, value field syntax changes, etc. The only reason I've decided to bump minor version is because some metrics will now have an additional label after the upgarde. If you believe it doesn't warrant a minor bump, then I'll increase the patch version instead.

zanhsieh:
> I don't think this change needs anything in README, as it's purely additive — no upgrading procedures, incompatibilities, value field syntax changes, etc. The only reason I've decided to bump minor version is because some metrics will now have an additional label after the upgarde. If you believe it doesn't warrant a minor bump, then I'll increase the patch version instead.

Please increase patch version instead. Thank you.

PR #3492, opened 2023-06-14
[prometheus-postgres-exporter] user from secret

sebX2:
The PR enables the prometheus-postgres-exporter to read the username from a Kubernetes secret.

In my use case, I store Postgres credentials inside vault secrets. On the basis of these vault secrets an external-secret operator instance automatically creates kubernetes secrets with the needed credentials. My deployment infrastructure itself does not have access to the vault. Thus I cannot dynamically add the value to the chart while rolling it out. In order to not loose my vault as a single point of truth for my complete credential set, I depend on that feature.

No related issue.

Since there is a reasonable default value for the user name, I decided to not validate that the username is only set once. It differs from the password validation approach since otherwise user would have been set to null in case of using secret as a source. This seemed pretty counterintuitive to me.

I tested:
- reading the username from a secret with successful authentication to a Postgres DB
- if the chart would default to user in case no secret is defined

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.11.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.4.5
+version: 4.5.0
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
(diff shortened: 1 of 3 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.11.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.4.4
+version: 4.4.5
```
zeritti, line 5:
```suggestion
version: 4.5.0
```

I would suggest increasing the minor version number, a new feature is being introduced after all.

sebX2: true

zeritti:
Thank you, @sebX2, looks good to me.

zeritti (approved):

PR #3497, opened 2023-06-16
[kube-prometheus-stack] Don't drop cadvisor container_cpu_cfs_throttled_seconds_total metrics

PavelGloba:

This PR returns previously dropped container_cpu_cfs_throttled_seconds_total metric.
Numerous dashboards and articles use this metric. I see no reason why it should be dropped by default. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 46.8.0
+version: 46.9.0
 appVersion: v0.65.2
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


GMartinez-Sisti:
This metric was intentionally removed on https://github.com/prometheus-community/helm-charts/issues/2133.

@SuperQ do you mind adding your thoughts here?

SuperQ:
The reason we drop this metric is that it duplicates `container_cpu_cfs_throttled_periods_total`. Users need to use periods because there is no seconds analog to the corresponding `container_cpu_cfs_periods_total`.

PavelGloba:
@GMartinez-Sisti @SuperQ 

It's not duplicating container_cpu_cfs_throttled_periods_total
There is no way to figure out how much time is actually spent on throttling in seconds without this metric.  With container_cpu_cfs_throttled_periods_total and container_cpu_cfs_periods_total you can see only the percentage of throttling

PR #3523, opened 2023-06-26
[prometheus-redis-exporter]: Spec Security Context

bkochendorfer:
We use both container and spec level security contexts. This change allows a user to add the spec level security context as well as the pod/container level security context.

- fixes #

#### Checklist

- [X] Chart Version bumped


jcpunk:
The second commit doesn't appear to have the DCO

bkochendorfer:
oops, thanks for pointing that out @jcpunk!

charts/prometheus-redis-exporter/templates/deployment.yaml
```diff
@@ -28,6 +28,10 @@ spec:
         - name: {{ . }}
       {{- end}}
       {{- end }}
+      {{- with .Values.specSecurityContext }}
```
zeritti, line 31:
```suggestion
      {{- with .Values.podSecurityContext }}
```

Security context is commonly referred to in the repository's charts by `securityContext`, `containerSecurityContext` and `podSecurityContext`. Since users and contributors are familiar with these fields, `podSecurityContext` would fit, would it not? 

Please, enter the new field in `values.yaml` with its default value.

zeritti (changes requested):

bkochendorfer:
@zeritti I noticed the other `securitycontext` value didn't have a default so I did not put a default here. Do you want me to include the new one, both, or neither?

Here is an example of both.
```yaml
securityContext:                 
  capabilities:                  
    drop: ["ALL"]                
                                 
podSecurityContext:              
  seccompProfile:                
    type: RuntimeDefault         
  runAsNonRoot: true             
  runAsUser: 10001               
  runAsGroup: 10001              
```

zeritti:
> I noticed the other securitycontext value didn't have a default so I did not put a default here. Do you want me to include the new one, both, or neither?

Thank you. Like with `securityContext` set to default `{}`, I would suggest inserting `podSecurityContext` set to `{}`. You can enter the parameters that you give in the example above, too, but probably commented out.

jcpunk:
My personal preference is to set as much in line with pss-restricted as possible by default.  I'm not sure too many folks in the wild apply extra hardening by hand and this would be a good way of noting which settings are expected to work with these containers.

bkochendorfer:
Folks let me know, I am happy to provide the defaults I am using (those commented in the example) as the default.

zanhsieh (approved):

zeritti (approved):

PR #3533, opened 2023-06-29
[prometheus-postgres-exporter] Add tpl functionality to the datasource config of Postgres-exporter

hansstandaert:
What this PR does / why we need it

When we want to do sub-charting (eg. in a postgres chart) we need to duplicate the datasource config code in the values file. We want to avoid duplicating this configuration as the postgresql-chart is used by multiple applications and we want to avoid having to duplicate all these values in each sub-chart using this.

Furthermore the tpl function does not incur any form of regression on the previous setup thus the change is non-breaking.

Checklist
 - [x] Chart Version bumped
 - [x] Title of the PR starts with chart name (e.g. [prometheus-couchdb-exporter])

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.11.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.5.0
+version: 4.6.0
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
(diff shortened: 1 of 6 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.11.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.5.0
+version: 4.6.0
```
zanhsieh, line 5: @hansstandaert 
Thanks for the contribution. Since minor version bumped, do you mind to mention something that the user need to know for upgrading, e.g. characters need to escape, in README.md upgrade section?

zanhsieh (changes requested):

charts/prometheus-postgres-exporter/Chart.yaml
hansstandaert: @zanhsieh Thank you for the quick reply. I took care of it and updated the README.md file

zanhsieh (approved):

zanhsieh (approved):

PR #3555, opened 2023-07-05
[prometheus-pingdom-exporter] Add support for Serviceaccount Volume and VolumeMount

DannyLyubenov:

There is the need to extract secrets from Vault for the Pingdom account and we are using the Vault CSI driver which needs to mount the volume onto the container. To achieve this we need to extend the chart to create `ServiceAccount`, `Volume` and `VolumeMounts`

N/A

Hi @monotek and @rpahli could you please review the chart when you have a moment

#### Checklist

- [ ] Chart Version bumped


rpahli (approved):

DannyLyubenov:
Hi Rico (@rpahli) thank you for approving my PR, is there anything else that needs to be done before it gets merged? Many thanks

charts/prometheus-pingdom-exporter/templates/deployment.yaml
```diff
@@ -18,6 +18,7 @@ spec:
       labels:
         {{- include "prometheus-pingdom-exporter.selectorLabels" . | nindent 8 }}
     spec:
+      serviceAccountName: {{ template "prometheus-pingdom-exporter.serviceAccountName" . }}
```
monotek, line 21: should that not be?

```suggestion
      serviceAccountName: {{ template "prometheus-pingdom-exporter.serviceAccount.name" . }}
```

also would seem to be empty by default?

monotek (changes requested):

charts/prometheus-pingdom-exporter/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    {{- include "prometheus-pingdom-exporter.labels" . | nindent 4 }}
+  name: {{ template "prometheus-pingdom-exporter.serviceAccountName" . }}
```
monotek, line 7:
```suggestion
  name: {{ template "prometheus-pingdom-exporter.serviceAccount.name" . }}
```

charts/prometheus-pingdom-exporter/templates/deployment.yaml
DannyLyubenov: I don't this this is correct because of `helpers.tpl` - https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-pingdom-exporter/templates/_helpers.tpl#L53C1-L62

I get this error if i change it to `serviceAccount.name`
```
Error: template: prometheus-pingdom-exporter/templates/serviceaccount.yaml:7:20: executing "prometheus-pingdom-exporter/templates/serviceaccount.yaml" at <{{template "prometheus-pingdom-exporter.serviceAccount.name" .}}>: template "prometheus-pingdom-exporter.serviceAccount.name" not defined
```

DannyLyubenov: I tend to use `fullnameOverride` so all resources are named the same. Would you say that there should be a default name for the service account?

monotek (approved):

PR #3559, opened 2023-07-06
[prometheus-blackbox-exporter] PodSecurityPolicy is a cluster level r…

carinadigital:
…esource.

PodSecurityPolicy is a cluster level resource. It does not use **namespace**.
See https://v1-24.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#podsecuritypolicy-v1beta1-policy

Namespace field is ignored, so this just removes some possible confusion.

#### Checklist

- [ X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [ X] Chart Version bumped
- [ X] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/podsecuritypolicy.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/podsecuritypolicy.yaml
@@ -3,7 +3,6 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: {{ template "prometheus-blackbox-exporter.fullname" . }}-psp
-  namespace: {{ template "prometheus-blackbox-exporter.namespace" . }}
   labels:
     {{- include "prometheus-blackbox-exporter.labels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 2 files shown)


charts/prometheus-blackbox-exporter/templates/podsecuritypolicy.yaml
```diff
@@ -3,7 +3,6 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
```
monotek, line 3: can we rather remove this completly?
it's deprectaed sind k8s 1.25 anyway: https://kubernetes.io/docs/concepts/security/pod-security-policy/

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/podsecuritypolicy.yaml
Toasterson: Are there any news on this? Blackbox exporter is now blocking the Upgrade to 1.25 and Google may sson start to autoupgrade. I have removed it to test. Reading https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/ this seems save to do, as unless you have namespace based restrictions there are no restrictions in place.

PR #3565, opened 2023-07-09
[prometheus-postgres-exporter] Upgrade postgres-exporter version

adinhodovic:

Bump default postgres-exporter version. Queries.yaml is deprecated but not removed, so still should be safe to use.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "0.11.1"
+appVersion: "0.13.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.6.0
+version: 4.7.0
 home: https://github.com/prometheus-community/postgres_exporter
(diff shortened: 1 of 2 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "0.11.1"
+appVersion: "0.13.1"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 4.6.0
+version: 4.7.0
```
zanhsieh, line 5: @adinhodovic 
Thanks for the contribution. Since the minor version was bumped, can you write what the users should know when they upgrade from previous version?  E.g. what features were deprecated, what values in values.yaml were no longer supported. Otherwise please just bump to patch version, e.g. 4.6.x.

zanhsieh (changes requested):

zanhsieh (approved):

monotek (approved):

PR #3574, opened 2023-07-10
[prometheus-rabbitmq-exporter]  allow RabbitMQ URL to be passed as a templated field

davido912:
Currently, the values passed to the RabbitMQ URL env var has to be hardcoded. This is especially cumbersome to maintain when integrating the Prometheus Exporter with other dependencies in one chart when the Prometheus Exporter is pulled as dependency. This makes it easy to build the reference URL in cases where the RabbitMQ k8s service endpoint is used and so forth.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-rabbitmq-exporter/Chart.yaml
+++ b/charts/prometheus-rabbitmq-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Rabbitmq metrics exporter for prometheus
 name: prometheus-rabbitmq-exporter
-version: 1.6.1
+version: 1.7.0
 appVersion: v0.29.0
 home: https://github.com/kbudde/rabbitmq_exporter
 sources:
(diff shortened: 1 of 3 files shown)


charts/prometheus-rabbitmq-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Rabbitmq metrics exporter for prometheus
 name: prometheus-rabbitmq-exporter
-version: 1.6.0
+version: 1.6.1
```
monotek, line 4: new feature should raise minor version

```suggestion
version: 1.7.0
```

monotek (changes requested):

charts/prometheus-rabbitmq-exporter/Chart.yaml
davido912: gotcha, thanks! :)

monotek (approved):

PR #3579, opened 2023-07-10
[Alertmanager] add enabled property into config block

figuerascarlos:
Due to `values.schema.json` validation is not posible to set config block to `false`. I would like to set config to `false` because I  use extraSecretMounts to populate the configuration  due some sensible information.

#### Checklist

- [x] Chart Version bumped


charts/alertmanager/Chart.yaml
```diff
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 0.33.1
+version: 0.33.2
```
monotek, line 9: new feature should raise minor version.
```suggestion
version: 0.34.0
```

monotek (changes requested):

AnastasiosDados:
Nice, I was having the same issue as I want to inject my config via extraVolumeMounts and extraVolumes and it is not possible at the moment as it enforces to have config values.

monotek (approved):

PR #3582, opened 2023-07-11
[prometheus-rabbitmq-exporter] allow passing an external ConfigMap to override env vars

davido912:
Following https://github.com/prometheus-community/helm-charts/pull/3574, I added a way to template the URL so that parent charts can template the URL into it without having to hardcode it. However, it appears subcharts don't read values from the [parent chart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/).

Therefore, to allow for customisation of the envs vars, I moved all the env vars to a configmap and added the option to add an externally supplied configmap which will override whatever is in the first one. 

According to the [Kubernetes API reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/), for envFrom: 

"List of sources to populate environment variables in the container. The keys defined within a source must be a C_IDENTIFIER. All invalid keys will be reported as an event when the container is starting. When a key exists in multiple sources, the value associated with the last source will take precedence. Values defined by an Env with a duplicate key will take precedence. Cannot be updated."

So the last overriding configmap which was added here will override, whereas anything in env will override everything in envFrom. Therefore, when for the password, when a secret is supplied, env takes precedence. 

I hope this was clear and would appreciate feedback. Couldn't think of a better way to give some control back to the user.

#### Checklist

- [x] Chart Version bumped


davido912:
@monotek perhaps you can assist in getting this through or checking this out, given you checked the previous PR :)

charts/prometheus-rabbitmq-exporter/templates/deployment.yaml
```diff
@@ -25,7 +25,10 @@ spec:
 {{ toYaml .Values.podLabels | indent 8 }}
         {{- end }}
       annotations:
-{{ toYaml .Values.annotations | indent 8 }}
+        checksum/exporter-config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- if .Values.annotations }}
+          {{- toYaml .Values.annotations | nindent 8 }}
+        {{- end }}
```
zanhsieh, line 31:
```suggestion
        {{- with .Values.annotations }}
          {{- toYaml .| nindent 8 }}
        {{- end }}
```

charts/prometheus-rabbitmq-exporter/templates/configmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: {{ template "prometheus-rabbitmq-exporter.configMapName" . }}
+  labels:
+    app: {{ template "prometheus-rabbitmq-exporter.name" . }}
+    chart: {{ template "prometheus-rabbitmq-exporter.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  {{- if .Values.additionalLabels }}
+{{ toYaml .Values.additionalLabels | indent 4 }}
+  {{- end }}
```
zanhsieh, line 12:
```suggestion
  {{- with .Values.additionalLabels }}
    {{- toYaml . | indent 4 }}
  {{- end }}
```

zanhsieh (changes requested):

davido912:
@zanhsieh applied the changes

zanhsieh (approved):

zanhsieh (approved):

davido912:
Apparently we're missing one more review (@desaintmartin / @Juanchimienti / @monotek ) 🙌

monotek (approved):

monotek (approved):

PR #3618, opened 2023-07-20
[kube-prometheus-stack] bumping chart version to allow updated dependent chart versions to be updated

smbambling:
Bumps the chart version in Charts.yaml to force the dependent charts to fetch updated version when the chart is published.

This specifically is to fetch the latest Grafana Helm chart grafana-6.58.4 (https://github.com/grafana/helm-charts/releases/tag/grafana-6.58.4) which correct a bug with pulling dashboards via the sidecar container.

Without a Chart.lock which was removed in a previous commit this seems like the best method for updating dependent chart versions.

N/A

Without a Chart.lock which was removed in a previous commit this seems like the best method for updating dependent chart versions. If there is another preferred method I'm happy to follow that process.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 48.1.1
+version: 48.1.2
 appVersion: v0.66.0
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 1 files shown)


GMartinez-Sisti (approved):

GMartinez-Sisti:
I think this is ok. I can't think of another way to do this too, although unrelated PRs would eventually bump the chart version and cause this too.

smbambling:
@GMartinez-Sisti Thanks for the quick review and approval. 

I've been waiting a few days in the hope that another PR would be merged and render this moot but no such luck.

PR #3620, opened 2023-07-20
[alertmanager] Add additional configuration options.

### LINKED ISSUE #2622: Ha doesn't work correctly

Hi
I changed the replica to 2. but I got an error that said can't resolve another instance of the alert manager.
```
msg=refresh result=failure addr=alertmanager-mw-0.alertmanager-mw-headless:9094 err="1 error occurred:\n\t* Failed to resolve alertmanager.alertmanager-headless:9094: lookup alertmanager.alertmanage-headless on 10.96.0.10:53: no such host\n\n"
```
I can curl the addresses from another pod in the same namespace but I got an error in the alert manager logs.

### What's your helm version?

v3.8.0

v1.21.0

### Which chart?

alertmanager

### What's the chart version?

latest

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

 helm upgrade alertmanager-mw -n alertmanager -f values-production.yaml ../alertmanager

_No response_

sam-mcbr:

This PR fixes a bug when trying to deploy alertmanager HA. The cluster port is set to `9094` but the StatefulSet only ever exposes `9093`. This PR additionally adds some extra configuration options at allow more flexible uses of the config-reloader.

- fixes #2622 (closed due to inactivity)

Would love to get this merged quickly as the port issue is a blocker for me deploy the chart. Happy to discuss and iterate on this if needed! Thanks!

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -28,6 +28,9 @@ spec:
       {{- if (and (eq .Values.service.type "NodePort") .Values.service.nodePort) }}
       nodePort: {{ .Values.service.nodePort }}
       {{- end }}
+    {{- with .Values.service.extraPorts }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
   selector:
     {{- include "alertmanager.selectorLabels" . | nindent 4 }}
 ---
@@ -54,6 +57,9 @@ spec:
       targetPort: clusterpeer
       protocol: UDP
       name: cluster-udp
+    {{- end }}
+    {{- with .Values.service.extraPorts }}
+    {{- toYaml . | nindent 4 }}
     {{- end }}
   selector:
     {{- include "alertmanager.selectorLabels" . | nindent 4 }}
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -4,6 +4,9 @@ metadata:
   name: {{ include "alertmanager.fullname" . }}
   labels:
     {{- include "alertmanager.labels" . | nindent 4 }}
+    {{- with .Values.service.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
   {{- with .Values.service.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
@@ -40,6 +43,9 @@ metadata:
   name: {{ include "alertmanager.fullname" . }}-headless
   labels:
     {{- include "alertmanager.labels" . | nindent 4 }}
+    {{- with .Values.service.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
   namespace: {{ include "alertmanager.namespace" . }}
 spec:
   clusterIP: None
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -53,10 +53,6 @@ spec:
       targetPort: clusterpeer
       protocol: TCP
       name: cluster-tcp
-    - port: {{ .Values.service.clusterPort }}
-      targetPort: clusterpeer
-      protocol: UDP
-      name: cluster-udp
     {{- end }}
     {{- with .Values.service.extraPorts }}
     {{- toYaml . | nindent 4 }}
--- a/charts/alertmanager/templates/services.yaml
+++ b/charts/alertmanager/templates/services.yaml
@@ -47,11 +47,11 @@ spec:
       name: http
     {{- if or (gt (int .Values.replicaCount) 1) (.Values.additionalPeers) }}
     - port: {{ .Values.service.clusterPort }}
-      targetPort: {{ .Values.service.clusterPort }}
+      targetPort: clusterpeer
       protocol: TCP
       name: cluster-tcp
     - port: {{ .Values.service.clusterPort }}
-      targetPort: {{ .Values.service.clusterPort }}
+      targetPort: clusterpeer
       protocol: UDP
       name: cluster-udp
     {{- end }}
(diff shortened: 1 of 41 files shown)


sam-mcbr:
@monotek @naseemkullah Don't mean to be a bother, but I would love to try to get a fix in so I can run in HA quickly so I can deploy this. I'm more than happy to make any necessary edits/improvements. Just wanted to give a heads up that this is a priority for me 😄

sam-mcbr:
And apologies, I did just notice a few other config options that would be useful so I added those. But this is ready to go now, thanks!

naseemkullah (approved):
Lgtm

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -97,9 +97,20 @@ spec:
         - name: {{ .Chart.Name }}-{{ .Values.configmapReload.name }}
           image: "{{ .Values.configmapReload.image.repository }}:{{ .Values.configmapReload.image.tag }}"
           imagePullPolicy: "{{ .Values.configmapReload.image.pullPolicy }}"
+          {{- with .Values.configmapReload.extraEnv }}
+          env:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           args:
+            {{- if not (hasKey .Values.configmapReload.extraArgs "config-file") }}
```
naseemkullah, line 105: Config-file vs watched-dir? Why's that again?

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -153,6 +169,11 @@ spec:
             - name: http
               containerPort: 9093
               protocol: TCP
+            {{- if or (gt (int .Values.replicaCount) 1) (.Values.additionalPeers) }}
```
naseemkullah, line 172: Apologies I'm a bit out of the loop. Why are there two ways to set this again? What happens if you define additional peers but replicas is 1 and vice versa?

charts/alertmanager/templates/statefulset.yaml
Sammcb: Hi! (Sorry, not at my work computer so replying from another account). The config-reloader can watch individual files, and then populate environment variables and output to a new file. I just wanted more flexibility about the options passed to the config reloader in general.

charts/alertmanager/templates/statefulset.yaml
Sammcb: Not sure, just followed the current structure for supporting HA. Values file didn't specify how additional peers could be used.

charts/alertmanager/templates/services.yaml
```diff
@@ -47,13 +50,16 @@ spec:
       name: http
     {{- if or (gt (int .Values.replicaCount) 1) (.Values.additionalPeers) }}
     - port: {{ .Values.service.clusterPort }}
-      targetPort: {{ .Values.service.clusterPort }}
+      targetPort: clusterpeer
```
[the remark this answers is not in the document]
Sammcb: Just curious, how would this be a breaking change? I did add a named port to the StatefulSet, and thought since the service port can be changed it would just connect to the names STS port (which can also have its value set). This matches what was done for the regular http port. Happy to change it back to the `clusterPort` though

charts/alertmanager/templates/statefulset.yaml
naseemkullah: np, but is there a use case where user passes a `watched-dir` key to `extraArgs` ? if so how would that be handled here

Sammcb: Yeah, that wouldn't work. I can make a change to check and see if that value has been set as well.

charts/alertmanager/templates/services.yaml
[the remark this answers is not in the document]
Sammcb: Actually, @monotek or @naseemkullah can you provide a little more insight into why both a TCP and UDP port was added here originally? Nothing was exposed on the STS except the `http` port, but I did add the `clusterpeer` port in this MR to match was this service was expecting. However, I just noticed that I only added a TCP port on the STS. Is a UDP port necessary as well? And @monotek if you could please provide more detail on why targeting a named port here would be a breaking change that would be wonderful (I thought that even though the named port on the pod is static, it won't matter since the port it's set to and exposed on the service is the cluster port value). Thanks so much!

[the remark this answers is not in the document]
Sammcb: Looking back at the [Alertmanager chart in the Prometheus chart](https://github.com/prometheus-community/helm-charts/tree/prometheus-15.14.0/charts/prometheus/templates/alertmanager), it looks like all that was done was another `containerPort` was added for the meshpeer and then that was made available on the headless service. Do you all have any hesitations to this approach (I can make the changes tomorrow)? Thanks again!

charts/alertmanager/templates/statefulset.yaml
sam-mcbr: Thanks for waiting, just made the change to allow overriding `watched-dir`!

charts/alertmanager/templates/services.yaml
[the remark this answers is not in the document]
sam-mcbr: So this is just a named port referencing the port on the STS. That value is still configurable and the value that the service exposes is still configurable. So then this named port won't matter right? (Just internal connection between the service and the STS, like was done for the `http` port)

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -153,6 +169,14 @@ spec:
...
+            - name: clusterpeer-tcp
+              containerPort: {{ $svcClusterPort }}
```
sam-mcbr, line 174: @monotek Here is where the custom value would apply on the STS.

charts/alertmanager/templates/services.yaml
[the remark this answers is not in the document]
sam-mcbr: Hi @monotek! I would love to keep this PR moving, so if you or @naseemkullah have any more suggestions I would be more than happy to implement those and try to get this merged sometime soon. Thanks 😄

monotek (approved):

sam-mcbr:
Thank you all so much!

PR #3627, opened 2023-07-22
[prometheus-snmp-exporter] Add extra manifests template to snmp-exporter chart

walker-tom:
Adds a value to the `prometheus-snmp-exporter` chart to allow you to deploy additional Kubernetes resources, such as OpenShift Routes, as part of this Helm Chart. This is the same as what has been done for the `prometheus` chart among others.

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 1.6.0
+version: 1.7.0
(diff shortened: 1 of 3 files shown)


xiu (approved):
I have nothing against it as such but wouldn't it be better to use this chart as a subchart and add additional resources in the main one?

walker-tom:
You could definitely do that instead, the downside being that then you have to maintain the parent chart separately but for larger sets of extra resources it is likely the better way to go. I'm using this chart in Argo CD and reference the chart repository directly, which means I get latest changes automatically, so this way I can just extend my values file and keep this workflow. This also keeps it in sync with the other charts in this repo.

PR #3649, opened 2023-07-31
[prometheus-rabbitmq-exporter] fix connection parameters not passing to the deployment

Dgotlieb:
The parameters of `rabbitmq.user`, `rabbitmq.password` and `rabbitmq.url` didn't pass to the deployment's env variables,
which cause getting 401 (as it used the defaults)
``` yaml
rabbitmq:
  url: http://myrabbit:15672
  user: guest
  password: guest
```

``` bash
time="2023-07-31T12:47:21Z" level=error msg="Error while retrieving data from rabbitHost" error="<nil>" host="http://myrabbit:15672" statusCode=401
```

The chart never worked before while passing the values from the `values.yaml` file.
In a roundabout way it fixed [this issue](https://github.com/kbudde/rabbitmq_exporter/issues/343)

#### Checklist

- [X] Chart Version bumped


charts/prometheus-rabbitmq-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Rabbitmq metrics exporter for prometheus
 name: prometheus-rabbitmq-exporter
-version: 1.8.0
+version: 1.9.0
```
monotek, line 4: if this is a fix it should only raise the patch version.

```suggestion
version: 1.8.1
```

monotek (changes requested):

monotek (approved):

PR #3656, opened 2023-08-02
[kube-prometheus-stack] Add support for PDB apiVersion to be specified

### LINKED ISSUE #3655: [kube-prometheus-stack] PDB uses wrong api version

After k8s upgrade from 1.24 to 1.25 there are some api deprecations. One such [deprecation](policy/v1beta1: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125) is PodDisruptionBudget apiVersion

All across the project, we use Capabilities [build it object](https://helm.sh/docs/chart_template_guide/builtin_objects/) from helm.
The problem arrises when trying to install the chart with tools that don't use helm instal, but helm template and kubectl apply. On helm template the Capabilities feature uses k8s client version which might be different from the remote one.

Allow apiVersion overriding from values.yaml

None I can think of

### Additional context.

_No response_

titirigaiulian:

- fixes #3655 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -239,6 +239,7 @@ alertmanager:
   ##
   podDisruptionBudget:
     enabled: false
+    apiVersion: {}
     minAvailable: 1
     maxUnavailable: ""
 
@@ -2630,6 +2631,7 @@ prometheus:
   ##
   podDisruptionBudget:
     enabled: false
+    apiVersion: {}
     minAvailable: 1
     maxUnavailable: ""
 
@@ -3656,6 +3658,7 @@ thanosRuler:
   ##
   podDisruptionBudget:
     enabled: false
+    apiVersion: {}
     minAvailable: 1
     maxUnavailable: ""
 
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -239,7 +239,7 @@ alertmanager:
   ##
   podDisruptionBudget:
     enabled: false
-    apiVersion: {}
+    apiVersion: ""
     minAvailable: 1
     maxUnavailable: ""
 
@@ -2631,7 +2631,7 @@ prometheus:
   ##
   podDisruptionBudget:
     enabled: false
-    apiVersion: {}
+    apiVersion: ""
     minAvailable: 1
     maxUnavailable: ""
 
@@ -3658,7 +3658,7 @@ thanosRuler:
   ##
   podDisruptionBudget:
     enabled: false
-    apiVersion: {}
+    apiVersion: ""
     minAvailable: 1
     maxUnavailable: ""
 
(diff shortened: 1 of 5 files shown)


QuentinBisson:
Can you please sign the DCO?

titirigaiulian:
> Can you please sign the DCO?

done. Ready for review

charts/kube-prometheus-stack/values.yaml
```diff
@@ -2630,6 +2631,7 @@ prometheus:
   ##
   podDisruptionBudget:
     enabled: false
+    apiVersion: {}
```
QuentinBisson, line 2634: Should this not be a string? And now that I think about this, why do you need to be able to configure the PDB version?

GMartinez-Sisti: > And now that I think about this, why do you need to be able to configure the PDB version?

I was wondering that too, but we're already doing it on the chart:

- https://github.com/prometheus-community/helm-charts/blob/39299ac90a3c8bc9912ca7ca289788556b5aa007/charts/kube-prometheus-stack/templates/alertmanager/ingress.yaml#L10
- https://github.com/prometheus-community/helm-charts/blob/39299ac90a3c8bc9912ca7ca289788556b5aa007/charts/kube-prometheus-stack/templates/prometheus/ingress.yaml#L9

GMartinez-Sisti: Totally missed that! I see that this PR is trying to fix an issue when the chart isn't deployed using helm (https://github.com/prometheus-community/helm-charts/issues/3655). This looks like an edge case :/. The [_helpers.tpl](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl) file uses the same logic to detect other api versions, so updating this logic means we would have to allow the same for all of the other apis.

This can be easily overcome using the parameter [`--kube-version`](https://helm.sh/docs/helm/helm_template/) on `helm template`. Leaving a comment with this info on the [issue](https://github.com/prometheus-community/helm-charts/issues/3655).

titirigaiulian: Hi, thanks for the review.

titirigaiulian: Hi, is there any chance to have this reviewed/approved? It will help us unblock some upgrades

GMartinez-Sisti: We might have something better for this. I see you have to duplicate the API version key on each of the workload PDBs, and you might end up with the same duplication if you need it for ingress or other type of resource. There is the `kubeTargetVersionOverride` root key that should address this.

What do you think of adding logic to respect this key? Then you only need to change a single key.

titirigaiulian: Thanks for the input @GMartinez-Sisti. I think the actual key used is `kubeVersionOverride`, but for some reason it didn't work. I'll close the PR and issue, as we fixed it on the tooling side with a client version upgrade that covers the majority of usecases.

GMartinez-Sisti: Thank you for the feedback and understanding 👍

PR #3663, opened 2023-08-03
[kube-prometheus-stack] Quote enums for consistency

bsloeserwij:
Applying the crds with the k8s module of Ansible failed due to a contructor error. It turned out to be unquoted equal signs. It is also more consistent to quote these signs.

Quotes are added to the equal signt in crd-alertmanagerconfigs.yaml. This will fix the appliance of the crd using the ansible module kubernetes.core.k8s (which uses PyYAML under the hood) which errors if these equal signs are not quoted.

No issue reported

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 48.2.3
+version: 48.2.4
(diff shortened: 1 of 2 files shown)


QuentinBisson:
Can you add those quotes in the source repo instead as this change will be overriden https://github.com/prometheus-operator/prometheus-operator/tree/main/example/prometheus-operator-crd?

bsloeserwij:
Done, see https://github.com/prometheus-operator/prometheus-operator/pull/5869

bsloeserwij:
I will close this PR, since the change should be done somewhere else.

PR #3670, opened 2023-08-04
[kube-state-metrics]fix: add namespaceSelector to kube-state-metrics …

oludayofafiolu:
…servicemonitor

- fixes #

#### Checklist

- [ ] Chart Version bumped


dotdc:
I'm a bit scared that there might be cases where this could break things for some user. (I don't have anything in mind, just a concern)

@oludayofafiolu What do you think about adding this as a new feature instead?

```
prometheus:
  monitor:
    namespaceSelector: {}
```

```
{{- if .Values.prometheus.monitor.namespaceSelector }}
  namespaceSelector:
    matchNames:
    {{- with .Values.prometheus.monitor.namespaceSelector }}
      {{- toYaml . | nindent 6 }}
    {{- end }}
{{- end }}
```

My code hasn't been tested yet.

PS: sorry for the delay!

oludayofafiolu:
Yeah I'm okay with adding it as a feature 👍🏾

dotdc:
> Yeah I'm okay with adding it as a feature 👍🏾, matchNames should be a list as stated in the operator [docs](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#namespaceselector)
> 
> How about this :
> 
> ```
>   namespaceSelector:
>     matchNames:
>     {{- with .Values.prometheus.monitor.namespaceSelector }}
>       {{- toYaml . | nindent 6 }}
>     {{- else }}
>       any: false
>     {{- end }}
> ```
> 
> According to the docs "If matchNames is empty and any is false, it means that the objects are selected from the current namespace.", This means we don't need to have {{ .Release.Namespace }}

I think an if statement like in my example would match the file's code style better: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-state-metrics/templates/servicemonitor.yaml

The chart version also needs to be bumped.

mrueg (approved):

dotdc (approved):

PR #3677, opened 2023-08-06
[prometheus-node-exporter] feat: allow changing the default revisionHistoryLimit

tberreis:
You may want to clean up old ReplicaSets so that these don't appear in metrics, OPA violations, ArgoCD UI and so on.
Therefore you want to overwrite the Kubernetes default revision history limit.

The Prometheus Node Exporter DaemonSet has been adapted to allow the user changing the default revisionHistoryLimit.

N/A

N/A

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.21.0
+version: 4.22.0
 appVersion: 1.6.0
 home: https://github.com/prometheus/node_exporter/
 sources:
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -13,6 +13,7 @@ spec:
   selector:
     matchLabels:
       {{- include "prometheus-node-exporter.selectorLabels" . | nindent 6 }}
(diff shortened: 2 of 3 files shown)


rouke-broersma:
@tberreis fyi we are running a custom admission controller that sets the revision history limit to 0 for all deployments etc in the cluster. It would have been better if instead of forcing the revision history limit to 10 you would have made it so the revision history limit is only set if set by the user. Now we are getting sync-loops in argocd until we set the revision history limit to 0 manually in the values. Not a big deal but setting the default in the chart is kind of a breaking change and it's not really neccesary.

tberreis:
> @tberreis fyi we are running a custom admission controller that sets the revision history limit to 0 for all deployments etc in the cluster. It would have been better if instead of forcing the revision history limit to 10 you would have made it so the revision history limit is only set if set by the user. Now we are getting sync-loops in argocd until we set the revision history limit to 0 manually in the values. Not a big deal but setting the default in the chart is kind of a breaking change and it's not really neccesary.

Totally agree! That's how I started with this topic but on some repositories I had to remove the conditional (see https://github.com/kubereboot/charts/pull/47#pullrequestreview-1560408741 for example).
Therefore I decided to use the Kubernetes defaults here. Sorry to hear that it broke your customization.

rouke-broersma:
It's not a big deal and I appreciate that someone is going around making the revision history limit configurable 😁

PR #3689, opened 2023-08-10
[alertmanager] Add optional seperate ingress per replica

knutgoetz:

This pull request adds the ability to add a separate ingress for each Alertmanager replica. This allows for the replicas to be exposed and used as a highly available cluster by Prometheus instances outside the cluster.

This feature was [already implemented in the kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/commit/6b8379e817013ba945ef7945101bb294cfcea367) chart and has been adapted for this implementation. There are a few key differences to note:

- This implementation does not support Ingress `apiVersion < networking.k8s.io/v1`
- You cannot set a static nodePort for the services per replica. This is because statically assigning the port can lead to collisions when using more than one replica. If you are using only one replica, you can use the default service.
#### Checklist

- [x] Chart Version bumped


monotek:
Can't you use a new k8s service which uses one of the pods?
So only one ingress would be needed.

knutgoetz:
Not sure, if i understand your suggestion correctly.
I need to send an alert from prometheus  to _every_ alertmanager  pod to achieve high availability for the alertmanager cluster.  It does not suffice to send the alert only to of the replicas, i.e. use simply one ingress+service which does the load balancing. That's why an extra service per replica is needed.

 I could reach each of these services via one ingress with stuff like path based routing (different path per service, then deleting the path prefix) or with an ingress controller which supports mirroring the request to multiple services but this would need some additional work and relies on the specific ingress controller.
I thought that having one ingress per service is the simplest way to go.

monotek (dismissed):

monotek (approved):

PR #3695, opened 2023-08-16
[prometheus] Update external_labels

tinhtq:
Update external_labels for prometheus helm chart


charts/prometheus/values.yaml
```diff
@@ -219,6 +219,12 @@ server:
   storagePath: ""
 
   global:
+    ## The labels to add to any time series or alerts when communicating with
+    ## external systems (federation, remote storage, Alertmanager).
+    external_labels: []
+    # - key: value
+    # - app: test
+
```
zanhsieh, line 227: This won't work. `$.Values.server.global.external_labels` or `.Values.server.global.external_labels` does not get referred to where labels get customerized.

zanhsieh (changes requested):

PR #3699, opened 2023-08-17
[alertmanager] Add common meta labels

benjaminjb:

This PR adds the ability to add custom labels to the Kubernetes objects defined by this Helm chart. It does this by 

1) adding `extraLabels` to the `alertmanager.labels` template (and to the `values.yaml`); and 
2) adding `alertmanager.labels` to the PVC defined by the sts.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,8 +6,8 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.6.0
-appVersion: v0.27.0
+version: 1.7.0
+appVersion: v0.26.0
 kubeVersion: ">=1.19.0-0"
 keywords:
   - monitoring
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.5.0
+version: 1.5.1
 appVersion: v0.25.0
 kubeVersion: ">=1.19.0-0"
 keywords:
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.5.1
+version: 1.6.0
 appVersion: v0.25.0
 kubeVersion: ">=1.19.0-0"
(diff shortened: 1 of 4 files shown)


charts/alertmanager/Chart.yaml
```diff
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.5.0
+version: 1.5.1
```
monotek, line 9: adding a feature should raise minor version

```suggestion
version: 1.6.0
```

charts/alertmanager/templates/_helpers.tpl
```diff
@@ -41,6 +41,9 @@ helm.sh/chart: {{ include "alertmanager.chart" . }}
 app.kubernetes.io/version: {{ . | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- with .Values.commonMetaLabels}}
```
monotek, line 44: please rename to extraLabels:
```suggestion
{{- with .Values.extraLabels }}
```

Please also add "extraLabels" to values.yaml.

monotek (changes requested):

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -228,6 +228,8 @@ spec:
   volumeClaimTemplates:
     - metadata:
         name: storage
+        labels:
+          {{- include "alertmanager.labels" . | nindent 10 }}
```
zeritti, line 232: Setting dynamic labels in `statefulset.spec.volumeClaimTemplates` like `app.kubernetes.io/version` will prevent an upgrade unless the statefulset gets deleted. This is because those fields are immutable and cannot be patched. Labels in such a case have to be fairly static, e.g. corresponding to the selector labels. It is useful to define a set of labels specifically for statefulset's persistent volume claims if such labels are desired.

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -228,6 +228,10 @@ spec:
   volumeClaimTemplates:
     - metadata:
         name: storage
+        labels:
+          {{- if .Values.persistence.labels }}
```
monotek, line 232: why are extralabels not added here?

monotek (changes requested):

charts/alertmanager/templates/statefulset.yaml
benjaminjb: Do we want this to be `extraLabels` + `persistance.labels`? That way it would be easy to add labels to everything (which is what `extraLabels` should do) and also easy to specify labels just for the PVC. 

(I don't have a use-case for that last part -- I just want a few labels on everything created by this chart. But some other charts in this repo, e.g., Prometheus, allow a lot of label customization, so I'm happy to add some customization of the PVC here.)

PR #3725, opened 2023-08-25
[kube-prometheus-stack] Default port on which ETCD exposes its metrics

xogoodnow:
By default ETCD exposes its metrics on port 2379 and not 2381

By default ETCD exposes its metrics on port 2379 and not 2381
No service (by default) is listening on port 2381 on the cluster, so after installing the chart, the metrics for etcd would not be monitored unless the port in values.yaml (or ETCD configuration) is changed.

- fixes #
Changing eh default port for etcd metrics from 2381 to 2379 in the following path
kubeEtcd.service. port
AND
kubeEtcd.service. targetPort
#### Checklist

- [x] Chart Version bumped


QuentinBisson (changes requested):
The helm chart prefers to enable secure settings per default and port 2381 is etcd TLS port that is used to fetch metrics. It is possible that your clusters are either old or not providing a secure access to etcd in which case you need to configure the port to 2379 but that should only be the case for your clusters.

GMartinez-Sisti:
Hi @xogoodnow, thank you for the contribution. Please check https://github.com/prometheus-community/helm-charts/pull/2264 regarding the ETCD ports.

TLDR: https://github.com/kubernetes/kubernetes/blob/8b132ea40a250e9acf08ab2fadb9a8c2fd811d74/cmd/kubeadm/app/constants/constants.go#L95-L96

xogoodnow:
Hi @GMartinez-Sisti , Just deploy the latest version of the cluster and install the chart, it does not work.
I just tested it on our stage and it did not work. (if you have the time I recommend you just test this with the latest version of k8s 1.28 and see if it works by default)
cheers

zeritti:
My impression is that introducing 2381 as an Etcd metrics port in the chart was based on a specific internal change in _kubeadm_, judging from the link to its constants above and in the original PR. From Etcd [metrics-endpoint](https://etcd.io/docs/v3.5/op-guide/monitoring/#metrics-endpoint)

> Each etcd server exports metrics under the /metrics path on its client port and optionally on locations given by --listen-metrics-urls.

The [client port](https://etcd.io/docs/v3.5/op-guide/configuration/) is indeed 2379/tcp and supposed to always be protected by TLS.

_kubeadm_ introduced port 2381 in a [change](https://github.com/kubernetes/kubernetes/commit/99b64f1e8e93b254aa3435c5a52e283391fb1d95) to access the Etcd's `/health` endpoint for health checks/liveness probes _without_ TLS replacing thus the previous health check implementation. It makes use of the Etcd flag above that can enable additional URLs serving `/metrics` and `/health`. Furthermore, as the health check is insecure, it opens the port at [localhost](https://github.com/kubernetes/kubernetes/blob/3cf3702d1e638548ba9fcfd957ebae43e1a573cb/cmd/kubeadm/app/phases/etcd/local.go#L261) - it is not supposed to be accessed from outside.

The changes in _kubeadm_ have no effect on how Etcd primarily serves its metrics: through the client port.

xogoodnow:
Hi, considering that this would be a breaking change, do we have any estimation on how and when to make these changes on the chart?

jkroepke:
Bump the helm major chart version + upgrade nodes on the README.md is the way to go for an breaking change and needs to be part of the PR itself.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -941,7 +941,17 @@ grafana:
...
+  ##
+  # persistence:
+  #   enabled: true
+  #   type: sts
+  #   storageClassName: <StorageClassName>
+  #   accessModes:
+  #     - ReadWriteOnce
+  #   size: 20Gi
+  #   finalizers:
+  #     - kubernetes.io/pvc-protection
```
GMartinez-Sisti, line 954: These shouldn't be here. Can you please remove them?

GMartinez-Sisti (changes requested):
As @jkroepke said, please do:

> Bump the helm major chart version + upgrade notes on the README.md is the way to go for an breaking change and needs to be part of the PR itself.

xogoodnow:
Hi
Should something be done on my end?

jkroepke:
> Hi Should something be done on my end?

Yes, please take note of the review comment from @GMartinez-Sisti

xogoodnow:
hi @jkroepke,
the changes have been made
Please check and if there is something that I must do , let me know

jkroepke:
Hi @xogoodnow 

currently, I'm missing everything what is requested.

PR #3754, opened 2023-09-04
[prometheus-kafka-exporter] add support for valueFrom in env variables

marcfrederick:

Adds support for non key-value environment variables to the prometheus-kafka-exporter chart. This makes it possible mount keys Secrets/ConfigMaps into the container and use the values in the args.

This makes it possible to e.g. use `kafkaServers` from a ConfigMap:

```yaml
env:
  - name: KAFKA_SERVER
    valueFrom:
      configMapKeyRef:
        name: kafka-server
        key: endpoint
kafkaServer:
  - $(KAFKA_SERVER)
```

@gkarthiks @golgoth31 @zeritti 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-kafka-exporter/templates/deployment.yaml
+++ b/charts/prometheus-kafka-exporter/templates/deployment.yaml
@@ -81,9 +81,8 @@ spec:
           {{- end }}
           {{- end }}
           env:
-          {{- range $env := .Values.env }}
-            - name: {{ $env.name }}
-              value: {{ $env.value }}
+          {{- with .Values.env }}
+            {{- toYaml . | nindent 12 }}
           {{- end }}
           {{- if .Values.sasl.scram.enabled }}
             - name: SCRAM_USERNAME
--- a/charts/prometheus-kafka-exporter/Chart.yaml
+++ b/charts/prometheus-kafka-exporter/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v1.6.0"
 description: A Helm chart to export the metrics from Kafka in Prometheus format using the kafka-exporter from https://github.com/danielqsj/kafka_exporter
 name: prometheus-kafka-exporter
 home: https://github.com/danielqsj/kafka_exporter
-version: 2.3.0
+version: 2.4.0
 kubeVersion: ">=1.19.0-0"
 sources:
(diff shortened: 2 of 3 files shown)


charts/prometheus-kafka-exporter/templates/deployment.yaml
```diff
@@ -80,10 +80,9 @@ spec:
             - {{ . }}
           {{- end }}
           {{- end }}
+          {{- with .Values.env }}
           env:
-          {{- range $env := .Values.env }}
-            - name: {{ $env.name }}
-              value: {{ $env.value }}
+            {{- toYaml . | nindent 12}}
```
zeritti, line 85: Thank you, @marcfrederick, for your PR. 

```suggestion
            {{- toYaml . | nindent 12 }}
```

I would suggest to add a comment/example in the values file that e.g. a secret/cm can also be used. For example

```yaml
  ## env sets environment variables to pass to the container.
  ## Can be set as name/value pairs, read from secrets or configmaps.
  env: []
    # - name: <ENV_VAR_NAME>
    #   value: <value>
    # - name: PASSWORD
    #   valueFrom:
    #     secretKeyRef:
    #       name: mysecret
    #       key: password
    #       optional: false
```

marcfrederick: I've pushed an updated commit incorporating these changes

charts/prometheus-kafka-exporter/templates/deployment.yaml
```diff
@@ -80,10 +80,9 @@ spec:
             - {{ . }}
           {{- end }}
           {{- end }}
+          {{- with .Values.env }}
           env:
```
zeritti, line 84: This line would disable the whole field if `.Values.env` were null. We would not like to do that since some variables are already defined in the template. One way would be starting below the line:

```suggestion
          env:
          {{- with .Values.env }}
```

marcfrederick: Good point, sorry I completely missed the other two variables. I've updated the PR accordingly and squashed the commits.

zeritti (approved):

PR #3762, opened 2023-09-05
[kube-prometheus-stack] fix reloader-web's app protocol

abaguas:

To integrate this port with applications such as istio the applicationProtocol should be set explicitly.
One solution to enable this integration is to add an `appProtocol` attribute to the port as explained in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) and implemented in this PR. Another alternative is to prefix the port with `http-`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/service.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/service.yaml
@@ -42,8 +42,11 @@ spec:
     {{- end }}
     port: {{ .Values.prometheus.service.port }}
     targetPort: {{ .Values.prometheus.service.targetPort }}
-  - appProtocol: http
-    name: reloader-web
+  - name: reloader-web
+    {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+    {{- if semverCompare "> 1.20.0-0" $kubeTargetVersion }}
+    appProtocol: http
+    {{- end }}
     port: 8080
     targetPort: reloader-web
   {{- if .Values.prometheus.thanosIngress.enabled }}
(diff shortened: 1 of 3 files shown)


abaguas:
If you prefered I can rename the port to `http-reloader-web` or add the `appProtocol` to the other `http` ports

GMartinez-Sisti:
Hi @abaguas! Thank you for the contribution 🙏 

Looks like the `appProtocol` feature graduated on K8S `1.20`, since we support `1.19` it would be nice to have some validation before adding it. Maybe something like this to append it conditionally:

https://github.com/prometheus-community/helm-charts/blob/3352ef86286a7e8f07266ff3da1a575aa7368c8f/charts/kube-prometheus-stack/templates/_helpers.tpl#L170-L179

WDYT?

abaguas:
Makes sense, thank you for the tip. I implemented it and tested with

```
$ helm package . && helm template kube-prometheus-stack kube-prometheus-stack-50.3.2.tgz -n monitoring --set kubeTargetVersionOverride="1.19.0" > manifests.yaml      
$ helm package . && helm template kube-prometheus-stack kube-prometheus-stack-50.3.2.tgz -n monitoring --set kubeTargetVersionOverride="1.20.0" > manifests-after.yaml
$ diff manifests.yaml manifests-after.yaml 
1389a1390
>     appProtocol: http
1584a1586
>     appProtocol: http
```

charts/kube-prometheus-stack/templates/prometheus/service.yaml
```diff
@@ -43,6 +43,10 @@ spec:
     port: {{ .Values.prometheus.service.port }}
     targetPort: {{ .Values.prometheus.service.targetPort }}
   - name: reloader-web
+    {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
```
QuentinBisson, line 46: Could you move this at the top?

QuentinBisson: Otherwise looks good

QuentinBisson (dismissed):

QuentinBisson:
@abaguas Can you sign your commits?

QuentinBisson (approved):

PR #3766, opened 2023-09-07
[kube-state-metrics] add option to set updateStrategy to Recreate

### LINKED ISSUE #3764: [kube-state-metrics] Option to set strategy to Recreate

Hi,

I want to set the Update strategy to `Recreate`. Right now the default option is `RollingUpdate`. I want to change that but looks like I cannot with the current chart and templates.

I tried to find the `strategy` option in the `values.yaml` but I could not find it.

I also checked the [deployment template](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-state-metrics/templates/deployment.yaml) but I was not able to find it there too. 

### What's your helm version?

version.BuildInfo{Version:"v3.12.1", GitCommit:"f32a527a060157990e2aa86bf45010dfb3cc8b8d", GitTreeState:"clean", GoVersion:"go1.20.4"}

Client Version: v1.26.5, Kustomize Version: v4.5.7, Server Version: v1.26.5-gke.1400

### Which chart?

kube-state-metrics

### What's the chart version?

kube-state-metrics-5.11.1

### What happened?

When I have to upgrade the chart, a new pod is created for the `kube-state-metrics` deployment first and then the old pod is terminated. That results in 2 pods running at the same time for sometime. I don't want to run 2 pods at the same time.

Why I don't want is because that messes up the whole queries I am running. Why it messes it is because prometheus then scrapes 2 pods for the same metrics but with different `instance` label. Due to that, my queries start failing.

This is the error I start getting in my grafana dashboards.

```
Status: 500. Message: execution: found duplicate series for the match group {namespace="default", pod="my-pod-1"} on the right hand-side of the operation
```

I expected that the option to configure the deployment strategy to `Recreate` would be available in the helm values file.

### How to reproduce it?

Just try to upgrade the chart. That should create 2 pods at one time, which is exactly what I don't want.

NONE

```
helm upgrade kube-state-metrics prometheus-community/kube-state-metrics -n monitoring -f kube-state-metrics.yaml
```

I am fine with the `kube-state-metrics` going down for 10-20, even 60 seconds. But I don't want 2 kube-state-metrics pods running at the same time.

luvpreetsingh:

This PR includes changes which will enable us to change the Deployment Strategy for kube-state-metrics when using the `Kind: Deployment`.

- fixes #3764

N/A

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - prometheus
 - kubernetes
 type: application
-version: 5.12.1
+version: 5.13.0
 appVersion: 2.10.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -18,6 +18,10 @@ spec:
     matchLabels:
       {{- include "kube-state-metrics.selectorLabels" . | indent 6 }}
   replicas: {{ .Values.replicas }}
+  {{- if not .Values.autosharding.enabled }}
+  strategy:
+    type: {{ .Values.updateStrategy | default "RollingUpdate" }}
+  {{- end }}
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
   {{- if .Values.autosharding.enabled }}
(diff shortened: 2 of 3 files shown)


dotdc (approved):

dotdc:
Thank you for your contribution @luvpreetsingh !

PR #3772, opened 2023-09-10
[prometheus]: allow more flexible poddisruptionbudget config

starlightromero:

Allow for more flexible podDisruptionBudget configuration in the prometheus Helm chart. This specifically allows  the use of `minAvailable` on the `server`

- fixes #

#### Checklist

- [x] Chart Version bumped


charts/prometheus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: v2.46.0
-version: 24.3.0
+version: 24.3.1
```
zeritti, line 4: Thank you, @starlightromero, for your PR.

As presented, it would introduce an incompatible change and would therefore require bumping the major number. I would suggest that we keep the toggle avoiding thus a breaking change.

charts/prometheus/templates/pdb.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.server.podDisruptionBudget.enabled }}
+{{- if .Values.server.podDisruptionBudget }}
```
zeritti, line 1: This will allow us to keep the toggle whilst enabling other spec fields.

```suggestion
{{- if .Values.server.podDisruptionBudget.enabled }}
{{- $pdbSpec := omit .Values.server.podDisruptionBudget "enabled" }}
```

charts/prometheus/templates/pdb.yaml
```diff
@@ -7,8 +7,8 @@ metadata:
   labels:
     {{- include "prometheus.server.labels" . | nindent 4 }}
 spec:
-  maxUnavailable: {{ .Values.server.podDisruptionBudget.maxUnavailable }}
   selector:
     matchLabels:
       {{- include "prometheus.server.matchLabels" . | nindent 6 }}
+{{ toYaml .Values.server.podDisruptionBudget | indent 2 }}
```
zeritti, line 13:
```suggestion
  {{- toYaml $pdbSpec | nindent 2 }}
```

charts/prometheus/values.yaml
```diff
@@ -371,9 +371,7 @@ server:
   ## PodDisruptionBudget settings
   ## ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
   ##
-  podDisruptionBudget:
-    enabled: false
-    maxUnavailable: 1
+  podDisruptionBudget: {}
```
zeritti, line 374: The field can now include spec as desired (apart from the selector which is already preset in the template).

```suggestion
  podDisruptionBudget:
    enabled: false
    maxUnavailable: 1
    # minAvailable: 1
    ## unhealthyPodEvictionPolicy is available since 1.27.0 (beta)
    ## https://kubernetes.io/docs/tasks/run-application/configure-pdb/#unhealthy-pod-eviction-policy
    # unhealthyPodEvictionPolicy: IfHealthyBudget
```

zeritti (changes requested):

charts/prometheus/Chart.yaml
starlightromero: Good call out! Thanks @zeritti . If there is a way of queueing updates for the next major release, I suggest the code from the original PR be added. This would make the template similar to that of `kube-state-metrics`. I suggest this because it seems easier to maintain if all charts in this repo use a similar pattern

PR #2163, opened 2022-06-16
[kube-prometheus-stack] introduced servicePort field to select ingress target for alertmanager/prometheus

spr-mweber3:
This fixes https://github.com/prometheus-community/helm-charts/issues/1391. We had the same requirement as what was stated in the issue. We wanted to put an oauth2-proxy as sidecar next to Alertmanager and Prometheus. 

The goal was to have the components of the stack communicate unauthenticated but only the UI to be secured. We achieved that by introducing a new field. You can now set `alertmanager.ingress.servicePort` and `prometheus.ingress.servicePort' to point the ingress to an additional port on the service.

If you don't specify a servicePort on the ingress, then it defaults to the behavior it was before. So the change here does not introduce a breaking change.

This has proven to work very well for uns.

Example:

```
alertmanager:
  service:
    additionalPorts:
    - name: authenticated
      port: 8081
      targetPort: 8081
  ingress:
    enabled: true
    servicePort: 8081
  alertmanagerSpec:
    containers:
    - name: oauth2-proxy
      image: quay.io/oauth-proxy/oauth2-proxy:v7.3.0
      args:
      - --upstream=http://127.0.0.1:9093
      - --http-address=0.0.0.0:8081
      - ...
      ports:
      - containerPort: 8081
        name: oauth-proxy
        protocol: TCP
      resources: {}

prometheus:
  service:
    additionalPorts:
    - name: authenticated
      port: 8081
      targetPort: 8081
  ingress:
    enabled: true
    servicePort: 8081
  prometheusSpec:
    containers:
    - name: oauth2-proxy
      image: quay.io/oauth-proxy/oauth2-proxy:v7.3.0
      args:
      - --upstream=http://127.0.0.1:9093
      - --http-address=0.0.0.0:8081
      - ...
      ports:
      - containerPort: 8081
        name: oauth-proxy
        protocol: TCP
      resources: {}

```

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #3779, opened 2023-09-11
[prometheus] Adding additionalPorts and servicePort

mortenbirkelund:

Similar to #2163 for the kube-prometheus-stack, this PR will make it possible to add an oauth proxy as a sidecar container, and change the ingress to point to an additional port in your service, that then goes to your oauth proxy. Nice and simple way of adding oauth to Prometheus. 
No issues opened for this chart, but this have been solved for the kube-prometheus-stack chart, so why not also make it possible in the prometheus chart. 

I have checked the helm chart with my changed, and it seems to work fine, but let me know if any changes is required. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: v2.47.0
-version: 24.4.1
+version: 24.5.0
 kubeVersion: ">=1.19.0-0"
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

charts/prometheus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: prometheus
 appVersion: v2.47.0
-version: 24.4.0
+version: 24.4.1
```
zeritti, line 4: This change should be considered minor:

```suggestion
version: 24.5.0
```

zeritti (changes requested):

charts/prometheus/Chart.yaml
mortenbirkelund: Agree, it should be a minor version.

zeritti (approved):

PR #3828, opened 2023-09-26
[prometheus-kafka-exporter] Add nodePort for external use

KimJoonSeo:
#### What this PR does / why we need it
I added support for the nodePort configuration for external use.
#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #3830, opened 2023-09-27
[prometheus-kafka-exporter] Add LoadBalancerIp

KimJoonSeo:
I added an loadBalancerIP configuration to specify a static IP when using NodePort service type.

- This PR has a similar context with PR #3828 .

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-kafka-exporter/values.yaml
+++ b/charts/prometheus-kafka-exporter/values.yaml
@@ -71,9 +71,10 @@ service:
   port: 9308
   labels: {}
   annotations: {}
-  # Specifies a custom nodePort for external use.
-  # Must be used with service.type=NodePort.
+  # Specifies a custom nodePort for external use. It must be used with service.type=NodePort.
   nodePort: null
+  # Specifies a reserved static IP. It must be used with service.type=LoadBalancer.
+  loadBalancerIP: null
 
 liveness:
   enabled: false
--- a/charts/prometheus-kafka-exporter/values.yaml
+++ b/charts/prometheus-kafka-exporter/values.yaml
@@ -71,7 +71,7 @@ service:
   port: 9308
   labels: {}
   annotations: {}
-  # Specifies a custom nodePort for external use. It must be used with service.type=NodePort.
+  # Specifies a custom nodePort for external use. Used with service.type=NodePort.
   nodePort: null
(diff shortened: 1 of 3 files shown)


zeritti:
> I added an loadBalancerIP configuration to specify a static IP when using NodePort service type.

You probably meant _when using LoadBalancer service type_. Using `spec.type=NodePort` with `spec.loadbalancerIP` won't work as the current code does not allow it. Furthemore, the latter applies only to the LoadBalancer type.

charts/prometheus-kafka-exporter/values.yaml
```diff
@@ -71,9 +71,10 @@ service:
   port: 9308
   labels: {}
   annotations: {}
-  # Specifies a custom nodePort for external use.
-  # Must be used with service.type=NodePort.
+  # Specifies a custom nodePort for external use. It must be used with service.type=NodePort.
   nodePort: null
+  # Specifies a reserved static IP. It must be used with service.type=LoadBalancer.
```
zeritti, line 76:
```suggestion
  # Specifies a reserved static IP. Used with service.type=LoadBalancer.
  # The field is deprecated, implementation-specific annotations should be set instead.
```

charts/prometheus-kafka-exporter/values.yaml
```diff
@@ -71,9 +71,10 @@ service:
   port: 9308
   labels: {}
   annotations: {}
-  # Specifies a custom nodePort for external use.
-  # Must be used with service.type=NodePort.
+  # Specifies a custom nodePort for external use. It must be used with service.type=NodePort.
```
zeritti, line 74: The field need not be used as it is optional.

```suggestion
  # Specifies a custom nodePort for external use. Used with service.type=NodePort.
```

zeritti (changes requested):

KimJoonSeo:
> > I added an loadBalancerIP configuration to specify a static IP when using NodePort service type.
> 
> You probably meant _when using LoadBalancer service type_. Using `spec.type=NodePort` with `spec.loadbalancerIP` won't work as the current code does not allow it. Furthermore, the latter applies only to the LoadBalancer type.

That's correct. I meant to do it when using LoadBalancer service type.

PR #3838, opened 2023-09-29
[prometheus-elasticsearch-exporter]: Fix preStop command

m3adow:

This PR fixes the `preStop` command for the elasticsearch-exporter deployment. Right now, it's using `/bin/bash sleep 20`, but bash does not exist in the container image. Thusly, the fix is to use the existing `/bin/sleep` command to sleep 20 seconds.

#### Checklist

- [X] Chart Version bumped


zeritti (approved):
Thank you, @m3adow, LGTM.

PR #3872, opened 2023-10-07
[prometheus-opencost-exporter] new OpenCost Prometheus exporter chart

mattray:

This PR provides [OpenCost](https://opencost.io/) as a Prometheus exporter for tracking Kubernetes cloud costs. OpenCost is a CNCF Sandbox project that is frequently used as a metrics source for other TSDBs or dashboards. There are [already instructions on manually configuring this](https://www.opencost.io/docs/integrations/opencost-exporter) but we figured it would be useful to make it part of the Prometheus community Helm Charts.

The existing [OpenCost Helm Chart](https://github.com/opencost/opencost-helm-chart) was used as the source for this but it provides much more unnecessary configuration. We'll keep them synced as necessary.

#### Checklist

- [x] Chart Version bumped


charts/prometheus-opencost-exporter/templates/_helpers.tpl
```diff
@@ -0,0 +1,99 @@
...
+{{- if .Values.commonLabels}}
+{{ toYaml .Values.commonLabels }}
+{{- end }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "prometheus-opencost-exporter.selectorLabels" -}}
+app: {{ include "prometheus-opencost-exporter.name" . }}
```
monotek, line 50: was this chart created by the following command?

```
helm create prometheus-opencost-exporter
```

if so, please do it again with a more recent helm version, so the new deafults for labels etc are used.

monotek (changes requested):

charts/prometheus-opencost-exporter/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+appVersion: 1.106.2
+version: 0.1.0
+description: Prometheus OpenCost Exporter
+home: https://github.com/prometheus-community/opencost_exporter
```
mrueg, line 4: This repository does not exist. Will this be created?

mattray: Updating to https://github.com/opencost/opencost-helm-chart and we'll incorporate this over there

mattray:
Updates made based on @monotek and @mrueg's reviews. Thanks!

charts/prometheus-opencost-exporter/Chart.yaml
monotek: but were comes the exporter from?

mattray: It's from the documentation at https://www.opencost.io/docs/integrations/opencost-exporter and the existing general-purpose OpenCost Helm chart: https://github.com/opencost/opencost-helm-chart. I've opened an issue to host it there: https://github.com/opencost/opencost-helm-chart/issues/128

mattray:
@monotek @mrueg did I misunderstand the `home` field? Is this the application (OpenCost at https://github.com/opencost/opencost) or should I maintain an upstream Helm chart and periodically sync to here? I'd prefer this location to be authoritative and not maintain 2 locations.

monotek:
> @monotek @mrueg did I misunderstand the `home` field? Is this the application (OpenCost at https://github.com/opencost/opencost) or should I maintain an upstream Helm chart and periodically sync to here? I'd prefer this location to be authoritative and not maintain 2 locations.

No. The home should be where the docker container comes from.

mattray:
@monotek thanks, updated!

monotek (changes requested):

monotek (dismissed):

PR #3878, opened 2023-10-11
[prometheus-adapter] Use full duration format for default certificate values

maikelvl:

Cert-Manager converts simple duration values to their full format: `1h` ends up as `1h0m0s`

GitOps tools like Argo CD keep seeing this as a change and never reach a 'synced' state because they see this as a diff: `1h0m0s != 1h`. Using the full format fixes that.

#### Checklist

- [x] Chart Version bumped


steven-sheehy (approved):
LGTM

PR #3919, opened 2023-10-19
[prometheus-blackbox-exporter] Allow mounting secret as pods env

mlachevre-pass:
When we query a large amount of site using blackbox sometimes authenticate could be a real problem when we can only mount secret in files. Here we could set it in env and pass variable to query. This value keeps optionnal.

#### Checklist

- [x] Chart Version bumped


mlachevre-pass:
Rebase done since bump chart version has been made

charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 8.6.1
+version: 8.6.2
```
monotek, line 4: new feature should raise minor version....

monotek (changes requested):

charts/prometheus-blackbox-exporter/Chart.yaml
mlachevre-pass: sorry, bad bump version on my side

charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 8.6.1
+version: 8.7.1
```
monotek, line 4:
```suggestion
version: 8.7.0
```

monotek (changes requested):

charts/prometheus-blackbox-exporter/Chart.yaml
mlachevre-pass: stupid action of my part, i had your commit

monotek (approved):

Issue #2483, opened 2022-09-23
[prometheus-blackbox-exporter] Adding NET_RAW capability is not enough to run rootless container


zerkms:
### Describe the bug a clear and concise description of what the bug is.

If just `securityContext` container added with `NET_RAW` capability it wouldn't be enough to run rootless.

Prober would fail with the following logs:

```
ts=2022-09-23T03:40:29.727021679Z caller=main.go:144 module=icmp target=xtr-srv05-idrac.wialus.co.nz level=info msg="Creating socket"
ts=2022-09-23T03:40:29.727065001Z caller=main.go:144 module=icmp target=xtr-srv05-idrac.wialus.co.nz level=debug msg="Unable to do unprivileged listen on socket, will attempt privileged" err="socket: permission denied"
ts=2022-09-23T03:40:29.727094666Z caller=main.go:144 module=icmp target=xtr-srv05-idrac.wialus.co.nz level=error msg="Error listening to socket" err="listen ip4:icmp 0.0.0.0: socket: operation not permitted"
ts=2022-09-23T03:40:29.727127158Z caller=main.go:352 module=icmp target=xtr-srv05-idrac.wialus.co.nz level=error msg="Probe failed" duration_seconds=0.049229119
```

What it's missing is setting `net.ipv4.ping_group_range` too in the pod's `securityContext`:

```yaml
      securityContext:
        sysctls:
          - name: net.ipv4.ping_group_range
            value: "0 2147483647"
```

See also: https://github.com/prometheus/blackbox_exporter/issues/147#issuecomment-901302559

### What's your helm version?

n/a

### What's your kubectl version?

n/a

### Which chart?

prometheus-blackbox-exporter

### What's the chart version?

7.1.0

### What happened?

.

### What you expected to happen?

.

### How to reproduce it?

.

### Enter the changed values of values.yaml?

.

### Enter the command that you execute and failing/misfunctioning.

.

zerkms:
Not stale.

zerkms:
Not stale.

fgionghi:
We are having the same issue.

zerkms:
Not stale.

zerkms:
Not stale.

gkavboy:
actual

mossad-zika:
Not stale.

akash123-eng:
Not stale

akash123-eng:
We are also facing this issue.any update on this ?

taind772:
[...]

Issue #2360, opened 2022-08-10
[prometheus-blackbox-exporter] New icmp handling with capabilities as documented breaks icmp checks


sfudeus:
### Describe the bug a clear and concise description of what the bug is.

Just some hours ago, version 7.0.0 of prometheus-blackbox-exporter was released. In that, the capabilities handling for ICMP and the NET_RAW capability was changed (and documented). 
The guided way of just adding the NET_RAW capability is not sufficient (at least for me in my k3s cluster). It is still not possible to do icmp checks (with user still set to 1000 and without granting privileged). 
What does work and could be a better way (at least for k8s>1.18) is to suggest
```
podSecurityContext:
  sysctls:
  - name: net.ipv4.ping_group_range
    value: "0 65536"
  ```
which doesn't even require NET_RAW.

See here for details on sysctls in podSecurityContext: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

### What's your helm version?

version.BuildInfo{Version:"v3.9.2", GitCommit:"1addefbfe665c350f4daf868a9adc5600cc064fd", GitTreeState:"clean", GoVersion:"go1.18.4"}

### What's your kubectl version?

Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:21:56Z", GoVersion:"go1.18.4", Compiler:"gc", Platform:"darwin/amd64"} Kustomize Version: v4.5.4 Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.8+k3s2", GitCommit:"fe3cecc219175ea85d7a95ed9e44349d94734bc7", GitTreeState:"clean", BuildDate:"2022-07-06T20:40:43Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/arm64"}

### Which chart?

prometheus-blackbox-exporter

### What's the chart version?

7.0.0

### What happened?

ICMP checks fails despite following the migration guide.

```
Logs for the probe:
ts=2022-08-10T21:08:25.154190692Z caller=main.go:352 module=icmp target=s3.us-west-1.amazonaws.com level=info msg="Beginning probe" probe=icmp timeout_seconds=9.5
[...]

PR #3616, opened 2023-07-19
[prometheus-blackbox-exporter] ICMP: Suggest setting `net.ipv4.ping_group_range` in `podSecurityContext`

fatz:
#### What this PR does / why we need it
To enable ICMP probes `net.ipv4.ping_group_range` must be set in some environments.

This PR adds a hint to the release notes aside of the hint to set `securityContext.capabilities.add: ["NET_RAW"]` for enabling ICMP.

This will IMO make it easier for users to get ICMP probes working

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [ ] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [ ] Chart Version bumped
- [ ] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #3925, opened 2023-10-20
[prometheus-blackbox-exporter] ICMP Pod Security Policy Update

sachasmart:
The current way with adding `NET_RAW` capability is not sufficient for the ICMP module to working correctly. Interestingly, the solution doesn't even require adding `NET_RAW` as a value, yet I kept it as is incase there is intention elsewhere.

Looks like other people had similar issues:

- fixes:
    - #2483
    - #2360

#### Checklist

- [x] Chart Version bumped


andrejshapal:
Hello, 
It would be more clean to add in podSpec:
https://github.com/prometheus-community/helm-charts/blob/16219b723e20e41cdfdba3944434acea3c7033d2/charts/prometheus-blackbox-exporter/templates/_helpers.tpl#L108

monotek (commented):
not sure if a comment in the readme (like in https://github.com/prometheus-community/helm-charts/pull/3616) is preferable to adding other option automatically which are not reflected in the values?

charts/prometheus-blackbox-exporter/templates/deployment.yaml
```diff
@@ -35,4 +35,13 @@ spec:
       {{- end }}
     spec:
     {{- include "prometheus-blackbox-exporter.podSpec" . | nindent 6 }}
+    {{- with .Values.podSecurityContext }}
```
sfudeus, line 38: If I'm not mistaken, you are doing it twice now. The line above includes the helper, so you set securityContext twice on podLevel.

charts/prometheus-blackbox-exporter/templates/_helpers.tpl
```diff
@@ -156,6 +156,11 @@ priorityClassName: "{{ . }}"
 {{- with .Values.podSecurityContext }}
 securityContext:
 {{ toYaml . | indent 2 }}
+{{- if has "NET_RAW" .Values.securityContext.capabilities.add }}
```
sfudeus, line 159: I'd suggest to not make this dependent on NET_RAW, because in the end these are 2 different ways to achieve the same goal (of NET_RAW would work out of the box). 
With the ping_group_range, you don't don't need NET_RAW anymore.

PR #3934, opened 2023-10-24
Populating dynamic honorLabels and honorTimestamps in kubelet ServiceMonitor

andrejshapal:
Hello, 

I have faced an issue, when due to circumstances, cadvisor metric timestamp is from the past. Due to this, prometheus rejects metrics with reason "Error on ingesting out-of-order samples". 
I don't see any advantage in using exporters timestamp over prometheus timestamp, therefore I want to try to reject honor labels. 
Unfortunately, I found it is impossible within kube-prometheus-stack helm chart. 
This PR gives an opportunity to set honorLabels/Timestamps without creating separate ServiceMonitor.

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 52.0.1
+version: 52.10.1
 appVersion: v0.68.0
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


andrejshapal:
@GMartinez-Sisti done

andrejshapal:
I am aware of this feature, but I am not sure I wan't to apply this settings for all targets.

andrejshapal:
@GMartinez-Sisti Actually, I set 10 for purpose as previous bumps for +10. But If you say it should be +1, no problem.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 52.0.1
+version: 52.1.1
```
GMartinez-Sisti, line 24: Sorry for the confusion, I just suggested the minor change on the previous, but this should be what we want, since we're adding a feature:
```suggestion
version: 52.1.0
```

andrejshapal:
@GMartinez-Sisti Sure. Thank you.

GMartinez-Sisti (approved):
Thanks @andrejshapal !

PR #3998, opened 2023-11-12
[prometheus] automountServiceAccountToken to be configurable for Service Account, Deployment, and Statefulset all at once.

ianhlee:
1. While rendering, the existing entry `.Values.serviceAccounts.automountServiceAccountToken` is checked if it is `true` or `false`. Since this flag being `true` is a default behavior from Kubernetes, users will usually try to turn it off by setting this flag as `false`, which will end up such change not being reflected into a Helm release at all.
   1. This PR uses a Helm template function `kindIs "bool"`, instead of just checking if its `true` or `false`, so that it no longer ignores the value set as `false`.
2. The existing entry `.Values.serviceAccounts.automountServiceAccountToken` is only for `ServiceAccount`, so it misses `Deployment` and `StatefulSet`.
   1. This PR adds a new entry, `.Values.server.automountServiceAccountToken`, opting out of automounting Kubernetes API credentials for `Deployment`, `StatefulSet`, and `ServiceAccount` using this one flag. Referred to the other charts in this repository for this pattern.
   2. It overrides the existing entry `.Values.serviceAccounts.automountServiceAccountToken` if set.

- fixes #

#### Checklist

- [x] Chart Version bumped


ianhlee:
ready for review.

zanhsieh (approved):

zeritti (changes requested):
Thank you, @ianhlee, for your PR and spotting the defficiency in the service account.

The proposed fix seems unnecessarily complicated, though, considering that what is needed is
- remove the conditional
- set default in values to true

If the property had been hard set in the service account, an option to override it in workloads would then have made sense. Now, as it already can be set in the service account, I'd keep it in this one place. And just one field in the values, please. I also suggest not to comment fields out.

ianhlee:
Hi @zeritti,

Thank you for the review.

Your comment makes sense to me. Can we clarify two things prior to proceeding?

1. I wonder if there's a benefit to keep rendered manifests simple, by leaving `automountServiceAccountToken` not set at all unless specified, rather than the chart's default behavior to always set it as `true`. Today, it's vanilla Kubernetes' default behavior, so it may not confuse anyone. Howwever, keeping the chart to avoid setting it at all unless specified, seems to be a way for the chart and the community to worry or keep managing one less thing.
2. Since the chart provides an option to avoid creating a new `ServiceAccount` just for the given Helm release, I believe it still should provide an option to manage this `automountServiceAccountToken` from `Deployment` or `StatefulSet` manifests as well. I looked up the other charts in this repository and ended up bringing this pattern into this chart, too.

How does it sound to you?

Pleaes advise. I look forward to hearing back from you and to merge something that hopefully benefits open source communities for the first time in my life 🤩

zeritti (approved):
> I wonder if there's a benefit to keep rendered manifests simple, by leaving automountServiceAccountToken not set at all unless specified, rather than the chart's default behavior to always set it as true. Today, it's vanilla Kubernetes' default behavior, so it may not confuse anyone. However, keeping the chart to avoid setting it at all unless specified, seems to be a way for the chart and the community to worry or keep managing one less thing.

The difficulty that I see with fields commented out is that 1) they are not well visible to users (we are not documenting the values), 2) they are not well visible to yaml tools and 3) they are ignored during validation/linting/CI unless each is specifically set during the tests (something which we should amend with this chart), potentially hiding errors.

> Since the chart provides an option to avoid creating a new ServiceAccount just for the given Helm release, I believe it still should provide an option to manage this automountServiceAccountToken from Deployment or StatefulSet manifests as well. I looked up the other charts in this repository and ended up bringing this pattern into this chart, too.

Most charts in the repo that set the field do so in the workload template as well as in the service account template, indeed, and that via a single values' field.

PR #4014, opened 2023-11-15
[jiralert] add envFrom config

jon-rei:

This PR adds the envFrom config for jiralert pods. We need it to include a secret created outside the helm chart as an environment variable, as we want to use environment variables in the jiralert config.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jiralert/Chart.yaml
+++ b/charts/jiralert/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: jiralert
 description: A Helm chart for Kubernetes to install jiralert
 type: application
-version: 1.6.1
+version: 1.7.0
 appVersion: "v1.3.0"
 home: "https://github.com/prometheus-community/jiralert"
 keywords:
--- a/charts/jiralert/templates/deployment.yaml
+++ b/charts/jiralert/templates/deployment.yaml
@@ -72,6 +72,10 @@ spec:
               value: {{ $value | quote }}
           {{- end }}
         {{- end }}
+        {{- with .Values.envFrom }}
+          envFrom:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
       serviceAccountName: {{ include "jiralert.serviceAccountName" . }}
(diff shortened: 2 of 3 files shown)


jkroepke (changes requested):
Could you please a example config in https://github.com/prometheus-community/helm-charts/blob/main/charts/jiralert/ci/test-values.yaml ?

This ensures, that the feature is covered by CI.

jon-rei:
@jkroepke I've added a example config but it seems like that the Lint and Test check failed because of this:
```
Warning  Failed     7m51s (x12 over 9m59s)  kubelet            Error: secret "jiralert-token" not found
```
Is it possible to create a secret in the test pipeline?

jkroepke:
@jon-rei Do you want to implement something like this? https://github.com/prometheus-community/helm-charts/pull/3593/files

jon-rei:
I may have misunderstood then adding and example config to the `test-values.yaml` file. I thought I could just add the secret name there and it would work. But unfortunately it cannot even find the secret created by the helm chart.

jkroepke:
The name of the secret based on Release.Name and chart-test is using a random Release Name by default. It's not possible to reference the own secrets, since it is not predictive.

The approch of extraManifests (https://github.com/prometheus-community/helm-charts/pull/3593/files) is the only way to add additional Manifests into the test context.

If you feel, this is over the top, please let me now.

jon-rei:
Understood, in the end my goal for adding the `envFrom` was that we use SealedSecrets for our secret management and I want to use the secret created by SealedSecret outside of the Helm Chart in the Jiralert pod.
I'm not sure if the extraManifest would be a use case for Jiralert. So I think this might be a bit over the top.

jkroepke:
@jon-rei okay, then in this case, please revert the change in test and I will approve your request

jon-rei:
@jkroepke done.

jkroepke (approved):
LGTM 👍

PR #4026, opened 2023-11-20
[prometheus-adapter] Only set priorityClassName if non-empty

iangudger:
Argo CD rejects the deployment when the priorityClassName key is present without a value.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/templates/deployment.yaml
+++ b/charts/prometheus-adapter/templates/deployment.yaml
@@ -117,7 +117,9 @@ spec:
         {{- toYaml .Values.affinity | nindent 8 }}
       topologySpreadConstraints:
         {{- toYaml .Values.topologySpreadConstraints | nindent 8 }}
-      priorityClassName: {{ .Values.priorityClassName }}
+      {{- with .Values.priorityClassName }}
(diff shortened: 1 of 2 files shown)


charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -117,7 +117,9 @@ spec:
         {{- toYaml .Values.affinity | nindent 8 }}
       topologySpreadConstraints:
         {{- toYaml .Values.topologySpreadConstraints | nindent 8 }}
+      {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
```
steven-sheehy, line 121: Don't repeat yourself
```suggestion
      {{- with .Values.priorityClassName }}
      priorityClassName: {{ . }}
```

steven-sheehy (changes requested):

charts/prometheus-adapter/templates/deployment.yaml
iangudger: Done.

steven-sheehy (approved):

PR #4042, opened 2023-11-27
[kube-prometheus-stack]use metrics from this charts kubelet-service

YuleZ:
This PR adds the kubelet's service name from this helm-chart to prometheus-rules in order to avoid execution errors when using k8s-services that target kubelet as well.

In my case the elastic-search operator creates a similar service for Kubelet which errors in combination with kubelet-prometheus-rules from this helm-chart.

```
execution: 
found duplicate series for the match group 
{instance="x.x.x.x:10250"} on the right hand-side of the operation: 
[
{__name__="kubelet_node_name", endpoint="https-metrics", instance="x.x.x.x:10250", job="kubelet", metrics_path="/metrics", namespace="kube-system", node="node1", service="elastic-search-kubelet"},
{__name__="kubelet_node_name", endpoint="https-metrics", instance="x.x.x.x:10250", job="kubelet", metrics_path="/metrics", namespace="kube-system", node="node1", service="kube-prometheus-stack-kubelet"}
]
 ;many-to-many matching not allowed: matching labels must be unique on one side
 ```

See above

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -21,7 +21,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 54.2.2
+version: 54.2.3
 appVersion: v0.69.1
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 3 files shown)


YuleZ:
Review please
@andrewgkew @gianrubio @gkarthiks @GMartinez-Sisti @QuentinBisson @scottrigby @Xtigyro

jkroepke:
@YuleZ Modify the recording rules is the wrong approach here because the metrics are still collected twice. 

The service monitor for the kubelet server should configured more restrictive to avoid this.

Issue #3621, opened 2023-07-20
[promethus-snmp-exporter] Unable to use configmap to load/reload configuration


tuctboh:
### Describe the bug a clear and concise description of what the bug is.

I installed the chart using -

helm install prometheus-snmp-exporter --set config=" " prometheus-community/prometheus-snmp-exporter

I realize thats not proper, but I had tried everything else I could think of to get the configmap option to kick in. It kept failing so I did that and it worked. GREAT!

So then I started to try to edit the config map and then ran into problems. I first tried with the stock supplied snmp.yml, and no luck, I was told the file was too big. So I cut the file down to just ONE auth (V2) and just ONE module (if_mib). That worked, but then it never showed up as usable in Prometheus. I took a look at the log, and it ends with -

ts=2023-07-20T16:30:41.186Z caller=main.go:186 level=error msg="Error reloading config" err="yaml: unmarshal errors:\n  line 3: field TBOH_v2 not found in type config.plain\n  line 10: field if_mib not found in type config.plain"

TBOH_v2 was my auth config, if_mib was my module config. 

I logged on with a shell and took a look at the snmp.yml its using, and it looks right, starts off

```
# WARNING: This file was auto-generated using snmp_exporter generator, manual changes will be lost.
auths:
  TBOH_v2:
    community: NunyaBizne$$
    security_level: noAuthNoPriv
    auth_protocol: MD5
    priv_protocol: DES
    version: 2
modules:
  if_mib:
    walk:
```
which to the best of my extremely limited knowledge is fine.

So no snmp for me right now

### What's your helm version?

version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.3"}

### What's your kubectl version?

[...]

PR #4043, opened 2023-11-28
[prometheus-snmp-exporter] Add extraVolumes support and passing auth param

tr3mor:

This PR upgrades snmp exporter to version 0.24.1 which supports submitting config as multiple files (see https://github.com/prometheus/snmp_exporter/issues/628). This provides the ability to move the auth part into a separate file and load it separately as a secret (from Vault for example).

This PR also introduces support for extraVolumes and extraVolumeMounts to support this feature.

My use case for this feature:
I have an init container that is responsible for getting the auth part of the config from the vault and needs to pass it to snmp-exporter, for this, I need to create an emptyDir volume and mount it to both containers. 
Also, it is quite a common feature to have for helm-charts with a lot of use cases in general.

Finally, this PR adds ability to specify auth query param to be passed to exporter via service monitor configuration (see https://github.com/prometheus/snmp_exporter/issues/619)

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 2.1.0
+version: 3.0.0
 appVersion: v0.21.0
 home: https://github.com/prometheus/snmp_exporter
 sources:
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
 version: 3.0.0
-appVersion: v0.21.0
+appVersion: v0.24.1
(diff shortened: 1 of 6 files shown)


walker-tom (changes requested):
Thanks for the PR!

All the changes look good to me.

Please could you update the Chart version to be `3.0.0` because the new snmp_exporter used includes a breaking change from [snmp_exporter:v0.23.0](https://github.com/prometheus/snmp_exporter/releases/tag/v0.23.0) therefore users values may need to change to continue working.

Could you then please add a new section of the Chart README's [Upgrading Chart section](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-snmp-exporter#upgrading-chart) underneath the existing `To 1.0.0` and `To 2.0.0` sections which mentions that this breaking change and links to the [snmp_exporter Module and Auth Split Migration docs](https://github.com/prometheus/snmp_exporter/blob/main/auth-split-migration.md) so users know what changes need to be made.

Once these two changes are done I will get this approved and released, thanks!

tr3mor:
Hey @walker-tom , thank you for taking a maintainer role on this one!
I have introduced the requested changes, please let me know if the wording is not good. Thank you!

walker-tom (approved):
Changes look good, thanks! This closes #4066 and #3621.

PR #4089, opened 2023-12-14
[prometheus-blackbox-exporter] Support setting extraInitContainers and extraContainers to a string value to enable full templating

JaredTan95:
When we add extraContainers, we usually want to overwrite its contents with `--set`. However, due to the characteristics of `helm`, we need all the contents of the array to be set, as follows:

```yaml
extraContainers: 
  - name: oAuth2-proxy
    args:
      - -https-address=:9116
      - -upstream=http://localhost:9115
    image: openshift/oauth-proxy:v1.1.0
```

if we want to override image by set redundant parameters as follows:
```bash
--set extraContainers[0].image=my-resgistry/openshift/oauth-proxy:v1.1.0 # 👈That's what I really want to override
--set extraContainers[0].name=oAuth2-proxy \
--set extraContainers[0].args[0]=-https-address=:9116 \
--set extraContainers[0].args[1]=-upstream=http://localhost:9115 \ 
```

Just as other [communities](https://github.com/fluent/helm-charts/blob/b0f13f51067d4a32aeef461371db87c63a85909d/charts/fluent-bit/templates/_pod.tpl#L27) have done. With this PR, we can define the template in values:

```yaml
global:
 registry: my-registry
extraContainers:  |- # 👈 Thats the diffrence
  - name: oAuth2-proxy
    args:
      - -https-address=:9116
      - -upstream=http://localhost:9115
    image: {{ .Values.global.registry }}/openshift/oauth-proxy:v1.1.0 # 👈 Thats the diffrence
```
we can only set the registry：

```bash
helm install --set global.registry=another-registry
```

and the redered yaml would be like:

```yaml
extraContainers: 
  - name: oAuth2-proxy
    args:
      - -https-address=:9116
      - -upstream=http://localhost:9115
    image: another-registry/openshift/oauth-proxy:v1.1.0
```

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 8.6.2
+version: 8.7.0
 appVersion: v0.24.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
--- a/charts/prometheus-blackbox-exporter/Chart.yaml
+++ b/charts/prometheus-blackbox-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 8.6.1
+version: 8.6.2
 appVersion: v0.24.0
 home: https://github.com/prometheus/blackbox_exporter
 sources:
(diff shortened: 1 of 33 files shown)


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 8.6.1
+version: 8.6.2
```
monotek, line 4: this should be at least be a minor version change as a new feature is introduced.
```suggestion
version: 8.7.0
```

as it seems the old way to configure it is still supported?
otherwise major version upgrade would be needed.

monotek (changes requested):

monotek (approved):

PR #4096, opened 2023-12-18
[kube-state-metrics] Default to Recreate

### LINKED ISSUE #4095: [kube-state-metrics] Strategy RollingUpdate causes 'duplicate' metrics

Metrics from KSM have labels like `instance` and `pod` that are filled by the KSM pod in question. Which means that when one uses RollingUpdate, it happens that you get duplicate metrics. For example (ommited some labels for clarity:

```
kube_deployment_status_replicas_available{instance="x:8080"} 1
kube_deployment_status_replicas_available{instance="y:8080"} 1
```

and thus any query that doesn't utilises `max by (pod and/or instance)` will provide a 'bad' result. 

I personally think the default should be `recreate`. - Yes, it _will_ provide a gap in data during the time you make a new release, however the trade off is IMHO better to have a gap which doesn't really hurt, than having actual visual 'bad' data. 

The screenshot here shows the difference

image

The spike is a restart with RollingUpdate, and the small gap is a restart with Recreate

Make `recreate` rollout strategy the default and give guidance on how to prevent it otherwise.

NONE

### Additional context.

_No response_

wiardvanrij:

See issue/ticket.

- fixes #4095

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -20,7 +20,7 @@ spec:
   replicas: {{ .Values.replicas }}
   {{- if not .Values.autosharding.enabled }}
   strategy:
-    type: {{ .Values.updateStrategy | default "Recreate" }}
+    type: {{ .Values.updateStrategy | default "RollingUpdate" }}
   {{- end }}
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
   {{- if .Values.autosharding.enabled }}
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -20,7 +20,7 @@ spec:
   replicas: {{ .Values.replicas }}
   {{- if not .Values.autosharding.enabled }}
   strategy:
-    type: {{ .Values.updateStrategy | default "RollingUpdate" }}
+    type: {{ .Values.updateStrategy | default "Recreate" }}
   {{- end }}
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
   {{- if .Values.autosharding.enabled }}
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -37,8 +37,11 @@ autosharding:
 
 replicas: 1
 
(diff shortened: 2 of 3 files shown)


charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
   replicas: {{ .Values.replicas }}
   {{- if not .Values.autosharding.enabled }}
   strategy:
-    type: {{ .Values.updateStrategy | default "RollingUpdate" }}
+    type: {{ .Values.updateStrategy | default "Recreate" }}
```
dotdc, line 23: Please revert to `RollingUpdate`

charts/kube-state-metrics/values.yaml
```diff
@@ -37,8 +37,11 @@ autosharding:
 
 replicas: 1
 
-# Change the deployment strategy when autosharding is disabled
-# updateStrategy: Recreate
+# Change the deployment strategy when autosharding is disabled.
+# The default is "Recreate" in favor of preventing two instances from running, but this might create a small gap in data.
+# When using "RollingUpdate", it's wise to use a pessimistic approach when querying for "kube-state-metrics" and perform something like this:
+# max by(namespace, pod) (kube_pod_container_status_restarts).
+# updateStrategy: RollingUpdate
```
dotdc, line 44: I would rework this paragraph taking into account that `RollingUpdate` is the default.
You can also add : 
```yaml
# ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
```

dotdc (changes requested):
Like explained in https://github.com/prometheus-community/helm-charts/issues/4095, I would keep `RollingUpdate` as the default strategy.

mrueg:
I would suggest to keep RollingUpdate as the default as well, rather than Recreating. 
If you run ksm with more than 1 replica, you have to deal with the instance in your query as well.
If recreate fails for some reason and a rollback is needed, you will observe larger gaps in your metrics.

dotdc (changes requested):
@wiardvanrij Could you bump the chart to 5.15.4 so we can merge?

dotdc:
Done in #4318

PR #3066, opened 2023-02-23
[kube-prometheus-stack] Introduce failurePolicy "IgnoreOnInstallOnly"

jkroepke:
#### What this PR does / why we need it

Introduce a special failurePolicy "IgnoreOnInstallOnly" which sets the failurePolicy on Ignore on first installation and set it to Fail on the next upgrade run.

```
(venv) jok@CDT-MB-20200178 kube-prometheus-stack % helm template demo . --set prometheusOperator.admissionWebhooks.failurePolicy=IgnoreOnInstallOnly | grep failurePolicy
    failurePolicy: Ignore
    failurePolicy: Ignore
(venv) jok@CDT-MB-20200178 kube-prometheus-stack % helm template demo . --set prometheusOperator.admissionWebhooks.failurePolicy=IgnoreOnInstallOnly --is-upgrade | grep failurePolicy
    failurePolicy: Fail
    failurePolicy: Fail

```

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #4102, opened 2023-12-20
[kube-prometheus-stack] fix failure-policy on patch-webhook job

jonasbadstuebner:
@GMartinez-Sisti

This PR fixes the following error message on the admission-patch:
```
kube-prometheus-stack-admission-patch-nsmmw {"level":"fatal","msg":"patch-failure-policy IgnoreOnInstallOnly is not valid","source":"cmd/patch.go:106","time":"2023-12-20T09:52:44Z"}
```
by not directly using the value from the HelmRelease-Values.
This error occurs, if you set `.Values.prometheusOperator.admissionWebhooks.failurePolicy` to `IgnoreOnInstallOnly` in the current version of the Chart.

It also adds a new helper function for determining the correct `failurePolicy` based on the values.

#### Checklist

- [x] Chart Version bumped


jkroepke:
Pleas bump the chart version, thanks!

Xtigyro:
@DrBu7cher Please resolve the conflicts.

charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-patchWebhook.yaml
```diff
@@ -46,7 +46,7 @@ spec:
             - --webhook-name={{ template "kube-prometheus-stack.fullname" . }}-admission
             - --namespace={{ template "kube-prometheus-stack.namespace" . }}
             - --secret-name={{ template "kube-prometheus-stack.fullname" . }}-admission
-            - --patch-failure-policy={{ .Values.prometheusOperator.admissionWebhooks.failurePolicy }}
```
jkroepke, line 49: I ask myself, if it make sense to have `Fail` here if `.Values.prometheusOperator.admissionWebhooks.failurePolicy` is `IgnoreOnInstallOnly`

Because after the initial install, the policy can be set to Failure and the helm hook always runs after a complete installation.

What did you thing? @DrBu7cher

DrBu7cher: What would be your suggestion?

DrBu7cher: The whole idea behind this PR is that the value for this policy should be „Ignore“ when installing and „Fail“ when not installing. Previously, the hook would have been set up with the flag set to a value that is invalid, as the error message in my opening comment showed.

What value or behavior would make more sense to you?

jkroepke: > The whole idea behind this PR is that the value for this policy should be „Ignore“ when installing and „Fail“ when not installing.

I'm aware of this, since I [implement](https://github.com/prometheus-community/helm-charts/pull/3066) the idea of IgnoreOnInstallOnly, but I'm using cert-manager solution and I was not aware that the patch job also patch the policy.

After review the PR, I thought since the patch job runs after installation as additional hook, it cloud already set the failure-policy to `Failure`, if `.Values.prometheusOperator.admissionWebhooks.failurePolicy` is set to `IgnoreOnInstallOnly`. It would omit the required of a second `helm upgrade` after the initial deployment.

The current state in the PR is also fine for me. Its more like an optimization, if it make sense.

jkroepke: Otherwise, if the value here is always equal to the value inside webhooks, I does it make sense to have this parameter? If omit, the patch pod would not touch failurePolicy. That also seems fine to me, since we maintain the value directly via helm.

PR #2053, opened 2022-05-13
[prometheus-node-exporter] Make liveness and readiness probes customi…

tdrachenic:
#### What this PR does / why we need it:

I made liveness and readiness probes customizable to make node-exporter features like TLS and Basic-Auth fully available for chart users. When you enable those features in node-exporter configuration then probes crash. To use TLS, we need to change httpGet.scheme to HTTPS and for Basic-Auth we need to add httpHeaders in httpGet.

#### Checklist
- [X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #4114, opened 2023-12-27
[kube-state-metrics] Make liveness and readiness probes customizable

MikeKlebolt:
I made liveness and readiness probes customizable to make kube-state-metrics features like TLS and Basic-Auth fully available for chart users without requiring kube-rbac-proxy. When you enable those features in kube-state-metrics configuration then probes crash. To use TLS, we need to change httpGet.scheme to HTTPS and for Basic-Auth we need to add httpHeaders in httpGet.

I've also renamed `httpPort` to `servicePort` to bring it in-line with node-exporter.

This code was borrowed from #2053 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -49,10 +49,10 @@ spec:
       {{- toYaml . | nindent 6 }}
       {{- end }}
       containers:
-      {{- $httpPort := ternary 9090 (.Values.service.port | default 8080) .Values.kubeRBACProxy.enabled}}
+      {{- $servicePort := ternary 9090 (.Values.service.port | default 8080) .Values.kubeRBACProxy.enabled}}
       {{- $telemetryPort := ternary 9091 (.Values.selfMonitor.telemetryPort | default 8081) .Values.kubeRBACProxy.enabled}}
       - name: {{ template "kube-state-metrics.name" . }}
-        {{- if .Values.autosharding.enabled }}
+        {{- if  .Values.autosharding.enabled }}
         env:
         - name: POD_NAME
           valueFrom:
@@ -67,7 +67,7 @@ spec:
         {{-  if .Values.extraArgs  }}
         {{- .Values.extraArgs | toYaml | nindent 8 }}
         {{-  end  }}
-        - --port={{ $httpPort }}
+        - --port={{ $servicePort }}
         {{-  if .Values.collectors  }}
         - --resources={{ .Values.collectors | join "," }}
         {{-  end  }}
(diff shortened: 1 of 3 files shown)


MikeKlebolt:
@dotdc, @mrueg, @tariq1890.  Please let me know if there are any issues with this PR so we can get it merged.  Thanks!

MikeKlebolt:
@dotdc, @mrueg, @tariq1890, can we get the pipeline re-ran?  I've addressed the linting error.

eanveden:
@dotdc @mrueg @tariq1890

Our organization have a requirement to enable e2e TLS in our clusters, we need this PR for that to be possible. Any chance you could take a look?

Thank you!

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -147,17 +147,41 @@ spec:
         {{- end }}
         {{- end }}
         livenessProbe:
+          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
           httpGet:
+            {{- if .Values.kubeRBACProxy.enabled }}
```
dotdc, line 152: Maybe this should be configurable too, with a comment for RBAC proxy users in the values file?

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -147,17 +147,41 @@ spec:
...
+            port: {{ $servicePort }}
+            scheme: {{ upper .Values.livenessProbe.httpGet.scheme }}
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
+          successThreshold: {{ .Values.livenessProbe.successThreshold }}
+          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
         readinessProbe:
+          failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
           httpGet:
+            {{- if .Values.kubeRBACProxy.enabled }}
```
dotdc, line 170: Maybe this should be configurable too, with a comment for RBAC proxy users in the values file?

dotdc (commented):
Didn't test, but looks good for the most part.
Is there a reason why `host` is not configurable like the other parameters?

Also, could you update the branch?

MikeKlebolt:
> Didn't test, but looks good for the most part. Is there a reason why `host` is not configurable like the other parameters?
> 
> Also, could you update the branch?

I just grabbed the code that was available in #2053.  But upon further investigation this appears to not be necessary at all and I'm not sure why it got added later in #3336.  As a result I have removed `host`.

MikeKlebolt:
> Didn't test, but looks good for the most part. Is there a reason why `host` is not configurable like the other parameters?
> 
> Also, could you update the branch?

It shouldn't need to be configurable because 127.0.0.1 is hardcoded into the chart for kube-rbac-proxy elsewhere.  IMO that kinda falls outside the scope of this PR.

dotdc (approved):

dotdc:
Thank you for your contribution @MikeKlebolt !

PR #4124, opened 2024-01-04
[kube-prometheus-stack] Add value json schema file

dongjiang1989:

Adds json-schema for kube-prometheus-stack Chart values.
ref: https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1417,8 +1417,8 @@ kubeControllerManager:
     ## If null or unset, the value is determined dynamically based on target Kubernetes version due to change
     ## of default port in Kubernetes 1.22.
     ##
-    port: null
-    targetPort: null
+    port: ""
+    targetPort: ""
     # selector:
     #   component: kube-controller-manager
 
@@ -1456,13 +1456,13 @@ kubeControllerManager:
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks.
     ## If null or unset, the value is determined dynamically based on target Kubernetes version.
     ##
-    https: null
+    https: false
 
     # Skip TLS certificate validation when scraping
-    insecureSkipVerify: null
+    insecureSkipVerify: false
 
     # Name of the server to use when validating TLS certificate
-    serverName: null
+    serverName: ""
 
     ## MetricRelabelConfigs to apply to samples after scraping, but before ingestion.
     ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#relabelconfig
@@ -1748,8 +1748,8 @@ kubeScheduler:
     ## If null or unset, the value is determined dynamically based on target Kubernetes version due to change
     ## of default port in Kubernetes 1.23.
     ##
-    port: null
-    targetPort: null
+    port: ""
+    targetPort: "" 
     # selector:
     #   component: kube-scheduler
 
@@ -1786,13 +1786,13 @@ kubeScheduler:
     ## Requires proper certs (not self-signed) and delegated authentication/authorization checks.
     ## If null or unset, the value is determined dynamically based on target Kubernetes version.
     ##
-    https: null
+    https: false
 
     ## Skip TLS certificate validation when scraping
-    insecureSkipVerify: null
+    insecureSkipVerify: false
 
     ## Name of the server to use when validating TLS certificate
-    serverName: null
+    serverName: ""
 
     ## MetricRelabelConfigs to apply to samples after scraping, but before ingestion.
     ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#relabelconfig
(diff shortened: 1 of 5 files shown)


GMartinez-Sisti (changes requested):
Hi @dongjiang1989, thank you for the PR! While this change adds value to the chart, it can be highly disruptive. To accept this we need:

- Documentation on the feature
- Script/steps to update it if keys are changed

I've seen other repos where this was added and it caused issues for weeks and on all open PRs. Let me know your thoughts on this.

dongjiang1989:
> Hi @dongjiang1989, thank you for the PR! While this change adds value to the chart, it can be highly disruptive. To accept this we need:
> 
> * Documentation on the feature
> * Script/steps to update it if keys are changed
> 
> I've seen other repos where this was added and it caused issues for weeks and on all open PRs. Let me know your thoughts on this.

Hi @GMartinez-Sisti, thanks for your review.  It is very necessary to add verification to `value.schema.json` file. 

First, I added the `value.schema.json` file check in `Lint and Test Charts` (just for charts that contains this file); 
Then I added instructions to the README.md file. When updating `value.yaml`, I need to update the `value.schema.json` file synchronously.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1748,8 +1748,8 @@ kubeScheduler:
     ## If null or unset, the value is determined dynamically based on target Kubernetes version due to change
     ## of default port in Kubernetes 1.23.
     ##
-    port: null
-    targetPort: null
+    # port: null
```
jkroepke, line 1751: By comment such keys out, they also omit from values.schema.json. They should be included into schema and the type should anyOf `null|number`.

jkroepke (changes requested):
I'm not a huge fan of generated values.schema files. We are unable to check 4000 lines of generated values.schema.json. It's unknown to me, what can be break here.

A good example are port numbers. Currently its fine to quote ports and mark them as string which is not longer allowed. Forcing them to integer has not a big effect, since Kubernetes will denied non numeric numbers anyway, expect on `targetPort`.

What is definitive incorrect: `targetPort` is marked as integer in the spec, but Kubernetes also allows port names (string values) on `Service.spec.ports.targetPort`. Such valid configuration would be now denied by schema now.

Additional, comment out default values decreases the user experience for JetBrains users, since the Jetbrains IDE can build a keys-only schema based on the existing values file.

I'm big fan of values.schema.json, but not of auto-generated. I'm open to handcrafted schema that develop step-by-step.

GMartinez-Sisti:
@jkroepke I agree with most of your points, but I don't think handcrafted schema is manageable for an OSS helm chart like this. It needs to something that "just works™️ ", otherwise we're going to have to hand held everyone that wants to update the values file and is not aware of this.

jkroepke:
> It needs to something that "just works™️ "

Then, I'm against an values.schema.json for kube-prometheus-stack, because it just works without.

dongjiang1989:
> > It needs to something that "just works™️ "
> 
> Then, I'm against an values.schema.json for kube-prometheus-stack, because it just works without.
> 
> Alternatively, create a schema which always allows any of null, number, string type, if the value is not complex. This should be less disruptive. Or omit the type from schema that we have something like any key only schema.
> 
> Generate a schema based on the default values is a wrong way and would introduce a lot of issues.

@jkroepke  Do you have any other ways to better manage json-schema files?

jkroepke:
The you [tool](https://github.com/losisin/helm-values-schema-json) provide here is already good. It also supports [additional hints](https://github.com/losisin/helm-values-schema-json/blob/main/testdata/full.yaml) for schema generation. However, like https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/ says. Generate a schema from a existing values file is a good point to **start**. It need to be revalidate, which values may support multiple types and which keys are optional or required. Such an schema need constant maintenance. 

it would be an additional step for each PR author to maintain the comments. In worst case, maintainers have to educate such things. Like @GMartinez-Sisti says: 

> otherwise we're going to have to hand held everyone that wants to update the values file and is not aware of this.

At the moment, I think the maintenance costs are higher then a benefit. At the moment, I did not see any benefit.

If read https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/ again, it advertise features like `pattern` or `required` but both properties are not included into the generated file.

jkroepke:
@GMartinez-Sisti if you agree, I would recommend to close/decline this.

GMartinez-Sisti:
> @GMartinez-Sisti if you agree, I would recommend to close/decline this.

> At the moment, I think the maintenance costs are higher then a benefit. At the moment, I did not see any benefit.

I'm afraid this is true. Adding this would be a nice to have since it's not fixing any current problem, but it would cause extra steps to add new fields to the chart.

Also something we forgot, is that we use a lot of dependent charts, and we would need to check how to manage them, https://github.com/helm/helm/issues/10392 is a good example of this issue.

@dongjiang1989 thank you for the interest and help 🙏 but we will have to skip this one. Can't wait for the next contribution from you 😃 !

PR #4190, opened 2024-01-26
[prometheus-node-exporter] hostNetwork is now default to false

### LINKED ISSUE #4129: [prometheus-node-exporter] HostNetwork defaults to true

In the node-exporter Helm Chart the "hostNetwork" value is configurable but defaults to true.

### What's your helm version?

3.13.2

v1.15.5

### Which chart?

prometheus-node-exporter

### What's the chart version?

4.25.0

### What happened?

I have tracked multiple issues where people ask about why this is necessary ( #1986  , #3554 for example) and never got an answer.
If your Cluster is CIS compliant, chances are, you have a default firewall blocking all connections to hostnet exept the well known Kubernetes ports and so having hostNetwork to true will not work. Furthermore, it doesn't seem to be a good security practice.
Is there any reason hostNetwork does have to be true per Default?

hostNetwork should be false per default

### How to reproduce it?

_No response_

_No response_

helm install node-exporter-release prometheus-community/prometheus-node-exporter

_No response_

Sheikh-Abubaker:
If hostNetwork is default to true, it could lead to issues or conflicts and running a pod in the hostNetwork mode means it has access to the network interfaces of the host machine, this might not be secured, there could also be potential port conflicts, if the port that Node Exporter uses is already in use on the host, it could lead to security risks.

- fixes #4129 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,8 +6,8 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.26.0
-appVersion: 1.7.1
+version: 4.26.1
+appVersion: 1.7.0
 home: https://github.com/prometheus/node_exporter/
 sources:
   - https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - exporter
 type: application
 version: 4.26.0
-appVersion: 1.7.0
+appVersion: 1.7.1
 home: https://github.com/prometheus/node_exporter/
 sources:
   - https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -302,7 +302,7 @@ rbac:
 endpoints: []
 
 # Expose the service to the host network
-hostNetwork: true
+hostNetwork: false
 
 # Share the host process ID namespace
 hostPID: true
--- a/charts/prometheus-operator-admission-webhook/Chart.yaml
+++ b/charts/prometheus-operator-admission-webhook/Chart.yaml
@@ -2,8 +2,8 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.10.0
-appVersion: 0.71.0
+version: 0.10.1
+appVersion: 0.71.2
 home: https://github.com/prometheus-operator/prometheus-operator
 icon: https://github.com/prometheus-operator/prometheus-operator/raw/main/Documentation/logos/prometheus-operator-logo.png
 keywords:


zanhsieh (changes requested):
@Sheikh-Abubaker 

> If hostNetwork is default to true, it could lead to issues or conflicts and running a pod in the hostNetwork mode means it has access to the network interfaces of the host machine, this might not be secured, there could also be potential port conflicts, if the port that Node Exporter uses is already in use on the host, it could lead to security risks.

This is what `prometheus-node-exporter` intend to do: It exports the metrics on each node and wait prometheus to scrape. Did you ever wonder why only `chart/prometheus-node-exporter/templates/daemonset.yaml` exist? Why no `chart/prometheus-node-exporter/templates/deployment`?

Sheikh-Abubaker:
> This is what `prometheus-node-exporter` intend to do: It exports the metrics on each node and wait prometheus to scrape. Did you ever wonder why only `chart/prometheus-node-exporter/templates/daemonset.yaml` exist? Why no `chart/prometheus-node-exporter/templates/deployment`?

@zanhsieh `chart/prometheus-node-exporter/templates/daemonset.yaml` exist but not `chart/prometheus-node-exporter/templates/deployment` for prometheus-node-exporter as DaemonSets are primarily for system level tasks like monitoring, logging or other system level operations while Deployments are primarily for application level tasks like easy scaling, self healing, rollback etc, so I guess this clears why `hostNetwork` defaults to true as it needs to perform system level tasks it needs more visibility into the host's system metrics.

zanhsieh:
@Sheikh-Abubaker 
> chart/prometheus-node-exporter/templates/daemonset.yaml exist but not chart/prometheus-node-exporter/templates/deployment for prometheus-node-exporter as DaemonSets are primarily for system level tasks like monitoring, logging or other system level operations 
According to what you stated, why just not concentrate by using certain nodes to monitor that potentially saves more resources? And why is this chart called "prometheus-node-exporter"?

Sheikh-Abubaker:
> According to what you stated, why just not concentrate by using certain nodes to monitor that potentially saves more resources? And why is this chart called "prometheus-node-exporter"?

@zanhsieh 

This chart is called prometheus-node-exporter as it is a prometheus exporter which collects and exposes various machine level metrics such as CPU usage, memory usage, disk I/O and network statistics of host machines running node-exporter, and then prometheus collects those metrics from node-exporter allowing prometheus to monitor.

zanhsieh:
> @zanhsieh
> 
> This chart is called prometheus-node-exporter as it is a prometheus exporter which collects and exposes various machine level metrics such as CPU usage, memory usage, disk I/O and network statistics of host machines running node-exporter, and then prometheus collects those metrics from node-exporter allowing prometheus to monitor.

@Sheikh-Abubaker 
Actually I don't think this PR should be existed. The design targets of `prometheus-node-exporter` are:
- Export metrics, such as CPU, NIC, storage info, of each node to let prometheus to scrape it.
- In order to get those metrics, the exporter shall exist on each node and be able to access the raw metrics. In other words, this exporter shall be implemented as `DaemonSet`, and `have necessary permissions to access raw physical interfaces`.

So if you don't allow the DaemonSet to touch the physical hostnetwork, it won't be possible to expose the node network information, such as network in / network out / network packets in / network packets out. Even possible, the user might wonder why this chart called `prometheus-node-exporter` but doesn't match what metrics it exposed.

albanm:
Accessing the host network seems necessary indeed. But opening the port 9100 to the outside without explicit nodePort service or ingress rule is a pretty serious violation of trust in my opinion. For example this has been reported to me as a vulnerability by the qualys scanner.

Does anybody know if it is possible at all to have both (access host network for monitoring but expose port only internally) ?

PR #4191, opened 2024-01-26
[prometheus-node-exporter] Add Termination Msg Params to Daemonset

nfrazier08:

For our purposes of this chart, we would like to be able to customize the node-exporter daemonset with the following variables:

- For both the Node Exporter and KubeRBACProxy containers, variables have been added to configure `terminationMessagePath` and `terminationMessagePolicy`
- A variable was added to configure the `terminationGracePeriodSeconds`

- fixes #

#### Checklist

- [x] Chart Version bumped

@zeritti @gianrubio @zanhsieh

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.27.1
+version: 4.27.0
 appVersion: 1.7.0
 home: https://github.com/prometheus/node_exporter/
 sources:
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.26.1
+version: 4.27.1
 appVersion: 1.7.0
 home: https://github.com/prometheus/node_exporter/
 sources:
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -40,6 +40,8 @@ global:
 kubeRBACProxy:
   enabled: false
   env: {}
+  ##  env:
+  ##    VARIABLE: value
   image:
     registry: quay.io
     repository: brancz/kube-rbac-proxy
@@ -47,7 +49,7 @@ kubeRBACProxy:
     sha: ""
     pullPolicy: IfNotPresent
 
-  # List of additional cli arguments to configure kube-rbac-prxy
+  # List of additional cli arguments to configure kube-rbac-proxy
   # for example: --tls-cipher-suites, --log-file, etc.
   # all the possible args can be found here: https://github.com/brancz/kube-rbac-proxy#usage
   extraArgs: []
@@ -276,7 +278,8 @@ resources: {}
   #   cpu: 100m
   #   memory: 30Mi
 
-# CHANGED
+# Specify the container restart policy passed to the Node Export container
+# Possible Values: Always (default)|OnFailure|Never
 restartPolicy: Always
 
 serviceAccount:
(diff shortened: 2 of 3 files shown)


zanhsieh (changes requested):
@nfrazier08 
Can you separate feature branch and bug fix? You can choose patching the bug go first or the feature go first, but not both.

nfrazier08:
@zanhsieh Thanks for taking a look at this PR. I will do the bug fix in a following PR and have removed it from this PR.

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.26.1
+version: 4.27.1
```
zanhsieh, line 9:
```suggestion
version: 4.27.0
```

zanhsieh (changes requested):

zeritti (changes requested):
@nfrazier08 Thank you for your PR. This PR, though, covers too many different and unrelated changes on:
- termination params
- volume mounts
- restart policy
- support for environment variables
- container ports

Please, prefer related changes in a PR, ideally on a single topic/area. It is reasonable, indeed, to raise more than one PR on these changes.

nfrazier08:
@zeritti - Thank you, I will go ahead and split these changes up into separate PRs - starting with this one for adding termination params.

zeritti (changes requested):
When seeing a set of related parameters with a toggle, the first thing to occur to me is a map: the toggle if set enables all the fields in the map. An example of a map that one might name, say, `terminationMessageParams` in the values file:

```yaml
## values.yaml
# Enable or disable container termination message settings
# https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/
terminationMessageParams:
  enabled: false
  # If enabled, specify the path for termination messages
  terminationMessagePath: /dev/termination-log
  # If enabled, specify the Policy for termination messages
  terminationMessagePolicy: File
```

and in the template (example for the main container):

```yaml
## template
{{- if .Values.terminationMessageParams.enabled }}
{{- with .Values.terminationMessageParams }}
terminationMessagePath: {{ .terminationMessagePath }}
terminationMessagePolicy: {{ .terminationMessagePolicy }}
{{- end }}
{{- end }}
```

I have not yet seen a case where containers in the same pod are using different termination message parameters. If this indeed occurs as you foresee, the code has to be applied per container in both the values and the template like you show. Otherwise, it can be set in the values just once and shared across the two containers. 

Using `configureTerminationMessagePath` as a toggle may mislead as it does not configure just `terminationMessagePath`.

Alternatively, without a toggle:

```yaml
# Specify the path for termination messages
terminationMessagePath: ""
# Specify the Policy for termination messages
terminationMessagePolicy: ""
```

```yaml
{{- if .Values.terminationMessagePath }}
terminationMessagePath: {{ .Values.terminationMessagePath }}
{{- end }}
{{- if .Values.terminationMessagePolicy }}
terminationMessagePolicy: {{ .Values.terminationMessagePolicy }}
{{- end }}
```

nfrazier08:
@zeritti You are right, I like the idea of a toggle better and I don't think the terminationMessagePath or terminationMessagePolicy are going to be different for each container.

zeritti (changes requested):

charts/prometheus-node-exporter/values.yaml
```diff
@@ -361,10 +361,23 @@ nodeSelector:
   kubernetes.io/os: linux
   #  kubernetes.io/arch: amd64
 
+# Specify grace period for graceful termination of pods. Defaults to 30
```
zanhsieh, line 364:
```suggestion
# Specify grace period for graceful termination of pods. Defaults to 30 if null or not specified
```

zanhsieh (changes requested):

charts/prometheus-node-exporter/values.yaml
nfrazier08: Changed.

zanhsieh (approved):

PR #4225, opened 2024-02-05
[kube-prometheus-stack] allow override of for and severity rules

ps-xaf:
#### What this PR does / why we need it

Attemp to address the issue with fixed values "for" and "severity" in the helm chart, to make it more flexible to use and allow to overload those values without forking and customzing the entire chart. 

#### Which issue this PR fixes

- fixes #3384 

#### Special notes for your reviewer
```
$ docker run -it --rm --network host --workdir=/data --volume ~/.kube/config:/root/.kube/config:ro --volume $(pwd):/data --volume ~/.config/helm:/root/.config/helm quay.io/helmpack/chart-testing:v3.10.1 ct lint --chart-dirs . --charts . --validate-maintainers=false                                                                                                                                   Linting charts...
Version increment checking disabled.

------------------------------------------------------------------------------------------------------------------------
 Charts to be processed:
------------------------------------------------------------------------------------------------------------------------
 kube-prometheus-stack => (version: "56.6.3", path: ".")
------------------------------------------------------------------------------------------------------------------------

Hang tight while we grab the latest from your chart repositories...
... 
</>snip</>
...
Update Complete. ⎈Happy Helming!⎈
Saving 5 charts
Dependency crds did not declare a repository. Assuming it exists in the charts directory
Downloading kube-state-metrics from repo https://prometheus-community.github.io/helm-charts
Downloading prometheus-node-exporter from repo https://prometheus-community.github.io/helm-charts
Downloading grafana from repo https://grafana.github.io/helm-charts
Downloading prometheus-windows-exporter from repo https://prometheus-community.github.io/helm-charts
Deleting outdated charts
Linting chart "kube-prometheus-stack => (version: \"56.6.3\", path: \".\")"
Validating /data/Chart.yaml...
Validation success! 👍

Linting chart with values file "ci/01-provision-crds-values.yaml"...

==> Linting .

1 chart(s) linted, 0 chart(s) failed

Linting chart with values file "ci/02-test-without-crds-values.yaml"...

==> Linting .

1 chart(s) linted, 0 chart(s) failed

Linting chart with values file "ci/03-non-defaults-values.yaml"...

==> Linting .

1 chart(s) linted, 0 chart(s) failed

Linting chart with values file "ci/04-prometheus-operator-webhook-values.yaml"...

==> Linting .

1 chart(s) linted, 0 chart(s) failed

Linting chart with values file "ci/05-alertmanager-ingress-host-with-wildcard-values.yaml"...

==> Linting .

[...]

PR #4233, opened 2024-02-07
[kube-prometheus-stack] reduce severity of 'KubeHpaMaxedOut' to 'info'

abaguas:

The KubeHpaMaxedOut alerts when an HPA is running at maximum replicas for a given amount of time. I would claim there isn't necessarily application impact therefore the severity should be `info` instead of warning `warning`.
This is good to know to tune the HPA, and can be used to add context when paired with a highLatency or longQueue alert but on its own it doesn't require any action. In addition, there are legitimate cases where one may want to prevent an application to scale past a certain point, for example to avoid rate limiting. Therefore, I propose to reduce its severity to `info` and profit from the built-in `InfoInhibitor` alert and inhibition rule to forward with only together with higher severity alerts. 

I am conscious this is a breaking change and I am open for feedback. If my reasoning is contentious we could make it configurable.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 56.6.2
+version: 56.7.0
 appVersion: v0.71.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


GMartinez-Sisti:
Hi @abaguas, thank you for the PR. The file you're changing is automatically generated, please read the header to get instructions on the correct update process.

Regarding the actual change, since this is an individual config, the best way to change the severity is by using an `alert_relabel_configs`:

https://github.com/prometheus-community/helm-charts/blob/56d3c33402a13682349f8775489d6973c2356b1e/charts/kube-prometheus-stack/values.yaml#L3675-L3680

Something like this (not tested):
```yaml
- source_labels: [alertname]
regex: KubeHpaMaxedOut
target_label: severity
replacement: info
```

WDYT?

jkroepke:
Close this in favor of #4225

PR #4238, opened 2024-02-09
[prometheus-nginx-exporter] Fix Annotation Use in Chart

lukasklinger:
The prometheus-nginx-exporter has two ways to add additional annotations to the resources created by it, the `additionalAnnotations` field and the `podAnnotations` field. Using them in the current version of the chart either breaks the templating or - in the case of the `podAnnotations` field - has no effect.

This PR unifies the use of the `additionalAnnotations` field across the different resources and integrates the `podAnnotations` field into a new helper so that annotations added in this field have an effect.

#### Checklist

- [x] Chart Version bumped


lukasklinger:
Sorry, finally found the typo in my DCO stuff 😅

PR #4278, opened 2024-02-22
[prometheus-rabbitmq-exporter] Allows reading User from secret

sulochan:
Currently the exporter allows using password from secret. This patch allows users to read username from secret as well. This is used in cases where automated username is created during install process of rabbitmq service.

Signed-off-by: Sulochan Acharya <sulochan@gmail.com>


monotek (changes requested):
chart version must be raised

monotek (approved):

PR #4304, opened 2024-02-27
[prometheus-snmp-exporter] configmap-reload change image to quay.io

MohammadrezaTavakoli:
@walker-tom 

This PR Changes to configmap-reload image in snmp-exporter chart from jimmidyson/configmap-reload to quay.io/prometheus-operator/prometheus-config-reloader.

- fixes #
#3819 

#### Checklist
- [x] Chart Version Bumped
- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Title of the PR starts with chart name

PR #4305, opened 2024-02-27
[prometheus-snmp-exporter] use quay as default image repository

onedr0p:
We should not be using Dockerhub when these images are available on quay, this will make it more consistent with the other images from the other charts in this Helm repo.

This change should be transparent and need not a breaking change.

Depends on https://github.com/prometheus-community/helm-charts/pull/4304

- fixes #

#### Checklist

- [x] Chart Version bumped


onedr0p:
@walker-tom https://github.com/prometheus-community/helm-charts/pull/4304 should be merged first

charts/prometheus-snmp-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 3.0.0
+version: 3.0.2
```
walker-tom, line 4: I think this should probably be `5.0.0` as it is technically a breaking change as we're changing the default of an existing value. It would likely be transparent to users as you mentioned but if anyone is mirroring images in a private registry or has some security constraints on the registries they access etc then they would need to actively make changes.
```suggestion
version: 5.0.0
```

walker-tom (changes requested):
Thanks for the PR, good to get this over to Quay.io! Please could you make the requested Chart version change and add an extra section to the README to highlight that this is a breaking change version, similar to how we did in #4304. Then I'll be able to get this approved and merged 👍

charts/prometheus-snmp-exporter/Chart.yaml
onedr0p: Updated

walker-tom (approved):

PR #4312, opened 2024-02-28
[kube-prometheus-stack] Allow setting selectors for built-in ServiceMonitors

iandrewt:

This will make it easier to install the chart in environments where resources cannot be created in the kube-system namespace, such as a tightly scoped ArgoCD AppProject.

Without this change, to use the built-in ServiceMonitors with service.enabled=false, the cluster operator is expected to create services in the kube-system namespace that have labels that exactly match what the chart is expecting.

With this change, I can create these service resources with my own naming conventions and labels before installing the helm chart, and inform the ServiceMonitor on how to select them.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -54,7 +54,7 @@ coreDns:
   serviceMonitor:
     selector:
       matchLabels:
-        k8s-app: kube-dns
+        k8s-app: '{{ $.Release.Name }}'
 kubeEtcd:
   service:
     enabled: false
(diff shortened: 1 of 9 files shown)


jkroepke:
Hi, thanks for your PR!

Could you please add some working non defaults here?

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml

The goal is that the both conditions (if way/else way) are covered by ci tests

iandrewt:
Done - if the tests are running in a real environment, then likely only CoreDNS will work as it uses the default service labels for kube-dns, but the remaining services likely won't get readings. Otherwise, should pass any lint testing.

charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
```diff
@@ -40,3 +40,39 @@ customRules:
...
+    selector:
+      matchLabels:
+        component: kube-controller-manager
+coreDns:
+  service:
+    enabled: false
+  serviceMonitor:
+    selector:
+      matchLabels:
+        k8s-app: kube-dns
```
jkroepke, line 57:
```suggestion
        k8s-app: '{{ $.Release.Name }}'
```

PR #4313, opened 2024-02-28
[kube-state-metrics] Added documentation for updateStrategy

### LINKED ISSUE #4095: [kube-state-metrics] Strategy RollingUpdate causes 'duplicate' metrics

dotdc:

This PR aims to merge the documentation changes made in #4096 in order to fix #4095.

- fixes #4095

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 56.15.0
+version: 56.16.0
 appVersion: v0.71.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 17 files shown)


charts/kube-state-metrics/values.yaml
```diff
@@ -37,7 +37,9 @@ autosharding:
 
 replicas: 1
 
-# Change the deployment strategy when autosharding is disabled
+# Change the deployment strategy when autosharding is disabled.
+# ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
+# During a release, 'RollingUpdate' can lead to two running instances for a short period of time while 'Recreate' can create a small gap in data.
 # updateStrategy: Recreate
```
mrueg, line 43: Can you mention what the default updateStrategy is?

PR #4317, opened 2024-02-29
[kube-prometheus-stack] Allow changing port name of built-in ServiceMonitors

iandrewt:

Following on from #4312, I found while trying to use the built-in kube-dns service instead of just a custom externally-managed one for scraping metrics that the default configuration calls the metrics port `metrics`, not `http-metrics` like the chart expects.

This change allows overriding this value for all the built-in ServiceMonitors that create Services, and setting it will change the port name for the Helm managed services to ensure setting it without using custom Services continues to work.

- fixes #

#### Checklist

- [x] Chart Version bumped


jkroepke (approved):
At some point in the feature, each property is exposed :D

PR #4319, opened 2024-02-29
[kube-prometheus-stack] Allow changing the reloader web port in the Prometheus Service

svetlimladenov:

This PR enables the configuration of the port value for the `reloader-web` port in the Prometheus Service Object. 

Port `8080` is a commonly used port, so it's possible that it's already taken.

Actually that's exactly what happened in our setup in MacStadium. We use a LoadBalancer Service to expose Prometheus to our clients. We reuse the same load balancer IP and unfortunately we already have a service listening on port 8080, so the Prometheus Service can't use it. 

With this change we are able to change the `reloader-web` Service Port to another value

- fixes #

#### Checklist

- [x] Chart Version bumped


Xtigyro (approved):

Xtigyro:
@svetlimladenov Thanks! And regards from Sofia!

PR #4337, opened 2024-03-05
[kube-state-metrics] Ignore global values for kube-state-metrics

megum1n:

Currently, I can set [`.Values.global.imageRegistry`](https://github.com/prometheus-community/helm-charts/blob/714322e38a04746fec8ec32f362bbdebc33ddbdd/charts/kube-prometheus-stack/values.yaml#L223) in the `kube-prometheus-stack` chart and this value will be used for all images in the chart and subcharts.
So, if I want to use my private registry for all images, I have two options:
1. Override registry for every image in values file (a lot of overrides)
2. Set global registry in values file and replicate all images while keeping original path (`/kube-state-metrics/kube-state-metrics`, `/prometheus/alertmanager`)

Almost all images in the chart are located in the `quay.io` registry, but `kube-state-metrics` image is from `registry.k8s.io`.
Private registries often use pull trough cache, so there is no option to keep original path in the registry. Path must start with the prefix, depending on the source registry.
i.e. `registry.k8s.io/kube-state-metrics/kube-state-metrics` => `private.registry.com/k8s/kube-state-metrics/kube-state-metrics`

I suggest a third option: 
- Set global registry in values file and use `.Values.ignoreGlobal` flag in `kube-state-metrics` chart to ignore it. Then I can safely override registry via `.Values.kube-state-metrics.image.registry` and there will be less overrides in my value file.

This w/a is kinda janky, please let me know if you have a better w/a for the problem.

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/templates/_helpers.tpl
+++ b/charts/kube-state-metrics/templates/_helpers.tpl
@@ -108,7 +108,15 @@ labelValueLengthLimit: {{ . }}
 Formats imagePullSecrets. Input is (dict "Values" .Values "imagePullSecrets" .{specific imagePullSecrets})
 */}}
 {{- define "kube-state-metrics.imagePullSecrets" -}}
-{{- range (concat .Values.global.imagePullSecrets .imagePullSecrets) }}
+
+{{- $pullSecrets := "" }}
+{{- if .Values.ignoreGlobal }}
+{{- $pullSecrets = .imagePullSecrets }}
+{{- else }}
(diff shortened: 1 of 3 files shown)


dotdc:
Hi @megum1n,
Apologies for the response time, and thank you for your detailed proposal!
While I understand your point, adding `.Values.ignoreGlobal` introduces complexity and inconsistency, so I think we should stick with overrides in values.

PR #4340, opened 2024-03-06
[prometheus-pushgateway] Add config for hard/soft podAntiAffinity

james-jra:

This PR adds simple configuration for hard/soft podAntiAffinity for prometheus-pushgateway. The implementation is based on the existing pattern in [alertmanager](https://github.com/prometheus-community/helm-charts/blob/f00193179b8d083e8b609ef587c72d66c633072e/charts/alertmanager/templates/statefulset.yaml#L58)

It is already possible to configure podAntiAffinity using the `affinity` config field, however you need to provide the entire podAntiAffinity YAML object. Notably, this includes a selector which matches on Pod labels, therefore requires the developer writing this configuration to know how Pod labels will be rendered. Adding a simple hard/soft `podAntiAffinity` configuration option therefore simplifies a common configuration requirement and provides some extra backwards compatibility for developers' configuration if the Pod labels do change in future.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-pushgateway/values.yaml
+++ b/charts/prometheus-pushgateway/values.yaml
@@ -213,6 +213,18 @@ containerSecurityContext: {}
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
 affinity: {}
 
+## Pod anti-affinity can prevent the scheduler from placing push-gateway replicas on the same node.
+## The default value "soft" means that the scheduler should *prefer* to not schedule two replica pods onto the same node but no guarantee is provided.
+## The value "hard" means that the scheduler is *required* to not schedule two replica pods onto the same node.
+## The value "" will disable pod anti-affinity so that no anti-affinity rules will be configured.
+##
+podAntiAffinity: ""
+
+## If anti-affinity is enabled sets the topologyKey to use for anti-affinity.
+## This can be changed to, for example, failure-domain.beta.kubernetes.io/zone
+##
+podAntiAffinityTopologyKey: kubernetes.io/hostname
+
 ## Topology spread constraints for pods
 ## Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
 topologySpreadConstraints: []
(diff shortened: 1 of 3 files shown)


james-jra:
@gianrubio @cstaud @zeritti (PR template recommends pinging Chart owners for quick resolution).
This PR will be followed by a similar one for the prometheus Helm Chart too.

charts/prometheus-pushgateway/values.yaml
```diff
@@ -213,6 +213,18 @@ containerSecurityContext: {}
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
 affinity: {}
 
+## Pod anti-affinity can prevent the scheduler from placing push-gateway replicas on the same node.
```
zeritti, line 216:
```suggestion
## Pod anti-affinity can prevent the scheduler from placing pushgateway replicas on the same node.
```

charts/prometheus-pushgateway/values.yaml
```diff
@@ -213,6 +213,18 @@ containerSecurityContext: {}
...
+## The default value "soft" means that the scheduler should *prefer* to not schedule two replica pods onto the same node but no guarantee is provided.
```
zeritti, line 217: I am not sure about the default value as the present default is an empty string.

```suggestion
## The value "soft" means that the scheduler should *prefer* to not schedule two replica pods onto the same node but no guarantee is provided.
```

zeritti (changes requested):
Thank you, @james-jra, for your PR. Please, see my comments below.

charts/prometheus-pushgateway/values.yaml
james-jra: Ah, sorry. The default value of an empty string will have no affect - no podAntiAffinity configuration will be applied. I'll fix this comment.

charts/prometheus-pushgateway/values.yaml
james-jra: Done

charts/prometheus-pushgateway/values.yaml
james-jra: Done

zeritti (approved):
Tnaks for your changes, @james-jra, LGTM.

PR #4344, opened 2024-03-06
[kube-state-metrics] make possible to globally override a specific image registry

yann-soubeyrand:

This feature is necessary in cases where repositories coming from different registries are stored in different namespaces in your registry. For example, this is the case with AWS ECR pull through cache feature, which uses the repository prefix to know where to fetch the image from (https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache-creating-rule.html).

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
   - kubernetes
 type: application
 version: 5.19.0
-appVersion: 2.12.0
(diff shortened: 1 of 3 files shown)


dotdc:
Hi @yann-soubeyrand,
Thank you for opening this pull request and apologies for the response time, 
Global settings are intended to be global, if you need fine-grained configuration, it's best to leave the global values unset and configure the registry individually for each image instead.

Issue #4033, opened 2023-11-22
[kube-prometheus-stack] Conflicting secret name


applejag:
### Describe the bug a clear and concise description of what the bug is.

Tried to do an update via `helm upgrade`, but got the error:

`rendered manifests contain a resource that already exists. Unable to continue with update: Secret "prometheus-prometheus" in namespace "prometheus" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "ri-prometheus-operator"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "prometheus"`

It seems to be caused by this file: https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-54.1.0/charts/kube-prometheus-stack/templates/prometheus/secret.yaml

Having conflicts with a file that the Prometheus Operator generates here: https://github.com/prometheus-operator/prometheus-operator/blob/v0.69.1/pkg/prometheus/statefulset.go#L122-L148

### What's your helm version?

version.BuildInfo{Version:"v3.13.2", GitCommit:"v3.13.2", GitTreeState:"", GoVersion:"go1.21.3"}

### What's your kubectl version?

Client Version: v1.28.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.8

### Which chart?

kube-prometheus-stack

### What's the chart version?

54.1.0

### What happened?

The secret from this chart is for Thanos stuff, and I don't even use Thanos. The secret that's blocking my upgrades is an empty secret.

### What you expected to happen?

The secret deployed by this chart does not have a conflicting name.

### Enter the changed values of values.yaml?

```yaml
fullnameOverride: prometheus
cleanPrometheusOperatorObjectNames: true
```

### Enter the command that you execute and failing/misfunctioning.

```bash
helm upgrade ri-prometheus-operator ./ -f path-to-my-values/values.yaml
```

timebertt:
Seems to be caused by https://github.com/prometheus-community/helm-charts/pull/3918 (version 52).
cc @thameezb

The secret is always created
https://github.com/prometheus-community/helm-charts/blob/f61d89979f347f31c42399ff1ec084b3870c523a/charts/kube-prometheus-stack/templates/prometheus/secret.yaml#L1

[...]

PR #3918, opened 2023-10-19
[kube-prometheus-stack] fixes thanos configs

thameezb:
#### What this PR does / why we need it
Updates thanos configs to allow the usage of existing secrets or the option of creating a new secret object
Full design discussion can be found in the issue 

#### Which issue this PR fixes
fixes https://github.com/prometheus-community/helm-charts/issues/3869

#### Checklist
- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #4346, opened 2024-03-07
[kube-prometheus-stack] Change thanos unused secret condition

bogdan-at-adobe:

Following https://github.com/prometheus-community/helm-charts/pull/3918 a secret has been added for the thanos configs, however that secret gets applied regardless of whether the user uses it or not (if not used, it is just empty). This fix changes the condition to better match its purpose.

It is not a direct related issue but the cause is the same in https://github.com/prometheus-community/helm-charts/issues/4033 

#### Checklist

- [x] Chart Version bumped


Allex1 (approved):

charts/kube-prometheus-stack/templates/prometheus/secret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.prometheus.enabled }}
+{{- if and .Values.prometheus.enabled .Values.prometheus.prometheusSpec.thanos.objectStorageConfig }}
```
jkroepke, line 1: What about

.Values.prometheus.prometheusSpec.thanos.objectStorageConfig.existingSecret

?

If defined, then this secret can be omit, too? Like line 13

bogdan-at-adobe: I have applied the suggestion. Moved the if statement to the beginning of the manifest.

jkroepke (changes requested):
LGTM 👍

jkroepke (dismissed):
LGTM 👍

bogdan-at-adobe:
The error seems to be on a grafana-test pod, I don't think this PR should interfere. Could you please do a retest? @jkroepke

jkroepke (approved):
LGTM 👍

PR #4354, opened 2024-03-11
[prometheus-blackbox-exporter] fix: only template env if .Values.extraEnv is specified

adberger:

Currently when `.Values.extraEnv` is omitted, env is still templated in the deploymen spec.
ArgoCDs new [Server-Side Diff](https://argo-cd.readthedocs.io/en/stable/user-guide/diff-strategies/#server-side-diff) is not happy about it.

```
Failed to compare desired state to live state: failed to calculate diff: error calculating server side diff: serverSideDiff error: error removing non config mutations for resource Deployment/blackbox-exporter-prometheus-blackbox-exporter: error reverting webhook removed fields in predicted live resource: .spec.template.spec.containers: element 0: associative list with keys has an element that omits key field "name" (and doesn't have default value)
```

This PR solves this issue.

#### Checklist

- [x] Chart Version bumped


adberger:
not stale

monotek (approved):
The need to add ArgoCD specific fixes just sucks but approving anyway...

adberger:
> The need to add ArgoCD specific fixes just sucks but approving anyway...

agree..

PR #4369, opened 2024-03-20
Add 'release' label to prometheus-pushgateway service monitor

nbelov-magdv:
'release' label is required for integrating with kube-prometheus-stack

Adds 'release' label to prometheus-pushgateway ServiceMonitor.
This label is used by prometheus from kube-prometheus-stack for collecting metrics.

I cannot use {{ .Release.Name }} as value in values.yaml. So I've add it to template.

Also, I don't understand why kube-prometheus-stack uses 'release' label, but not 'app.kubernetes.io/instance' label.
But I am not good in k8s.

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus-pushgateway/templates/_helpers.tpl
+++ b/charts/prometheus-pushgateway/templates/_helpers.tpl
@@ -74,6 +74,7 @@ Selector labels
 {{- define "prometheus-pushgateway.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "prometheus-pushgateway.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+release: {{ .Release.Name }}
 {{- end }}
 
 {{/*


zeritti (changes requested):
Thank you, @nbelov-magdv, for your PR.

> 'release' label is required for integrating with kube-prometheus-stack

It is not entirely true. By default, Prometheus operator as deployed through kube-prometheus-stack sets a service monitor selector to `release: RELEASE_NAME`. Keeping it is not required, though, as this service monitor selector can be cleared by means of `prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues=false` and, optionally, a custom selector can be set through `prometheus.prometheusSpec.serviceMonitorSelector` ([serviceMonitorSelector](https://github.com/prometheus-community/helm-charts/blob/f1b95070c5ff3c43c400781c64a5dcc5f5f99d15/charts/kube-prometheus-stack/values.yaml#L3486)). 

If such a selector is set in prometheus CR, either default or custom, the corresponding label can be set in pushgateway chart's `serviceMonitor.additionalLabels`. Adding the release label to all manifests' labels and selectors by default seems unnecessary. We could, however, consider optionally inserting the release label in labels through a boolean field e.g. `releaseLabel`.

PR #4388, opened 2024-03-23
[prometheus-elasticsearch-exporter] add ability to customize apiversion for pod/service monitor

rashmichandrashekar:

Adding support for customizing api version for pod/service monitor as it is useful for support with Azure managed prometheus that looks for pod/service monitors in its own api group

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
+++ b/charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
@@ -1,11 +1,8 @@
 {{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
 {{- fail "Either .Values.podMonitor.enabled or .Values.serviceMonitor.enabled can be enabled at a time, but not both." }}
 {{- else if .Values.podMonitor.enabled }}
-{{- if .Values.azuremanagedprometheus.use_azuremanagedprometheus }}
-apiVersion: azmonitoring.coreos.com/v1
-{{- else }}
-apiVersion: monitoring.coreos.com/v1
-{{- end }}
+{{- if .Capabilities.APIVersions.Has .Values.podMonitor.apiVersion}}
+apiVersion: {{ .Values.podMonitor.apiVersion }}
 kind: PodMonitor
 metadata:
   name: {{ template "elasticsearch-exporter.fullname" . }}
@@ -51,4 +48,5 @@ spec:
     matchLabels:
       app: {{ template "elasticsearch-exporter.name" . }}
       release: "{{ .Release.Name }}"
+{{- end }}
 {{- end }}
--- a/charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
+++ b/charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
@@ -1,7 +1,11 @@
 {{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
 {{- fail "Either .Values.podMonitor.enabled or .Values.serviceMonitor.enabled can be enabled at a time, but not both." }}
 {{- else if .Values.podMonitor.enabled }}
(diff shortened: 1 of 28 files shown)


charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
```diff
@@ -1,7 +1,11 @@
 {{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
 {{- fail "Either .Values.podMonitor.enabled or .Values.serviceMonitor.enabled can be enabled at a time, but not both." }}
 {{- else if .Values.podMonitor.enabled }}
+{{- if .Values.azuremanagedprometheus.use_azuremanagedprometheus }}
+apiVersion: azmonitoring.coreos.com/v1
+{{- else }}
 apiVersion: monitoring.coreos.com/v1
```
DrFaust92, line 7: If schema for this objects is really the same(i have no idea), it might be easier to just template `apiVersion` value and keep `monitoring.coreos.com/v1` default for backwards compatibility as we might have more of these in the future (GCP has a similar config)

DrFaust92 (changes requested):

charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
DrFaust92: I would also potentially add a check with something like`{{- if .Capabilities.APIVersions.Has "azmonitoring.coreos.com/v1" }}` for extra validation

rashmichandrashekar: Thanks @DrFaust92 for the feedback. The schema is the same. I have addressed your feedback. Please let me know if this looks good.

rashmichandrashekar:
gentle ping @DrFaust92  @desaintmartin @zeritti. Please take a look and let me know if you see any concerns. Thanks!

DrFaust92 (approved):
LGTM, but im not maintainer 💀

charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
```diff
@@ -1,7 +1,8 @@
 {{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
 {{- fail "Either .Values.podMonitor.enabled or .Values.serviceMonitor.enabled can be enabled at a time, but not both." }}
 {{- else if .Values.podMonitor.enabled }}
-apiVersion: monitoring.coreos.com/v1
+{{- if .Capabilities.APIVersions.Has .Values.podMonitor.apiVersion}}
```
zeritti, line 4: This conditional should not be present, a literal is just being replaced with a variable after all.

charts/prometheus-elasticsearch-exporter/templates/servicemonitor.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.serviceMonitor.enabled }}
 ---
-apiVersion: monitoring.coreos.com/v1
+{{- if .Capabilities.APIVersions.Has .Values.podMonitor.apiVersion}}
```
zeritti, line 3: As in the comment above.

zeritti (changes requested):
Thank you, @rashmichandrashekar, for your PR. Please, see my comments below.

charts/prometheus-elasticsearch-exporter/templates/podmonitor.yaml
rashmichandrashekar: Thanks @zeritti. I have removed the check from pod and svc monitors. Please take a look. Also, how does the release process work? If there a scheduled release for the charts in this repo?

zeritti (approved):

rashmichandrashekar:
thanks @zeritti. Will you be able to merge this and create a release too?

PR #4400, opened 2024-03-27
[prometheus-adapter] allow using cluster roles for auth reader

vince-vibin:
When deploying the Prometheus Adapter in a Namespace the `auth-reader` role is still created in the `kube-system` namespace. Deploying into `kube-system` is not wanted for a lot of enterprise clusters.
This give the option to deploy the `auth-reader` role as a cluster role rather then a normal role. Now the User can easily deploy everything into one namespace

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 4.9.1
+version: 4.9.2
 appVersion: v0.11.2
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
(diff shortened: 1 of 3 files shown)


vince-vibin:
Hi @hectorj2f, @MattiasGees, @steven-sheehy could you please check this out 🙂

charts/prometheus-adapter/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 4.9.1
+version: 4.9.2
```
steven-sheehy, line 3: Should be a minor bump since it's a new feature.

```suggestion
version: 4.10.0
```

steven-sheehy (changes requested):

charts/prometheus-adapter/Chart.yaml
vince-vibin: should be fixed. forgot the DOC sign at first 😅

mmcloud:
Hi, studying this PR makes me ask is more secure here? Deploying the adapter using a namespaced scoped role binding on the kube-system or in another namespace but using the cluster scoped resource?

PR #4408, opened 2024-03-30
[prometheus-json-exporter]: Allow additional container args and service monitor relabelings

faisal-khalique:
- Adds the option to pass arguments to the container. Could be needed for many reason. For example collecting logstash metrics with many pipelines where the pipeline definitions vary based on environment. Having the option of providing argument `--config.file` will greatly simplify such deployments. Considering future use cases, adding an array instead of just specifying the argument `--config.file` now
- Add additional relabelings for service-monitor which can be used to drop unwanted labels from the metrics like `container`, `pod`, etc.

#### Checklist
- [x] Chart Version bumped


charts/prometheus-json-exporter/templates/deployment.yaml
```diff
@@ -42,6 +42,13 @@ spec:
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}
+          {{- if .Values.extraArgs }}
```
zanhsieh, line 45:
```suggestion
          {{- with .Values.extraArgs }}
````

File: charts/prometheus-json-exporter/templates/deployment.yaml [hunk 2]
```diff
@@ -42,6 +42,13 @@ spec:
...
+          args:
+          {{- range .Values.extraArgs }}
+          - {{ .name | quote }}
+          - {{ .value | quote }}
```
zanhsieh, line 49: I don't think this correct.

zanhsieh (changes requested):

charts/prometheus-json-exporter/templates/deployment.yaml
faisal-khalique: I believe it is. The changes in this PR is currently deployed in one of our environments.

faisal-khalique:
@zanhsieh - can you please check this PR?

zanhsieh (approved):

PR #4435, opened 2024-04-09
Fix env for jiralert chart

AKYD:

Trying to use `env` gives the following error:

`Error: template: jiralert/templates/deployment.yaml:70:43: executing "jiralert/templates/deployment.yaml" at <.Values.env>: nil pointer evaluating interface {}.env`

Also `values.yaml` doesn't show that `env:` is an option; one needs to check out the `deployment.yaml` to see that it's used

### DIFF
--- a/charts/jiralert/templates/deployment.yaml
+++ b/charts/jiralert/templates/deployment.yaml
@@ -67,7 +67,7 @@ spec:
           {{- end }}
         {{- with .Values.env }}
           env:
-          {{- range $key, $value := $.Values.env }}
+          {{- range $key, $value := . }}
             - name: {{ $key }}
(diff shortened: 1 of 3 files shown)


charts/jiralert/templates/deployment.yaml
```diff
@@ -67,7 +67,7 @@ spec:
           {{- end }}
         {{- with .Values.env }}
           env:
-          {{- range $key, $value := .Values.env }}
+          {{- range $key, $value := $.Values.env }}
```
jkroepke, line 70:
```suggestion
          {{- range $key, $value := . }}
```

would make more sense?

jkroepke (approved):

PR #4469, opened 2024-04-18
[prometheus-blackbox-exporter] Add value for hostPort

### LINKED ISSUE #4468: [prometheus-blackbox-exporter] hostPort setting

The chart has a `kind` value which allows the user to deploy blackbox as a DaemonSet instead of a Deployment.

When deployed as a DaemonSet, I want to connect to the individual blackbox pods instead of the load-balanced service, to probe from specific nodes.

In k8s there are several ways for communicating with daemon pods. See: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#communicating-with-daemon-pods

`NodeIP and Known Port` option using `hostPort` setting is the most suitable for me. However the chart doesn't support setting `hostPort`.

Modify the chart to allow setting `hostPort` in the container spec.

I have considered using the existing values to achieve the same effect. The only option seems using `extraManifests` value, but it is too cumbersome for a single line change in the existing manifest.

### Additional context.

_No response_

ozen:

- fixes #4468

#### Checklist

- [ ] Chart Version bumped

@desaintmartin 
@gianrubio 
@rsotnychenko 
@monotek

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
+++ b/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
@@ -245,6 +245,9 @@ containers:
   ports:
   - containerPort: {{ .Values.containerPort }}
     name: http
+    {{- if .Values.hostPort }}
+    hostPort: {{ .Values.hostPort }}
+    {{- end }}
   livenessProbe:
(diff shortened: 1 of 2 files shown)


monotek (changes requested):
chart version needs to be raised

monotek (approved):

PR #4482, opened 2024-04-23
[prom-label-proxy] add path templating

kogeler:

It is useful for the case when we use one domain name for a bunch of prom-label-proxy deployments. In this case, we have to use different paths to route requests, but we need templating for paths to implement it.

It doesn't break backward capabilities for existing value files.

#### Checklist
- [x] Chart Version bumped


jkroepke (approved):
LGTM 👍

PR #4488, opened 2024-04-25
[prometheus-snmp-exporter] Extend workload config

ngc4579:
- Introduce optional revisionHistoryLimit for Deployment / DaemonSet

@Miouge1 @xiu @walker-tom 

Make the Deployment / DaemonSet `revisionHistoryLimit` configurable by introducing an optional Parameter to `values.yaml`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 5.1.2
+version: 5.2.0
 appVersion: v0.25.0
 home: https://github.com/prometheus/snmp_exporter
 sources:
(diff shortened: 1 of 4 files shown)


charts/prometheus-snmp-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 5.1.1
+version: 5.1.2
```
walker-tom, line 4: Can we update this to be a minor version change rather than a patch/bugfix as its a new feature that's backwards compatible, thanks!
```suggestion
version: 5.2.0
```

walker-tom (changes requested):
Thanks for the PR!

walker-tom (approved):

charts/prometheus-snmp-exporter/Chart.yaml
ngc4579: Yes, of course - thanks for the hint.

PR #4497, opened 2024-04-29
[prometheus-elasticsearch-exporter] Allow to override imageRegistry globally

hierynomus:
…lobally

This PR allows to override the image registry for the `prometheus-elasticsearch-exporter` using a `global.imageRegistry` flag, which is a common way to override the registry when this chart is bundled as a sub-chart.

The PR is made such that it is backwards compatible with the current version of the chart so that only a patch increment is needed according to semver. 

@svenmueller @desaintmartin @zeritti 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/Chart.yaml
+++ b/charts/prometheus-elasticsearch-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 5.7.1
+version: 5.8.0
 kubeVersion: ">=1.10.0-0"
 appVersion: "v1.7.0"
 home: https://github.com/prometheus-community/elasticsearch_exporter
--- a/charts/prometheus-elasticsearch-exporter/Chart.yaml
+++ b/charts/prometheus-elasticsearch-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 5.7.0
+version: 5.7.1
 kubeVersion: ">=1.10.0-0"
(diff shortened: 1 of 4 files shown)


charts/prometheus-elasticsearch-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 5.7.0
+version: 5.7.1
```
zeritti, line 4: This change is not really a bug fix and as a result, the minor number should be raised - i.e. a new feature is being introduced.

```suggestion
version: 5.8.0
```

zeritti (changes requested):
Thank you, @hierynomus, for your PR. Please, see my comments below.

charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
```diff
@@ -75,7 +75,7 @@ spec:
           - secretRef:
               name: {{ .Values.envFromSecret }}
           {{- end }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: "{{- include "elasticsearch-exporter.image.repository" .}}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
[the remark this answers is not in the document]
hierynomus: Updated!

charts/prometheus-elasticsearch-exporter/Chart.yaml
hierynomus: Updated!

hierynomus:
@zeritti Many thanks for the fast review. Comments are addressed.

zeritti (approved):
Thank you, @hierynomus, LGTM.

PR #4503, opened 2024-04-30
[kube-prometheus-stack] remove job ttl api check

### LINKED ISSUE #4500: [kube-prometheus-stack] deployment via argocd stuck in "pending deletion" for kube-prometheus-stack-admission-create job

When deploying kube-prometheus-stack with argocd sometimes it gets stuck in "pending deletion" for object job kube-prometheus-stack-admission-create. We need to terminate the argocd sync and resync manually to get out of this situation, which is obviously not good in terms of full automation.

image

According to https://github.com/argoproj/argo-cd/issues/6880 this is a well known problem when a PreHook Job has a 

```
ttlSecondsAfterFinished: 0
```

defined, because then kubernetes deletes this job immediatly after it finished and also argocd wants to delete this job and so we have some kind of race condition here.

I would propose to make the value for ttlSecondsAfterFinished in https://github.com/prometheus-community/helm-charts/blob/9c41858ac9714483638d78fb560577dc37e55875/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml#L19 configurable via a helm value.

### What's your helm version?

3.14.3

v1.29.2

### Which chart?

kube-prometheus-stack

### What's the chart version?

58.2.2

### What happened?

When deploying kube-prometheus-stack with argocd sometimes it gets stuck in "pending deletion" for object job kube-prometheus-stack-admission-create. We need to terminate the argocd sync and resync manually to get out of this situation, which is obviously not good in terms of full automation.

the argocd sync of this chart works without failures are stuck problems

### How to reproduce it?

setup an argocd application with

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: kube-prometheus-stack-jokl
spec:
  destination:
    name: ''
    namespace: jokl-test
    server: 'https://kubernetes.default.svc'
  source:
    path: ''
    repoURL: 'https://prometheus-community.github.io/helm-charts'
    targetRevision: 58.2.2
    chart: kube-prometheus-stack
  sources: []
  project: default
  syncPolicy:
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true
 ```

since it is a race condition it doesn't happen always, but very often

NONE

n.a.

_No response_

### PR DESCRIPTION
**Pull request by DrFaust92**

Removes api check for `ttlSecondsAfterFinished` as the job resource is already `batch/v1` so the check for `batch/v1alpha1` is redundent.
It also allows overriding the existing value.

- fixes #4500

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 58.3.0
+version: 58.3.1
 appVersion: v0.73.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml
@@ -14,10 +14,7 @@ metadata:
     app: {{ template "kube-prometheus-stack.name" $ }}-admission-create
     {{- include "kube-prometheus-stack.prometheus-operator-webhook.labels" $ | nindent 4 }}
 spec:
-  {{- if .Capabilities.APIVersions.Has "batch/v1alpha1" }}
-  # Alpha feature since k8s 1.12
-  ttlSecondsAfterFinished: 0
-  {{- end }}
+  ttlSecondsAfterFinished: {{ .Values.prometheusOperator.admissionWebhooks.patch.ttlSecondsAfterFinished }}
   template:
     metadata:
       name:  {{ template "kube-prometheus-stack.fullname" . }}-admission-create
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by jkroepke — APPROVED** [review]
LGTM 👍

**Comment by tulsluper** [comment]
When I try to install the chart 58.3.1 with ArgoCD - it stucks on  "waiting for completion of hook batch/Job/kube-prometheus-stack-admission-create".
Everything fine with 58.3.0

Argo app additional settings:
- ServerSideDiff=true
- Server-Side Apply

**Comment by jkroepke** [comment]
Seems like an known bug in ArgoCD: https://github.com/argoproj/argo-cd/issues/8586

Since there is a different behavior between `ttlSecondsAfterFinished: 0` and `ttlSecondsAfterFinished` is omit, we need a condition, where `0` is true-ish, but empty value is false.

What about this?

```suggestion
  {{- if not (eq (print .Values.prometheusOperator.admissionWebhooks.patch.ttlSecondsAfterFinished) "") }}
  ttlSecondsAfterFinished: {{ .Values.prometheusOperator.admissionWebhooks.patch.ttlSecondsAfterFinished }}
  {{- end }}
```

With that condition, it would be possible to omit `ttlSecondsAfterFinished` completely to establish ArgoCD compatibility.

reefland:
Seeing `waiting for completion of hook batch/Job/prometheus-admission-create`, unlike @tulsluper , I have to roll-back to `58.2.2`, as `58.3.0` also hangs.

PR #4526, opened 2024-05-08
[kube-prometheus-stack] Add AlertmanagerConfig v1beta1 CRD

jkroepke:

This PR adds the AlertmanagerConfig v1beta1 CRD. The v1beta1 brings support for additional validation via ValidationWebhook.

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/charts/crds/crds/crd-alertmanagerconfigs.yaml
+++ b/charts/kube-prometheus-stack/charts/crds/crds/crd-alertmanagerconfigs.yaml
@@ -1,5 +1,4 @@
-# https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.73.2/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
----
+# https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.73.2/example/prometheus-operator-crd-full/monitoring.coreos.com_alertmanagerconfigs.yaml
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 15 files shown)


charts/kube-prometheus-stack/charts/crds/crds/crd-alertmanagerconfigs.yaml
```diff
@@ -8,6 +7,18 @@ metadata:
     operator.prometheus.io/version: 0.73.2
   name: alertmanagerconfigs.monitoring.coreos.com
 spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: prometheus-operator-admission-webhook
```
jkroepke, line 15: I got this from here: https://prometheus-operator.dev/docs/user-guides/webhook/#converting-alertmanagerconfig-resources

However, I guess this can not be never a valid configuration for kube-prometheus-stack. Such settings needs to be templated, but helm doesn't allow any templating here.

Any ideas to solve the issue?

GMartinez-Sisti: This will probably need to be patched from a hook since we can never know the name of the chart or we decide on a fixed name :/.

For the first option, if someone is using tools to manage these deployments the change will have to be ignored.

jkroepke: And if people apply CRD updates after upgrade, they may miss that, too.

jkroepke: What about an static service name?

GMartinez-Sisti: I agree that is the safer option.

PR #4533, opened 2024-05-10
[kube-prometheus-stack] fix the retentionSize value

karloramone:

There is a bug in the chart related to the `prometheus.prometheusSpec.retentionSize` value.

In the default values currently there is specified value for `prometheus.prometheusSpec.retention: 10d`.
When you try to add the value `prometheus.prometheusSpec.retentionSize` it will also generate the `prometheus.prometheusSpec.retention: 10d` from those default values.

This is a problem because according to the [docs](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects) only one of those values can be configured. 

```
If both time and size retention policies are specified, whichever triggers first will be used.
```

#### Checklist

- [x ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 58.5.0
+version: 58.5.1
 appVersion: v0.73.2
 kubeVersion: ">=1.19.0-0"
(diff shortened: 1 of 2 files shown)


QuentinBisson:
@karloramone where in the docs does it state that you can only configure one or the other? You can define both and Prometheus makes it an or between retention time and retention size. You can set the value to nil in your values if you do not want it

QuentinBisson:
I'm closing here as this is not a bug, feel free to reopen otherwise

karloramone:
@QuentinBisson Yes, looks like I misunderstood the docs and had some other problem. Thanks

PR #4536, opened 2024-05-13
[prometheus-ipmi-exporter] Optionally pull config from a Secret

yankcrime:

The way this currently works is that it builds a config file from values inlined via Helm, and this may need to include some sensitive information such as usernames and passwords that should be otherwise kept out of a repo in the cases where you're managing all of your cluster configuration via git.

This PR update the Chart to add an option to be able to reference a Secret instead.

#### Checklist

### DIFF
--- a/charts/prometheus-ipmi-exporter/templates/configmap.yaml
+++ b/charts/prometheus-ipmi-exporter/templates/configmap.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.configMapFile -}}
+{{- if not .Values.configSecret.enabled -}}
 {{- $fullName := include "prometheus-ipmi-exporter.fullname" . -}}
 apiVersion: v1
 kind: ConfigMap
--- a/charts/prometheus-ipmi-exporter/templates/deployment.yaml
+++ b/charts/prometheus-ipmi-exporter/templates/deployment.yaml
@@ -56,8 +56,16 @@ spec:
               memory: {{ .Values.resources.requests.memory }}
       volumes:
         - name: config-volume
+        {{- if .Values.configSecret.enabled }}
+          secret:
+            secretName: {{ .Values.configSecret.name }}
+            items:
+            - key: {{ .Values.configSecret.key }}
+              path: config.yml
+        {{- else }}
           configMap:
             name: {{ template "prometheus-ipmi-exporter.fullname" . }}
+        {{- end }}
         {{- with .Values.extraVolumes }}
             {{- toYaml . | nindent 8 }}
         {{- end }}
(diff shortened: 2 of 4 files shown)


charts/prometheus-ipmi-exporter/templates/configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if not .Values.configMapFile -}}
```
zeritti, line 1: The field `configMapFile` does not occur elsewhere in the chart, I reckon its purpose was somehow not implemented and can therefore safely be removed.

charts/prometheus-ipmi-exporter/templates/configmap.yaml
```diff
@@ -1,4 +1,4 @@
...
+{{- if not .Values.useSecret -}}
```
zeritti, line 1: I'd suggest a differently named toggle (more below):
```suggestion
{{- if not .Values.configSecret.enabled -}}
```

charts/prometheus-ipmi-exporter/templates/deployment.yaml
```diff
@@ -56,8 +56,16 @@ spec:
               memory: {{ .Values.resources.requests.memory }}
       volumes:
         - name: config-volume
+        {{- if .Values.useSecret }}
+          secret:
+            secretName: {{ .Values.secret.name }}
+            items:
+            - key: {{ .Values.secret.key }}
+              path: config.yml
+        {{- else }}
```
zeritti, line 65: With respect to the comment above:
```suggestion
        {{- if .Values.configSecret.enabled }}
          secret:
            secretName: {{ .Values.configSecret.name }}
            items:
            - key: {{ .Values.configSecret.key }}
              path: config.yml
        {{- else }}
```

charts/prometheus-ipmi-exporter/values.yaml
```diff
@@ -63,6 +63,11 @@ serviceMonitor:
 
   # Configuration file for ipmi_exporter
 
+useSecret: false
+secret:
+  name: ipmi-secret
+  key: config.yml
```
zeritti, line 69: We should be a bit more specific of the secret. If using a map for a single feature, I'd suggest putting the toggle in the map.

```suggestion
configSecret:
  enabled: false
  name: ipmi-secret
  key: config.yml
```

zeritti (changes requested):
Thank you, @yankcrime, for your PR. Please, see my comments below.

charts/prometheus-ipmi-exporter/templates/configmap.yaml
yankcrime: Thanks, makes sense and so I've amended the PR as you've suggested

zeritti:
Thank you for your changes, @yankcrime. Please, bump the chart's minor version number in Chart.yaml.

yankcrime:
Done, thanks for the help @zeritti 🫡

zeritti (approved):

PR #4547, opened 2024-05-18
[prometheus-mysql-exporter] Set GOMAXPROCS and GOMEMLIMIT environment variables based on container resources

jnoordsij:

Set `GOMAXPROCS` and `GOMEMLIMIT` environment variables based on container resources. This should reduce potential CPU throttling and OOMKills on containers.

The [`resourceFieldRef`](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-resourceFieldRef) is a very specific Kubernetes directive that is created specifically for passing resource-related values, which rounds up the CPU value to the nearest whole number (e.g. 250m to 1) and passes the memory as a numeric value; so `64Mi` would result in the environment variable being set to `67108864`. This by design makes it completely compatible with Go's API.

An example is documented within Kubernetes documentation itself: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-container-fields-as-values-for-environment-variables.

Inspired by https://github.com/traefik/traefik-helm-chart/pull/1029.

N/A

This creates an empty `env` key for those not setting resource values (and not using existing config secret). This is only a little ugly, but should not be harmful. Alternatively, the could be some combined conditional wrapper around the whole `env` block to only make it appear if a value is set, but that will look quite complicated and quite prone to (future) errors.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/Chart.yaml
+++ b/charts/prometheus-mysql-exporter/Chart.yaml
@@ -3,7 +3,7 @@ description: A Helm chart for prometheus mysql exporter with cloudsqlproxy
 name: prometheus-mysql-exporter
 version: 2.5.3
 home: https://github.com/prometheus/mysqld_exporter
-appVersion: v0.15.1
+appVersion: v0.16.0
 sources:
   - https://github.com/prometheus/mysqld_exporter
 maintainers:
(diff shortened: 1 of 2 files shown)


monotek (approved):

monotek (changes requested):
Pleas fix lint

jkroepke:
Looking at https://github.com/prometheus-operator/prometheus-operator/issues/501 - GOMAXPROCS should not set by that way.

monotek (commented):
I would also say adding some more generice itteration over an "extraEnv" value would be the way to go.

desaintmartin:
Is there a case where we do not want this to be applied?

monotek (approved):

PR #4560, opened 2024-05-24
Fix missing default scrapeInterval and evaluationInterval

troytop:
If you pass the default values.yaml to helm install without these two keys set, the secret in prometheus-prometheus-kube-prometheus-prometheus 😞  is set to an empty string, which is invalid and causes it to fail to start.


jkroepke:
dont get this.

```
{{- if .Values.prometheus.prometheusSpec.scrapeInterval }}
  scrapeInterval: {{ .Values.prometheus.prometheusSpec.scrapeInterval }}
{{- end }}
```

scrapeInterval is covered by an if condition. If scrapeInterval is empty, the if condition is false and the block is skipped

jkroepke:
Empty string is false, too.

If and default share the same
logic. There is no known condition where the if return true, but the default filter returns the default value.

QuentinBisson:
This had been inactive for almost a year, closing

PR #4561, opened 2024-05-24
[kube-prometheus-stack] Correctly set version to alertManagerSpec.version

EoinFarrell:

In the Prometheus CRD, the version field can be set using prometheusSpec.version and if not specified it falls back to prometheusSpec.image.tag ([source code link](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L57)). This allows the image tag to differ from the exact version of Prometheus run. AlertManager CRD currently does not support such decoupling.

This causes a problem when a custom AlertManager image is used (based off standard Prometheus AlertManager image but with some internal modifications) is used, and its tag schema does not follow the schema of the AlertManager image. The Prometheus-Operator container that syncs the StatefulSet state of the AlertManager CRD has a check to ensure the AlertManager code version is valid ([source code link](https://github.com/prometheus-operator/prometheus-operator/blob/6a5aa14000b74bd7dd0ce7664462cff83e0b0439/pkg/alertmanager/statefulset.go#L429-L462)). But without this decoupling, it is impossible to distinguish between the custom image tag and the actual base version of the AlertManager running. Example:

- Current Prometheus AlertManager version: v0.27.0
- Custom built image tag: 2024-05-24

Setting prometheusSpec.image.tag to 2024-05-24 which is required to pick up the image leads to Prometheus-Operator using this version for its reconcilliation, but as its not in the format v0.xx.x it fails with a message stating "unsupported Alertmanager major version". This PR adds the ability to decouple these two values, allowing AlertManager to correctly reconcile the version while using a custom image tag.
#### Checklist

- [x] Chart Version bumped


jkroepke (approved):
Same logic is used at Prometheus

https://github.com/prometheus-community/helm-charts/blob/f9dc2dd66657e678b665c08a3be67cf1011f7362/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L57

LGTM

jkroepke (approved):

jkroepke:
Thanks for the PR

PR #4562, opened 2024-05-24
[kube-state-metrics] separate ServiceMonitor values so they can be independently configured

kbweave:
This PR allows the two ServiceMonitor endpoints `http` and `metrics` to be separately configured. Relabelings that need to be done on one endpoint likely don't need to happen on the other.
The change is structured so it will be backwards compatible with existing values.

#### Checklist

- [x] Chart Version bumped


dotdc (approved):

dotdc:
Thank you for your contribution @kbweave !

PR #4570, opened 2024-05-27
[kube-prometheus-stack] introduce PVC claim retention field in prometheus operator

Woojciech:
This PR exposes persistentVolumeClaimRetentionPolicy field of Prometheus / PrometheusAgent, a CRD resource from Prometheus operator. PersistentVolumeClaimRetentionPolicy is a feature introduced for StatefulSets (by default in K8s 1.27.0+), this feature lets us decide whether the StatefulSet should delete or retain it's corresponding PVC's it is deleted or scaled. It is particularly useful in environments where stale volumes can grow in amount, causing additional costs or clutter in general. The CRD field itself was added to Prometheus operator around October last year, however not yet exposed in kube-prometheus-stack.

Change provided in this PR is based on the work already present in the [prometheus chart](https://github.com/prometheus-community/helm-charts/tree/c43f4b2d3351122dd0c4aa001dcf25dcffe84a02/charts/prometheus) , please see the special notes section.

This PR is based on the previous work which was made in the prometheus chart, as per links provided below. I wanted to add this exact feature into the chart and I found that it was already done in the linked prometheus operator chart.

Please let me know if relying on the previous implementation has any legal implications (licensing etc.) - if I should for example add authors of mentioned changes as co-authors or provide credits for their work in any other way.

* https://github.com/prometheus-community/helm-charts/blob/c43f4b2d3351122dd0c4aa001dcf25dcffe84a02/charts/prometheus/templates/sts.yaml#L17
* https://github.com/prometheus-community/helm-charts/blob/c43f4b2d3351122dd0c4aa001dcf25dcffe84a02/charts/prometheus/values.yaml#L588

#### Checklist

- [X] Chart Version bumped


charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -261,6 +261,11 @@ spec:
   storage:
 {{ tpl (toYaml .Values.prometheus.prometheusSpec.storageSpec | indent 4) . }}
 {{- end }}
+{{- if semverCompare ">= 1.27.0" (include "kube-prometheus-stack.operator.kubeVersion" .) }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: {{ ternary "Delete" "Retain" .Values.prometheusOperator.statefulSet.pvcDeleteOnStsDelete }}
+    whenScaled: {{ ternary "Delete" "Retain" .Values.prometheusOperator.statefulSet.pvcDeleteOnStsScale }}
+{{- end }}
```
jkroepke, line 268: What about

```suggestion
{{- with .Values.prometheus.prometheusSpec }}
  persistentVolumeClaimRetentionPolicy:
    {{- toYaml .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector | nindent 4 }}
{{- end }}
```

and a empty `scrapeConfigNamespaceSelector: {}` by default? That should also works out of the box, doesn't require the version check.

It also keeps the existing name schema.

jkroepke: Hi @Woojciech 

Sorry, I did a mistake and correct by suggestion. I hope question 1 and 2 is now more clear. A user could put:

```yaml
prometheus:
  prometheusSpec:
    persistentVolumeClaimRetentionPolicy:
      whenDeleted: Delete
      whenScaled: Retain
```

into the values to change the defaults.

> Could you please clarify on how do we know that in this case version is 1.27.0+?

In the default values, this should be appear:

```yaml
prometheus:
  prometheusSpec:
    persistentVolumeClaimRetentionPolicy: {}
```

which means, empty by default and the whole block might be not set in the values.

Additionally,. I'm shifting the condition to the end-user. The end-user should only define the flags, if he is aware that he is using 1.27. 

Your great documentation inside the values yaml also mention that 1.27 is required. I do not see the additional benefit of adding an technical condition. I

Woojciech: No problem, thanks for such detailed response :) I implemented your suggestions, I left whenDeleted and whenScaled fields in YAML commented out for reference, both set to default value which is Retain, let me know if that is fine, or if I should remove it at all.

PS. the nice docs are not necessarily mine, they were copied from the prometheus chart, so kudos to prometheus maintainers :)

jkroepke (dismissed):

Woojciech:
fixed trailing spaces, hopefully everything will be well formatted this time :)

Woojciech:
Hi @jkroepke can I please ask you for re-approval, as I fixed all the required checks?

jkroepke (approved):

PR #4597, opened 2024-06-07
[prometheus-node-exporter] permit to customize service clusterIP

nerzhul:
It permits to disable clusterIP which is not required to fetch metrics. On cluster with very huge node number always moving, it create a loadbalancer on each node using much memory and CPU due to watches from the service proxy

Permit to customize node-exporter ClusterIP field, permitting people to disable this ClusterIP which is always updating its backend when nodes are added created, generating much work on each {service,kube}-proxy pod on the cluster

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/service.yaml
+++ b/charts/prometheus-node-exporter/templates/service.yaml
@@ -19,6 +19,9 @@ spec:
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
 {{- end }}
   type: {{ .Values.service.type }}
+{{- if and (eq .Values.service.type "ClusterIP") .Values.service.clusterIP }}
+  clusterIP: "{{ .Values.service.clusterIP }}"
+{{- end }}
   ports:
     - port: {{ .Values.service.port }}
       {{- if ( and (eq .Values.service.type "NodePort" ) (not (empty .Values.service.nodePort)) ) }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-node-exporter/templates/service.yaml
```diff
@@ -19,6 +19,9 @@ spec:
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
 {{- end }}
   type: {{ .Values.service.type }}
+{{- if eq .Values.service.type "ClusterIP" }}
+  clusterIP: "{{ .Values.service.clusterIP }}"
+{{- end }}
```
zeritti, line 24: I'd suggest to consider only user supplied (non-null) values. Otherwise, we should leave the field's value unspecified.

```suggestion
{{- if and (eq .Values.service.type "ClusterIP") .Values.service.clusterIP }}
  clusterIP: "{{ .Values.service.clusterIP }}"
{{- end }}
```

zeritti (changes requested):
Thank you, @nerzhul, for your PR. Providing users with an option to set clusterIP is not uncommon, I reckon we can support the new field in the chart.

nerzhul:
rebased, squashed & signed

nerzhul:
@zeritti thanks for your comments, all is fixed :)

zanhsieh (dismissed):

PR #4614, opened 2024-06-19
[prometheus-postgres-exporter] Correct endpoint port name in service monitor

ruslansavchuk:
**Chart:** prometheus-postgres-exporter

**Issue**
When setting the service name to a custom value and enabling the service monitor, the Prometheus operator is unable to discover any pods for monitoring.

**Reason**
The specified service name is configured as a "targetPort" in the ServiceMonitor. 
According to the [documentation](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.Endpoint), the target port property refers to the port (or it name) in the pod, but we set service port name as the value of the target port. 
When default service name is used (http) everything works as expected because the default service port name is the same as the port name in the pod.


zeritti (approved):
Thank you, @ruslansavchuk, for your PR, LGTM.

PR #4633, opened 2024-06-26
[prometheus] add prometheus annotations to kube-state-metrics and prometheus-node-…

### LINKED ISSUE #4632: [prometheus] prometheus-node-exporter and kube-state-metrics sub-charts are useless if prometheus annotations are not set

The reason why a chart user would choose to enable `prometheus-node-exporter` and `kube-state-metrics`  sub-charts is to collect their metrics. However, without the prometheus annotations `prometheus.io/scrape` and `prometheus.io/port`. 

### What's your helm version?

v3.14.4

v1.28.1

### Which chart?

prometheus

### What's the chart version?

25.21.0

### What happened?

When installing the prometheus chart with `kube-state-metrics.enabled=true` and `prometheus-node-exporter=true` (these are the default values), one can notice that prometheus does not scrape metrics from those. 

I'd expect that prometheus starts scraping automatically from `kube-state-metrics` and `prometheus-node-exporter`

### How to reproduce it?

installing the prometheus chart with `kube-state-metrics.enabled=true` and `prometheus-node-exporter=true`

Adding pod annotations to both 
```
kube-state-metrics:
  enabled: true
  podAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "8080"
prometheus-node-exporter:
  enabled: true
  podAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "9100"
```

helm install prometheus prometheus-community/prometheus 

_No response_

zak905:
…exporter pods

This is a simple improvement that makes prometheus automatically scrap metrics from kube-state-metrics and node-exporter in case they are enabled. 

- fixes #4632 

#### Checklist

- [x ] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [ x] Chart Version bumped
- [x ] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)


zeritti (commented):
Thank you, @zak905, for your PR. The proposed change is unnecessary, though, as metrics endpoints of both kube-state-metrics and node exporter get discovered through role _endpoints_ (scrape config _kubernetes-service-endpoins_), i.e. through their services which already are annotated.

zak905:
Indeed. - as I mentionned in the issue, I have forgot to mention that I have deleted `kubernetes-service-endpoints`, as I wanted to keep the discovery for pods only. This is the reason why metrics were not showing up. Thanks.

Issue #4622, opened 2024-06-22
[kube-prometheus-stack] Add Gateway API templates


e3b0c442:
### Is your feature request related to a problem ?

I need to manually add HTTPRoutes when trying to use Gateway API for ingress

### Describe the solution you'd like.

Now that Gateway API has reached v1 and implementations are beginning to land, add Gateway API templates (HTTPRoute) that can be chosen in lieu of Ingress resources.

### Describe alternatives you've considered.

* Disabling ingress and adding Gateway API resources manually (current option)
* Continue using Ingress

### Additional context.

Using kube-prometheus-stack as a filler here, I understand the ingress is a part of the subordinate charts.

jkroepke:
@monotek @GMartinez-Sisti @QuentinBisson I think, adding Gateway API feels natural, since ingress API reached the feature complete state and may not receive new features.

Do you already have a good template for HTTPRoutes? Should we support Gateway as well? Do you have Gateway API already in use on your company and already develop a good base?

With some coordination, we can deliver the same experience for all end-users across all prometheus community helm charts.

GMartinez-Sisti:
> @monotek @GMartinez-Sisti @QuentinBisson I think, adding Gateway API feels natural, since ingress API reached the feature complete state and may not receive new features.
> 
> Do you already have a good template for HTTPRoutes? Should we support Gateway or BackendTLSPolicy as well? Do you have Gateway API already in use on your company and already develop a good base?
> 
> With some coordination, we can deliver the same experience for all end-users across all prometheus community helm charts.

I think we should support both for now, clusters that were created some time ago and are still being used (even if they are on the latest k8s version) are very likely to still use ingress controllers that might not support this.

jkroepke:
Definitive! I was not talking about removing ingress.

jkroepke:
An initial draft can be found here: https://github.com/prometheus-community/helm-charts/pull/4646

Some Feedback would be appreciated. @e3b0c442 I would like to know, if that would meets your expectation.

QuentinBisson:
I Can take a look but i'm not familiar with the gateway api yet. Also, is it enabled for the grafana depencency? If you did not we might want to make the gateway api enable flag global as well?

jkroepke:
No, but you can setup an PR which I can review/approve at grafana 😅

jkroepke:
[...]

PR #4646, opened 2024-06-28
[kube-prometheus-stack] Implement Gateway API

jkroepke:

The current template is just an approach. Once finalized, we add it anywhere in the chart.

Ref: #4622 

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -28,6 +28,31 @@ alertmanager:
       logLevel: {{ print "debug" | quote }}
 
 prometheus:
+  gatewayAPI:
+    httpRoute:
+      enabled: true
+      annotations:
+        key: value
+      labels:
+        key: value
+
+      hostnames:
+      - my-filter.example.com
+      parentRefs:
+      - name: acme-gw
+
+      matches:
+        - path:
+            type: PathPrefix
+            value: /
+
+      filters:
+        - type: RequestHeaderModifier
+          requestHeaderModifier:
+            set:
+              - name: my-header-name
+                value: my-new-header-value
+
   prometheusSpec:
     additionalConfig:
       logFormat: json
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -28,31 +28,6 @@ alertmanager:
       logLevel: {{ print "debug" | quote }}
 
 prometheus:
-  gatewayAPI:
-    httpRoute:
-      enabled: true
-      annotations:
-        key: value
-      labels:
-        key: value
-
-      hostnames:
-      - my-filter.example.com
-      parentRefs:
-      - name: acme-gw
-
-      matches:
-        - path:
-            type: PathPrefix
-            value: /
-
-      filters:
-        - type: RequestHeaderModifier
-          requestHeaderModifier:
-            set:
-              - name: my-header-name
-                value: my-new-header-value
-
   prometheusSpec:
     additionalConfig:
       logFormat: json
--- a/charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
+++ b/charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
@@ -1,0 +1,17 @@
+alertmanager:
+  ingress:
+    enabled: true
+    hosts:
+      - "*.test.com"
+
+
+prometheus:
+  ingress:
+    enabled: true
+    hosts:
+      - "*.test.com"
+  route:
+    main:
+      enabled: true
+      hostnames:
+        - "*.test.com"
--- a/charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
+++ b/charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
@@ -2,16 +2,35 @@ alertmanager:
   ingress:
     enabled: true
     hosts:
-      - "*.test.com"
+      - "*.example.com"
 
 
 prometheus:
   ingress:
     enabled: true
     hosts:
-      - "*.test.com"
+      - "*.example.com"
   route:
     main:
       enabled: true
       hostnames:
-        - "*.test.com"
+        - "*.example.com"
+      filters:
+        - type: RequestHeaderModifier
+          requestHeaderModifier:
+            set:
+              - name: my-header-name
+                value: my-new-header-value
+      additionalRules:
+        - hostnames:
+            - "foo.example.com"
+          filters:
+            - type: RequestHeaderModifier
+              requestHeaderModifier:
+                set:
+                  - name: my-header-name
+                    value: my-new-header-value
+          matches:
+            - path:
+                type: PathPrefix
+                value: /foo/
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -3303,6 +3303,42 @@ prometheus:
       #   hosts:
       #     - prometheus.example.com
 
+  # -- BETA: Configure the gateway routes for the chart here.
+  # More routes can be added by adding a dictionary key like the 'main' route.
+  # Be aware that this is an early beta of this feature,
+  # kube-prometheus-stack does not guarantee this works and is subject to change.
+  # Being BETA this can/will change in the future without notice, do not use unless you want to take that risk
+  # [[ref]](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1alpha2)
+  route:
+    main:
+      # -- Enables or disables the route
+      enabled: false
+
+      # -- Set the route apiVersion, e.g. gateway.networking.k8s.io/v1 or gateway.networking.k8s.io/v1alpha2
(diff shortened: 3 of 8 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -3251,6 +3251,25 @@ prometheus:
       #   hosts:
       #     - prometheus.example.com
 
+  gateway:
```
QuentinBisson, line 3254: Do you have other examples of this kind of flag upstream? Because gateway is used for a lot of other things right now liké Mimir gateways so maybe other upstream projects use better names liké gatewayapi?  It could also be they use `gateway`😅

jkroepke: No. I found nothing via GitHub Code search. Bitnami and Grafana did not have Gateway API yet, but I'm open for other names here.

But gatewayApi sounds more specific.

rouke-broersma: I would propose to skip the gateway level and use the httpRoute level only, this makes sense with gateway api terminology where the gateway resource is managed centrally by the cluster operator and the application developer only deals with routes. If the extra level is needed because you envision that tlsRoute etc might also at some point be supported I would propose the following structure:

```yaml
routes:
  http:
  tls:
  grpc:

etc etc
```

jkroepke: Still prefix the values with `gatewayAPI` - who knows what comes? 

Currently, only httpRoute is implemented and there are no plans to provide more

matzarah:
It would be great to see the "ingressPerReplica" feature replicated for this too as we are currently transitioning from Ingress to to Gateway API and have a need to expose each replica externally.

QuentinBisson:
Hey @jkroepke I asked for some expects feedback on this and this looks quite good.

As long as the tests are fixed, I'm fine with merging this :)

charts/kube-prometheus-stack/values.yaml
rouke-broersma: Sorry for not getting back to you. I still prefer `routes` as the terminology as this is how the concept is called in gateway api: https://gateway-api.sigs.k8s.io/concepts/security-model/#resources

But no hard feelings over calling it something else, I understand that since it's specific to gateway api you may want to scope the name better so there are no misunderstandings.

QuentinBisson: I would be fine with routes as well but we need to be careful about openshift users

rouke-broersma: That is a good point, could those perhaps be merged or are they not similar enough? Conceptually it would be nice if openshift routes could simply be treated as a different route type instead of it's own separate construct but I am not familiar enough with openshift to say.

I did a quick search and it seems like the authors of the helm chart library `common` at least have decided on `route` for now:

Both in truecharts and in the upstream version by bjw-s:

https://github.com/truecharts/public/blob/master/charts/library/common/values.yaml#L684
https://github.com/bjw-s/helm-charts/blob/main/charts/library/common/values.yaml#L653

And I also find it in other places:

https://github.com/harshaputtaswamy/oom/blob/master/kubernetes/common/common/templates/_ingress.tpl#L332

QuentinBisson: Then maybe we should follow along yes with that :) @jkroepke what do you think?

jkroepke: I like some ideas and I did not had v1alpha2 in scope.

@QuentinBisson and @rouke-broersma please take another look of that example.


```
helm template . -f ci/05-ingress-and-gateway-routes-values.yaml -s templates/prometheus/route.yaml
---
# Source: kube-prometheus-stack/templates/prometheus/route.yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: release-name-kube-promethe-prometheus
  namespace: default
  labels:
    app: kube-prometheus-stack-prometheus
    
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "65.5.1"
    app.kubernetes.io/part-of: kube-prometheus-stack
    chart: kube-prometheus-stack-65.5.1
    release: "release-name"
    heritage: "Helm"
spec:
  hostnames:
    - '*.test.com'
  rules:
    - backendRefs:
        - name: release-name-kube-promethe-prometheus
          port: 9090
      matches:
        - path:
            type: PathPrefix
            value: /

```

charts/kube-prometheus-stack/values.yaml
```diff
@@ -3303,6 +3303,42 @@ prometheus:
       #   hosts:
       #     - prometheus.example.com
 
+  # -- BETA: Configure the gateway routes for the chart here.
+  # More routes can be added by adding a dictionary key like the 'main' route.
+  # Be aware that this is an early beta of this feature,
+  # kube-prometheus-stack does not guarantee this works and is subject to change.
+  # Being BETA this can/will change in the future without notice, do not use unless you want to take that risk
```
jkroepke, line 3310: @QuentinBisson what did you think about that? We I understand this as long as this is BETA, we are allowed todo breaking changes without bump the major.

WDYT?

QuentinBisson: Looks good :)

charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+
+prometheus:
+  ingress:
+    enabled: true
+    hosts:
+      - "*.test.com"
+  route:
+    main:
+      enabled: true
```
QuentinBisson, line 15: Does it make sense to have both ingress and gateway enabled?

QuentinBisson: Would that not cause issues?

charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
```diff
@@ -28,6 +28,31 @@ alertmanager:
       logLevel: {{ print "debug" | quote }}
 
 prometheus:
+  gatewayAPI:
```
QuentinBisson, line 31: I think this values file does not match the new syntax right?

charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
jkroepke: Sure! For each value test file. helm-chart-test will do a full helm install. Increase the amount if ci/values.yaml would increase the CI times as well.

To avoid that, I merged distinct features into one values.yaml. It might be pointless to enabled ingress and routes, but I would like to have to feature under test coverage.

charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
jkroepke: Thanks!

charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
QuentinBisson: No worries that works for me :)

QuentinBisson (approved):

charts/kube-prometheus-stack/values.yaml
rouke-broersma: I am not a fan of the common chart and do not use them myself! It was just an example of what I found in the wild.

I like the current proposal, and I think it exposes sufficient fields for an initial version. Other fields can be added as needed when we get more experience with gateway api. 

I am not sure route kind is even necessary for prometheus, I think only http route can be valid? That said it doesn't hurt either and could provide easy extension later if a different route type becomes relevant to prometheus so I'm also fine with leaving it in.

jkroepke:
I looked into the per-replica route, but It's a bit too complex for now. I would suggest doing that in a different PR by an user which has more time.

jkroepke:
I also need an approval from @GMartinez-Sisti Since this PR contains modifications to the CI job.

GMartinez-Sisti (approved):

jkroepke:
Follow-up tasks:

- HTTP Route for Alertmanager
- HTTP Route for Alertmanager per Replica
- HTTP Route for Prometheus per Replica
- HTTP Route for Thanos Ruler

charts/kube-prometheus-stack/values.yaml
```diff
@@ -3303,6 +3303,42 @@ prometheus:
...
+  # [[ref]](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1alpha2)
+  route:
+    main:
+      # -- Enables or disables the route
+      enabled: false
+
+      # -- Set the route apiVersion, e.g. gateway.networking.k8s.io/v1 or gateway.networking.k8s.io/v1alpha2
+      apiVersion: gateway.networking.k8s.io/v1
```
z0rc, line 3318: I'm not sure this should be configurable. `v1alpha2` is already deprecated, IMHO there is there is value in supporting `v1` only.

rouke-broersma: Gateway api is too new to enforce only v1 api versions imo, there may be many implementations lacking behind, which would mean those users are unable to use this feature in the chart. There are also some update issues iirc, with regards to experimental versions of v1 api, which might cause users to still be on alpha api versions.

PR #4707, opened 2024-07-09
[kube-prometheus-stack] Add condition to Alert Manager default dashboard to prevent adding the dashboard when AM is disabled

Phenix66:

When AM is disabled, the default Grafana dashboard for AM is still added. Other services such as the Kube proxy, etcd, etc do not add their dashboards when disabled.

- fixes #

#### Checklist

- [x] Chart Version bumped


GMartinez-Sisti:
My first thought is that there might be folks deploying this stack without alertmanager enabled but still wanting to get the dashboard for whatever reason. 🤔

Phenix66:
> My first thought is that there might be folks deploying this stack without alertmanager enabled but still wanting to get the dashboard for whatever reason. 🤔

I had a similar thought and debated implementing a new value similar to `.Values.defaultRules` but I decided against that since there were already other dashboards that simply relied upon the relevant `enabled` value.

GMartinez-Sisti:
> > My first thought is that there might be folks deploying this stack without alertmanager enabled but still wanting to get the dashboard for whatever reason. 🤔
> 
> I had a similar thought and debated implementing a new value similar to `.Values.defaultRules` but I decided against that since there were already other dashboards that simply relied upon the relevant `enabled` value.

Fair enough. However this change needs to be done in a different way. please check the header of the file you changed.

Phenix66:
> > > My first thought is that there might be folks deploying this stack without alertmanager enabled but still wanting to get the dashboard for whatever reason. 🤔
> > 
> > 
> > I had a similar thought and debated implementing a new value similar to `.Values.defaultRules` but I decided against that since there were already other dashboards that simply relied upon the relevant `enabled` value.
> 
> Fair enough. However this change needs to be done in a different way. please check the header of the file you changed.

Updated to use the sync script

GMartinez-Sisti:
I'd like to get a second opinion on this, the change might be considered a breaking change because something that might be in use is removed use once this is merged. @jkroepke @QuentinBisson pls 🙏

jkroepke:
In general, I would favorite a more generic approach where individual dashboards can be toggled.

Phenix66:
> In general, I would favorite a more generic approach where individual dashboards can be toggled.

Added an individual toggle identical to the one used for node exporter to do the same thing.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 67.5.0
+version: 67.6.0
```
jkroepke, line 26:
```suggestion
version: 68.1.0
```

jkroepke (approved):
LGTM

PR #4736, opened 2024-07-24
[prometheus-node-exporter] prevent node exporter from being scheduled on fargate or virtual nodes

### LINKED ISSUE #4735: [prometheus-node-exporter] Prevent node exporter from being scheduled on fargate

When running on EKS clusters with a mix of fargate and EC2 nodes the node exporter pods can end up unschedulable and stuck in a pending state. 

Add a  `nodeAffinity` to the node-exporter by default

```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: eks.amazonaws.com/compute-type
            operator: NotIn
            values:
            - fargate
 ```

None

### Additional context.

There have been many previous Issues about this:

https://github.com/prometheus-community/helm-charts/issues/2876
https://github.com/prometheus-community/helm-charts/issues/2908
https://github.com/prometheus-community/helm-charts/issues/925

Grafanas [k8s-monitoring charts](https://github.com/grafana/k8s-monitoring-helm) have also added a node affinity to the node-exporter by default which prevents it from running on fargate - see https://github.com/grafana/k8s-monitoring-helm/pull/362. 
I think it would be useful to do the same here. 

If my understanding is correct, then for people that have already added this `nodeAffinity` this would not be a breaking change, it would simply be redundant - but this I am not completely sure about.

### PR DESCRIPTION
**Pull request by mariuskimmina**

This PR prevents node-exporter from being scheduled on fargate

- fixes #4735

If my understanding is correct, then for people that have already added this nodeAffinity this would not be a breaking change, it would simply be redundant - but this I am not completely sure about.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/templates/_helpers.tpl
+++ b/charts/prometheus-node-exporter/templates/_helpers.tpl
@@ -200,3 +200,38 @@ labelValueLengthLimit: {{ . }}
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{/*
+The default node affinity to exclude 
+- AWS Fargate 
+- Azure virtual nodes
+*/}}
+{{- define "prometheus-node-exporter.defaultAffinity" -}}
+nodeAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+    nodeSelectorTerms:
+    - matchExpressions:
+      - key: eks.amazonaws.com/compute-type
+        operator: NotIn
+        values:
+        - fargate
+      - key: type
+        operator: NotIn
+        values:
+        - virtual-kubelet
+{{- end -}}
+{{- define "prometheus-node-exporter.mergedAffinities" -}}
+{{- $defaultAffinity := include "prometheus-node-exporter.defaultAffinity" . | fromYaml -}}
+{{- with .Values.affinity -}}
+  {{- if .nodeAffinity -}}
+    {{- $_ := set $defaultAffinity "nodeAffinity" (mergeOverwrite $defaultAffinity.nodeAffinity .nodeAffinity) -}}
+  {{- end -}}
+  {{- if .podAffinity -}}
+    {{- $_ := set $defaultAffinity "podAffinity" .podAffinity -}}
+  {{- end -}}
+  {{- if .podAntiAffinity -}}
+    {{- $_ := set $defaultAffinity "podAntiAffinity" .podAntiAffinity -}}
+  {{- end -}}
+{{- end -}}
+{{- toYaml $defaultAffinity -}}
+{{- end -}}
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.42.1
+version: 4.43.0
 appVersion: 1.8.2
 home: https://github.com/prometheus/node_exporter/
 sources:
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.42.0
+version: 4.42.1
 appVersion: 1.8.2
 home: https://github.com/prometheus/node_exporter/
 sources:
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by zeritti — CHANGES_REQUESTED** [review]
Thank you, @mariuskimmina, for your PR. That would be a useful change for users of mixed EKS clusters. Similarly in AKS with virtual nodes (`type: virtual-kubelet`) which could also be supported in your change.

Now, `affinity` is a user settable field and may already include `nodeAffinity`. In your current implementation, the field would eventually get overwritten by the user's `nodeAffinity`. Both user's configuration and the template's field have to be merged. I'd suggest using a helper template for the purpose.

The other option is inserting the field directly in the values file's  `affinity`. You probably meant to do this. Personally, I'd prefer doing it as described above, though.

**Comment by mariuskimmina** [comment]
Hey @zeritti, thanks for taking the time to look at it. 

I updated the PR with a template approach now. 
So far I've only tested it using `helm template prometheus-node-exporter ./prometheus-node-exporter` so the default value is working fine, I haven't tested the merging with user defined nodeAffinity yet.

**Comment by mariuskimmina** [comment]
I found the Azure one you were talking about, my only concern here is that `key: type` seems so generic. If you want to I am fine with adding this as a default as well.

```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
            - key: type
              operator: NotIn
              values:
                - virtual-kubelet
 ```

mariuskimmina:
Added the Azure version and tested a few more cases, let me know what you think @zeritti

charts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -200,3 +200,36 @@ labelValueLengthLimit: {{ . }}
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{/*
+The default node affinity to exclude Fargate / Virtual nodes
```
zanhsieh, line 205:
```suggestion
The default node affinity to exclude
- AWS Fargate
- Azure Virtual nodes
```

zanhsieh (changes requested):

charts/prometheus-node-exporter/templates/_helpers.tpl
mariuskimmina: Thanks for the suggestion, adapted your changes

zanhsieh (dismissed):

monotek (dismissed):
i don't see why is this needed if there is already a generic solution available.

mariuskimmina:
> i don't see why is this needed if ther is already a genric solution available.

It's a common problem for people that run EKS with a mixture of Fargate and EC2 nodes, as seen by the number of opened issues around this in the past. I think this would be a sane default configuration to provide that can safe some people the trouble of having to search for this and then add it themselves.

zanhsieh (dismissed):

mariuskimmina:
Any more thought on this @monotek @zeritti @zanhsieh ?

I think adding this as default value would save quite a few people the time of having to search for this and add it themselves. I think if you are using a mixture of EC2 nodes and Fargate you'll always want to have this `nodeAffinity` and if you are not using Fargate it shouldn't have any negative impact either.

mariuskimmina:
Any update @monotek @zeritti @zanhsieh ?
Since the issue https://github.com/prometheus-community/helm-charts/issues/4735 got some support I wanted to bring this topic up again.

> Any more thought on this @monotek @zeritti @zanhsieh ?
> 
> I think adding this as default value would save quite a few people the time of having to search for this and add it themselves. I think if you are using a mixture of EC2 nodes and Fargate you'll always want to have this `nodeAffinity` and if you are not using Fargate it shouldn't have any negative impact either.

zanhsieh (dismissed):

monotek:
I let this decisson to the maintainers of the chart...

BobDu:
Given the extensive use of EKS and AKS, offering more out-of-the-box default configurations would be a better option. 

At least theoretically, it wouldn't cause any inconvenience to users who do not use these two services, right?

mariuskimmina:
Any update on this @zeritti @zanhsieh ?

Just rebased and bumped the chart version again. 
So far it seems as time goes on only more people are in favor of this.

zeritti (changes requested):
Since there is no indication of the default hard node affinity in the values file, please, enter a comment at the `affinity` field saying that the default node affinity excludes Fargate nodes and virtual kubelets from scheduling unless overriden by hard node affinity set in the field.

mariuskimmina:
Thanks again for reviewing @zeritti  
I added the requested comment to the affinity field add bumped the minor version instead

zeritti (approved):
LGTM

PR #4749, opened 2024-07-27
[kube-prometheus-stack] Fix mixins generation and fix update on mac

jkroepke:

https://github.com/kubernetes/kube-state-metrics/pull/2388 introduced a syntax error. The syntax error has an effect to our pipeline.

I take a look closer into this, since we are using fixed shasums for dependencies but somewhere else, floating version are used.

It turns out that the lockfile for Jsonnet dependencies in [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnetfile.lock.json) is located in the repository root. To respect the lockfile, I had to change the update mixin configuration to run the generation from the repository root.

I also fixed the update script for macOS.

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/hack/update_mixins.sh
+++ b/charts/kube-prometheus-stack/hack/update_mixins.sh
@@ -2,6 +2,14 @@
 
 set -euo pipefail
 
+sedi () {
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    sed -i '' "$@"
+  else
+    sed -i "$@"
+  fi
+}
+
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 trap 'rm -rf "${SCRIPT_DIR}/tmp"' EXIT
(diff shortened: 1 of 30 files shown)


charts/kube-prometheus-stack/hack/update_mixins.sh
```diff
@@ -2,6 +2,14 @@
 
 set -euo pipefail
 
+sedi () {
```
GMartinez-Sisti, line 5: This check is not bullet proof. Some folks, like me 😊, install the GNU version of sed on macOS and set it as the default one.

I would do something like this instead to detect the correct version:
```sh
if [[ "$(sed --version 2>/dev/null)" =~ ^.*GNU.*$ ]]; then
```

jkroepke: I added the logic from my helm-secrets project to cover all known cases. 

And I put the condition outside of the function, because than it needs to runs once only.

GMartinez-Sisti (approved):

PR #4757, opened 2024-07-30
[prometheus-elasticsearch-exporter] Add ability to customize created serviceaccount name

creiche:

values.serviceAccount.name does not have an effect when `create: true`.  It should customize the created serviceAccount name.

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/Chart.yaml
+++ b/charts/prometheus-elasticsearch-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 6.1.0
+version: 6.2.0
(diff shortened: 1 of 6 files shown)


zeritti (changes requested):
Thank you, @creiche, for your PR. Will you make a test case in the `ci` directory creating a service account? 

Actually, creating an SA of the default templated name requires setting two fields which is a bit awkward and unusual: one has to set `serviceAccount.create` and unset `serviceAccount.name`. I'd suggest to set the `serviceAccount.name` field in the values to an empty string.

creiche:
thanks @zeritti I added a test case.

zeritti:
Thank you, @creiche. The CI values file names have to match `*-values.yaml`, e.g. `serviceaccount-values.yaml`, otherwise, they won't be picked up.

zeritti (approved):
Thank you, @creiche, LGTM.

PR #4779, opened 2024-08-08
[prometheus-elasticsearch-exporter] Add tpl to nodeSelector and tolerations

denzhel:
We're moving our infra charts from Terraform to ArgoCD/ApplicationSets and we need the ability to template
some values with an environment name. In this case, we need to asign some tolerations and nodeSelector for Karpernter.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/Chart.yaml
+++ b/charts/prometheus-elasticsearch-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 7.2.0
+version: 6.3.0
 kubeVersion: ">=1.19.0-0"
 appVersion: "v1.7.0"
 home: https://github.com/prometheus-community/elasticsearch_exporter
--- a/charts/prometheus-elasticsearch-exporter/Chart.yaml
+++ b/charts/prometheus-elasticsearch-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 6.2.0
+version: 7.2.0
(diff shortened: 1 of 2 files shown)


charts/prometheus-elasticsearch-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Elasticsearch stats exporter for Prometheus
 name: prometheus-elasticsearch-exporter
-version: 6.2.0
+version: 7.2.0
```
zeritti, line 4: A minor change will do.
```suggestion
version: 6.3.0
```

charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
```diff
@@ -184,11 +184,11 @@ spec:
       {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
-        {{- toYaml . | nindent 8 }}
+        {{- tpl (toYaml . | nindent 8) $ }}
```
zeritti, line 187: I'd suggest to indent the block after applying the function on the value: 

```suggestion
        {{- tpl (toYaml .) $ | nindent 8 }}
```

charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
```diff
@@ -184,11 +184,11 @@ spec:
...
       {{- end }}
       {{- with .Values.tolerations }}
       tolerations:
-        {{- toYaml . | nindent 8 }}
+        {{- tpl (toYaml . | nindent 8) $ }}
```
zeritti, line 191:
```suggestion
        {{- tpl (toYaml .) $ | nindent 8 }}
```

zeritti (changes requested):
Thank you, @denzhel, for your PR. Please, see my comments below.

PR #4818, opened 2024-08-26

schnatterer:

```
failed to create typed patch object (..): .spec.scrapeConfigNamespaceSelector: field not declared in schema
```

```
prometheus:
  prometheusSpec:
    scrapeConfigNamespaceSelector: null
    scrapeConfigSelectorNilUsesHelmValues: null
    scrapeConfigSelector: null
```

At runtime, the operator yields some warnings about the missing fields, but apparently everything else works.

#### Checklist

- [x] Chart Version bumped


jkroepke:
Hi @schnatterer 

in general, I understand the use-case and I'm happy to assist here.

However, I also like constancy. In conclusion, all selectors should gain this capability.

@GMartinez-Sisti What did you think here? The selector having some hacks here. If you would ask me, I would favorite a cleanup.

Instead:

```
{{- if .Values.prometheus.prometheusSpec.scrapeConfigSelector }}
  scrapeConfigSelector:
{{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigSelector | indent 4) . }}
{{ else if .Values.prometheus.prometheusSpec.scrapeConfigSelectorNilUsesHelmValues  }}
  scrapeConfigSelector:
    matchLabels:
{{ else if not (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil) }}
  scrapeConfigSelector: {}
{{- end }}
```

I would replace the whole logic to:

```
{{- if not (or (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil) (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector "")) }}
  scrapeConfigSelector: {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigSelector | indent 4) . }}
{{- end }}
```

and I would like to move the default value to the helm values file, since we support `tpl`.

```yaml
  scrapeConfigSelector:
    matchLabels:
      release: '{{ $.Release.Name | quote }}'
```

In conclusion `scrapeConfigSelectorNilUsesHelmValues` could be removed

If `scrapeConfigSelector` is set to `{}`, it will be propagated to the template.

GMartinez-Sisti:
@jkroepke I like the idea, it does make it cleaner!

Regarding:

> SelectorNilUsesHelmValues could be removed 

We need to test if this is considered a breaking change.

jkroepke:
It must be considered as breaking change.

If end user set `SelectorNilUsesHelmValues` and doesn't change the default value, while we changing the default at values.yaml, there result would be a breaking change.

However, I feel thats fine by bumping major version + documentation.

schnatterer:
Hi @jkroepke and @GMartinez-Sisti, 
Thanks for picking up on the PR.
I'm also a fan of consistency and happy to improve the chart as a whole. 

`SelectorNilUsesHelmValues`  always has been a bit confusing to me, so getting rid of it would be nice. 

I tried your suggestion and it seems to fix my problem.

The default values fail, though.

```
Error: template: kube-prometheus-stack/templates/prometheus/prometheus.yaml:249:80: executing "kube-prometheus-stack/templates/prometheus/prometheus.yaml" at <eq .Values.prometheus.prometheusSpec.scrapeConfigSelector "">: error calling eq: incompatible types for comparison
```

This can be fixed by dropping the string comparison
```yaml
{{- if not (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil)}}
# instead of
{{- if not (or (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil) (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector "")) }}
```

Next is this:
```
Error: YAML parse error on kube-prometheus-stack/templates/prometheus/prometheus.yaml: error converting YAML to JSON: yaml: line 68: mapping values are not allowed in this context
```
Which could be fixed like so
```yaml
  scrapeConfigSelector:
      {{- tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigSelector | indent 4) . | nindent 2 }}
# instead of
  scrapeConfigSelector: {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigSelector | indent 4) . }}
```
Now templating succeeds but we have redundant quotes:

```bash
$ helm template .   |  grep scrapeConfig -A2
  scrapeConfigSelector:
      matchLabels:
        release: '"release-name"'
        
$  helm template prometheus-community/kube-prometheus-stack --version 58.2.1 | grep scrapeConfig -A2
  scrapeConfigSelector:
    matchLabels:
      release: "release-name"
```

This could be fixed by removing the `quote` function from the helm-values.
```yaml
    scrapeConfigSelector:
      matchLabels:
        release: "{{ $.Release.Name }}"
```
Then, however, the templated YAML contains single quotes instead of double quotes it contained before. Would that matter?
```yaml
 helm template .   |  grep scrapeConfig -A2
  scrapeConfigSelector:
      matchLabels:
        release: 'release-name'
```
To make sure we're all on the same boat, I pushed the current state of the POC, see 739fe4c . Please have a look at it and tell me what you think. If we're all happy with the concept, I can implement the other selectors.

I also see the downside of having to consider this as a breaking change. 
What would that mean for me as as contributor? Would you accept a PR by me that document the change and bump the major version? Or would a change like that delay the delivery of my fix?

schnatterer:
Ah, I missed the test cases for setting other options, which do not seem to work  as expected :grimacing: 

```shell
$ helm template . --values - <<EOF | grep 'scrapeConfigSelector' -A2
prometheus:
  prometheusSpec:
    scrapeConfigSelector: {}
EOF

  scrapeConfigSelector:
      matchLabels:
        release: 'release-name'
```

And this gives an unexpected result:
```shell
$  helm template . --values - <<EOF | grep 'scrapeConfigSelector' -A3
prometheus:
  prometheusSpec:
    scrapeConfigSelector:
       a: b
EOF

  scrapeConfigSelector:
      a: b
      matchLabels:
        release: 'release-name'
```

I'm a bit a loss here with my templating skills.

schnatterer:
As an alternative, could we agree on a less invasive fix, as originally proposed, but for all selectors?

This way we would at least have consistency, and no breaking change.

WDYT @jkroepke ?

jkroepke:
Coud you try to implement the nil check based on https://github.com/Masterminds/sprig/issues/53#issuecomment-483414063

schnatterer:
Interesting solution, @jkroepke 💡

Let me sum up the breaking change:

If you used to set `scrapeConfigSelectorNilUsesHelmValues` to  ` false`.  
You now have to set `scrapeConfigSelector.matchLabels=null`.

In other words:
```bash
#  What used to be:
$ helm template prometheus-community/kube-prometheus-stack --version 62.3.0 --values - <<EOF | grep scrapeConfigSelector
prometheus:
  prometheusSpec:
    scrapeConfigSelectorNilUsesHelmValues: false
EOF
  scrapeConfigSelector: {}

# Is now:
helm template . --values - <<EOF | grep 'scrapeConfigSelector' -A2
prometheus:
  prometheusSpec:
    scrapeConfigSelector:
      matchLabels: null
EOF
  scrapeConfigSelector:
    {}
```

And we can live with the additional line break after `scrapeConfigSelector` here 👆️, right? 

We can also live with the single quotes in the default values 👇️, right?

```bash
# What used to be:
$ helm template prometheus-community/kube-prometheus-stack --version 62.3.0 | grep scrapeConfigSelector -A3
  scrapeConfigSelector:
    matchLabels:
      release: "release-name"
# Is now:
$  helm template .  | grep 'scrapeConfigSelector' -A2
  scrapeConfigSelector:
    matchLabels:
      release: 'release-name'
```

The same is true for all `*NilUsesHelmValues` of the selectors you mentioned above, right?

Plus, I can now set the selector to `null` to remove it altogether for downward compatibility (my original use case).

So what I will do is
* implement this for all selectors mentioned above
* Remove all `*NilUsesHelmValues` from `values.yaml`
* Adapt the comment over the selectors to reflect `matchLabels: null` as new `NilUsesHelmValues`, and mention the option of setting it to `null` (please review this in my suggestion in b83ccf12)
* Add a section in the README `### From 62.x to 63.x` that describes the breaking change
* bump the major version

And after some polishing during review you will merge the PR?

I'm asking because the number of changes will make this quite an invest for me, and I want to make sure we all agree on this path, so my time is not wasted.

jkroepke:
> And we can live with the additional line break after scrapeConfigSelector here 👆️, right?

It is at least valid YAML. That what i checked. If it causes problems, I have already an idea to fix that.

> We can also live with the single quotes in the default values 👇️, right?

Kubernetes don't care if single or double quotes are used. The YAML document will be convered to an internal structure anyways. That why kubectl get yaml looks always different than the input. If you work with `helm template --release-name foo`, the quotes should be omit anyways.

> I'm asking because the number of changes will make this quite an invest for me, and I want to make sure we all agree on this path, so my time is not wasted.

I would agree here, however I also may a different opinion. If there is an breaking change which simplifies the maintenance, I would go of it. We are not writing enterprise software. 

That why I feel important to hear an different opinion an other maintainer, e.g. @GMartinez-Sisti or @QuentinBisson

charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -246,21 +246,8 @@ spec:
...
-{{ else }}
-  scrapeConfigSelector: {}
-{{- end }}
-{{- if .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector }}
-  scrapeConfigNamespaceSelector:
-{{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector | indent 4) . }}
-{{ else }}
-  scrapeConfigNamespaceSelector: {}
+{{- if not (kindIs "invalid" .Values.prometheus.prometheusSpec.scrapeConfigSelector) }}
+  scrapeConfigSelector: {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigSelector) . | nindent 4 }}
```
jkroepke, line 250: scrapeConfigSelector and scrapeConfigNamespaceSelector are distinct properties.

schnatterer: Good catch! Brought it back in 0b8eb3c1 and will convert it with the other properties if enough maintainers agree.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -3683,16 +3683,15 @@ prometheus:
...
 
-    ## If true, a nil or {} value for prometheus.prometheusSpec.scrapeConfigSelector will cause the
-    ## prometheus resource to be created with selectors based on values in the helm deployment,
-    ## which will also match the scrapeConfigs created
-    ##
-    scrapeConfigSelectorNilUsesHelmValues: true
-
     ## scrapeConfigs to be selected for target discovery.
-    ## If {}, select all scrapeConfigs
+    ## If matchLabels.release: "{{ $.Release.Name }}" the prometheus resource will be created
```
schnatterer, line 3687: @jkroepke did you have a look at this? Is it correct this way?

schnatterer:
@GMartinez-Sisti @QuentinBisson Could you state your opinion, if we should go forward with the breaking changes? [This comment](#issuecomment-2317206528)  sums up what we plan, and [jkroepke states his opinion in this comment](#issuecomment-2317295445).

If you have doubts about a breaking change, we could go with my original approach in 7157047, which is minimally invasive but harder to maintain. I'd still be happy with this approach, because I only suggested a small fix and would be happy to get it faster and with less effort :slightly_smiling_face:

QuentinBisson:
Thanks for the great work @schnatterer and @jkroepke 🚀

charts/kube-prometheus-stack/values.yaml
jkroepke: It looks correct to me for the Selectors. Keep in mind that the namespaceSelector behave a bit differently.

For If null case. also mention that the entire resource is ignored.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -3683,16 +3683,15 @@ prometheus:
...
+    ## with selectors based on values in the helm deployment, which will also match the scrapeConfigs created
+    ## If matchLabels: null" select all scrapeConfigs
```
jkroepke, line 3689:
```suggestion
    ## To remove matchLabels from the selector condition, explicit set matchLabels to null. If not other selectors are configured, prometheus-operator will select all scrapeConfigs.
    ## To remove the release label from the matchLabels condition, explicit set release to null. 

This comment may clarifies, if end-users want to used different selectors other than matchLabels or release.

File: charts/kube-prometheus-stack/values.yaml [hunk 4]
```diff
@@ -3683,22 +3683,24 @@ prometheus:
...
+    ##
+    scrapeConfigSelector:
+      matchLabels:
+        release: "{{ $.Release.Name }}"
     ## Example which selects scrapeConfigs with label "prometheus" set to "somelabel"
     # scrapeConfigSelector:
     #   matchLabels:
     #     prometheus: somelabel
 
     ## If nil, select own namespace. Namespaces to be selected for scrapeConfig discovery.
```
schnatterer, line 3702: Shouldn't this be `If empty, select own namespace`?
`scrapeConfigNamespaceSelector: {}` empty object means select own NS?

schnatterer:
I created an examplary unit test for a selector and a namespaceSelector in 6f3d8269.
This way we should be able to avoid any misunderstandings.

Could you please check if the spec of the tests are correct and complete, @jkroepke?

I would then go forward and implement the same for the other selectors.

The test can be run using
```
helm unittest -f "unittests/prometheus/*.yaml" .
```

Can I somehow add a warning about the deprecated `NilUsesHelmValues` fields, to avoid unpleasant surprises for users?

jkroepke:
Yeah keep it for know, the PR is already large enough.

charts/kube-prometheus-stack/values.yaml
jkroepke: https://github.com/prometheus-operator/prometheus-operator/blob/781d6df6bc8e48b010a74348f19526f66c5e2c54/pkg/apis/monitoring/v1/prometheus_types.go#L177-L179


```
	// Namespaces to match for ScrapeConfig discovery. An empty label selector
	// matches all namespaces. A null label selector matches the current
	// namespace only.
```

Thats what I understand here.

jkroepke:
I looked through the code and first insights looks good. I will also take a integration test on my side, but I expect that this happens in the next few days.

schnatterer:
Thanks for the preliminary review, @jkroepke . 
I'll also happen to have an application at hand, which I will use as an integration test.

For that, I manually deployed the current HEAD commit on the GH pages of my fork:

```
helm repo add https://schnatterer.github.io/prom-helm-charts   
helm template  schnatterer-prom-helm-charts/kube-prometheus-stack --version 63.0.0
 ```
Maybe that's useful for you integration test as well.
 
What about the deprecation warning? I have seen that as a user in the Jenkins chart: 
 https://github.com/jenkinsci/helm-charts/blob/jenkins-5.5.16/charts/jenkins/templates/deprecation.yaml
 Do we want a mechanism like this in the prometheus-stack chart as well?

**Comment by jkroepke** [comment]
> What about the deprecation warning?

Why not. If you think it's adds an value, then please add it with an commentary, that it can be removed 02/2025. I have mixed feeling, but it could help end-users. 

But please do not explain in the message, what needs to be todo, just reference the Changelog note.

**Comment by schnatterer** [comment]
Done.
Why can it be removed 02/2025? My experience with enterprises is that they have much longer upgrade cycles. I therefore took the liberty to mention 09/25 for deletion, which is a year after deprecation. It still might no be enough for some 😬
I might argue: Why delete it at all? The warnings are separated into a file so are not too painful in terms of maintenance.

Anyways, with e38007a  the PR is ready for review.
We'll probably rebase and squash at the end?

**Comment by jkroepke** [comment]
I still having the position that we are writing OSS support, not enterprise software. But okay. I'm I don't care.

File: charts/kube-prometheus-stack/templates/deprecation.yaml [hunk 5]
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.checkDeprecation }}
```
jkroepke, line 1: Please link

`https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-62x-to-63x`

jkroepke (approved):
From technical side, everything looks good to me. But I would like to see some changes related to docs.

schnatterer:
Thanks for your review, @jkroepke. Good catches. Resolved in 39d577aa.
Should I go ahead with squashing or do we want another approval first?

jkroepke:
Based on your doc changes, I remember that we have only prometheus in scope. We can also this logic to AM and Ruler.


* https://github.com/prometheus-community/helm-charts/blob/3612c3b728884fd7e6564c87ed3dfb29db7e7936/charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml#L65-L76
* https://github.com/prometheus-community/helm-charts/blob/3612c3b728884fd7e6564c87ed3dfb29db7e7936/charts/kube-prometheus-stack/templates/thanos-ruler/ruler.yaml#L55-L70

I can do this on evening, then it should be complete.

I feel confident enough to approve and merge it by myself. Just for concepts, I also like to hear a second opinion.

schnatterer:
Wow, that process of linting is dragging on. For me as a contributor it would make things a lot faster, if the issue template would contain a pointer to a doc that tells me how to run the different linters locally, e.g. using docker.

jkroepke:
> Wow, that process of linting is dragging on. For me as a contributor it would make things a lot faster, if the issue template would contain a pointer to a doc that tells me how to run the different linters locally, e.g. using docker.

Good point, could you create an issue for that?

jkroepke:
@schnatterer Thanks for you contribution!

QuentinBisson:
Yes thank you both for the great work

schnatterer:
🎉 Thanks for accepting my proposal and for the many discussions @jkroepke, @GMartinez-Sisti and @QuentinBisson !

Great result!

Pionerd:
We use kube-prometheus-stack as a sub chart. We are unable to set the `serviceMonitorSelector` back to null due to https://github.com/helm/helm/pull/12879

Any suggestions? For us this is a breaking change.

jkroepke:
Allowing false as alternative would be sufficient?

Pionerd:
Perfect

KyriosGN0:
@Pionerd @jkroepke i have the same issue, the fix is to use the old key or set `matchLabels` to be false ?

jkroepke:
There is no known fix yet, only a proposal.

jkroepke:
Not sure, if this would work right now:

```
kube-prometheus-stack:
  prometheus:
    prometheusSpec:
      additionalConfig:
        serviceMonitorNamespaceSelector: ~
```

PR #4851, opened 2024-09-12
[kube-prometheus-stack] Fix jobLabel on prometheus-node-exporter

monchaio:
When crate prometheus with existing chart I found that jobLabel from metrics of node exporters are not "node-exporter".
After investigation I found that we need to use commonLabels instead of podLabels.
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/templates/_helpers.tpl#L46.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 62.6.0
+version: 62.6.1
 appVersion: v0.76.1
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -2122,7 +2122,7 @@ nodeExporter:
 ##
 prometheus-node-exporter:
   namespaceOverride: ""
-  podLabels:
+  commonLabels:
     ## Add the 'node-exporter' label to be used by serviceMonitor to match standard common usage in rules and grafana dashboards
     ##
     jobLabel: node-exporter


jkroepke:
hm. I looked in that PR and I don't understand *WHY* is solves your issues. I'm unable to find the root cause here.

Normalls, pod and service monitor taking the label from the pod and service I can see that service and pod labels are defined.

monchaio:
> hm. I looked in that PR and I don't understand _WHY_ is solves your issues. I'm unable to find the root cause here.
> 
> Normalls, pod and service monitor taking the label from the pod and service I can see that service and pod labels are defined.

Because prometheus-node-exporter change theirs _helper.tpl to use commonLabels instead of podLabels https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/templates/_helpers.tpl#L46

In older version of prometheus-node-exporter they use podLabels.
```
{{ include "prometheus-node-exporter.selectorLabels" . }}
{{- with .Chart.AppVersion }}
app.kubernetes.io/version: {{ . | quote }}
{{- end }}
{{- with .Values.podLabels }}
{{ toYaml . }}
{{- end }}
```

Without jobLabel on servicemonitoring. Any metrics from node-exporter have job label in format "<release>-prometheus-node-exporter" and break many rule that still use job label "node-exporter".

```
node_cpu_seconds_total{mode="idle",job="<release>-prometheus-node-exporter"}
```

jkroepke:
If the Service Monitor requires that label, using prometheus.monitor.additionalLabels would enough?

zeritti:
> Without jobLabel on servicemonitoring. Any metrics from node-exporter have job label in format "<release>-prometheus-node-exporter" and break many rule that still use job label "node-exporter".

The label expected by the service monitor, `jobLabel`, at the Node exporter's service has been set in `prometheus-node-exporter.service.labels` by default to `node-exporter` since 62.3.0 ([values.yaml](https://github.com/prometheus-community/helm-charts/blob/6a6225de96e3bbe93872fec5d81ac6b067d20cce/charts/kube-prometheus-stack/values.yaml#L2139)). Prometheus operator sets the Node exporter's job name correctly to the label's value.

jkroepke:
Thats what I observe. The current config should just work. Or I'm wrong?

PR #4855, opened 2024-09-13
[prometheus-blackbox-exporter] add commonLabels to pod

es1o:

Add `commonLabels` to Pod in Deployment.

#### Checklist

- [ ] Chart Version bumped


charts/prometheus-blackbox-exporter/templates/deployment.yaml
```diff
@@ -21,6 +21,9 @@ spec:
     metadata:
       labels:
         {{- include "prometheus-blackbox-exporter.selectorLabels" . | nindent 8 }}
+        {{- if .Values.commonLabels }}
+{{ toYaml .Values.commonLabels | indent 8 }}
+        {{- end }}
```
monotek, line 26: this should not be necessary: https://github.com/es1o/helm-charts/blob/common-labels/charts/prometheus-blackbox-exporter/templates/_helpers.tpl#L46-L48

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/deployment.yaml
es1o: @monotek It is necessary because `prometheus-blackbox-exporter.labels` are not used here, only `prometheus-blackbox-exporter.selectorLabels` that doesn't have `.Values.commonLabels`.

monotek (approved):

PR #4610, opened 2024-06-18
[prometheus] fix: Add apiVersion and kind to STS persistent volume claim

grzesuav:
#### What this PR does / why we need it

Add explicitly `persistentVolumeClaim` API version and kind  - they are added by API server which causes a drift in argo cd state - and having them do not make any charm.

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #4865, opened 2024-09-19
[alertmanager] Add apiVersion and kind to StatefulSet volumeClaimTemplates

### LINKED ISSUE #4848: [prometheus-alertmanager] ArgoCD can't take a fullsync StatefulSet

K8s is adding automatic some strings in the volumeClaimTemplates after deploy:

StatefulSet in the cluster:

```
....
  volumeClaimTemplates:
    - kind: PersistentVolumeClaim
      apiVersion: v1
....
```

StatefulSet in the helm chart:

```
...
  {{- if .Values.persistence.enabled }}
  volumeClaimTemplates:
    - metadata:
        name: storage
      spec:
...
```

Screenshot 2024-09-11 at 12 24 32

### What's your helm version?

v3.13.2

1.28

### Which chart?

alertmanager

### What's the chart version?

1.12.0

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

i'm using an ArgoCD

_No response_

schedin:
#### Background
[ArgoCD](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#declarative) may take Helm charts as inputs, but does not install the Helm charts directly, instead it uses the `helm template` command that produces YAML file resources and then injects them directly. The result will look almost like a `helm install` command, but the deployment will not show up using `helm list`.

By using this mechanism ArgoCD can perform its own “diffing” and warn when the desired state does not match the cluster (for example if a person modified the resources after the deployment). It has the drawback that if some mechanism modifies the deployed resource(s) in an automatic way, ArgoCD will show the sync status as “OutOfSync” with a yellow arrow up icon.

The Alertmanager StatefulSet has a `volumeClaimTemplates:`  with one volume claim template. This template is missing the fields `apiVersion` and `kind`. When installed by Helm, Kubernetes will automatically add these missing fields (by a mechanism I have not digged deeper into).

This pull request will make the StatefulSet look more similar when comparing `helm template` with how the resource looks like in the cluster. This will make ArgoCD happy (with a green checkmark icon).

This pull request is analogous to what #4610 fixed.

- fixes #4848

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/alertmanager/Chart.yaml
+++ b/charts/alertmanager/Chart.yaml
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.12.0
+version: 1.12.1
 appVersion: v0.27.0
 kubeVersion: ">=1.19.0-0"
 keywords:
--- a/charts/alertmanager/templates/statefulset.yaml
+++ b/charts/alertmanager/templates/statefulset.yaml
@@ -230,7 +230,9 @@ spec:
         {{- end }}
   {{- if .Values.persistence.enabled }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
         name: storage
       spec:
         accessModes:


monotek (changes requested):
please solve your issiues in argocd not in the chart.

schedin:
I'm not affiliated with ArgoCD. It is correct that this issue could be resolved in the realm of ArgoCD, without contributing to the Prometheus Community charts.

I submitted this pull request because of the similar pull request #4610 that was already merged, but code reviewed by others (@zeritti) in this case.

xorinzor:
@monotek this is something that should be fixed in the helm template, not patched by everyone in their ArgoCD deployment.

PR #4870, opened 2024-09-20
Add ability to set topologySpreadConstraints for prometheus-snmp-exporter

timp87:


walker-tom (approved):
Sorry for the delay in looking at this, changes look good thanks for the PR!

PR #4875, opened 2024-09-26
[prometheus-mongodb-exporter] Support multi-target exporter

### LINKED ISSUE #4052: [prometheus-mongodb-exporter] Multi Target Support

Since version 0.40.0 the personal mongodb exporter ships multitarget support. It would be great if the mongodb exporter deployed via helmchart could support multitarget. With the current helmchart it is only scraping a single instance.

https://github.com/percona/mongodb_exporter#multi-target-support

Tbd

Tbd

### Additional context.

_No response_

BapRx:

Allow configuring multiple targets for the exporter v0.40.0 and later.

- fixes #4052

Until the exporter supports reading passwords from a file, we are limited to choose out of two solutions:
- generate `MONGODB_URI` automatically from the target list BUT share the same username/password
- create a secret manually with the content of `MONGODB_URI` but also manage the serviceMonitor list separately

I don't really know what's the best solution, open for suggestions! I went for the first solution which matches my use case but if you think the other option is better for most people, we can rethink the logic.

#### Checklist

- [x] Chart Version bumped


zeritti (changes requested):
Thank you, @BapRx, for your PR. Please, see my comments below.

BapRx:
Hello @zeritti, thank you for your review, I pushed some changes to fix the issues you mentioned.

vladimirpetr:
Thank you, @BapRx. Please, remember to bump the chart version.

zeritti:
Thank you, @BapRx, for your changes. I will still validate the changes over the coming days. As to the chosen use case of a single set of credentials, I am confident that we can start with that. 
I would also suggest that we consider creating a CI test case - Prometheus operator CRDs are being included in a CI run and hence, we can make use of it.

zeritti:
A CI values file for `ct` must have name in `*-values.yaml`, e.g. `servicemonitor-multi-target-values.yaml`, otherwise it gets ignored.

zeritti (approved):

PR #4899, opened 2024-10-04
[kube-prometheus-stack] support kubelet endpoint slices

### LINKED ISSUE #4897: [kube-prometheus-stack] support --kubelet-endpointslice=true

with 65.0.0, `--kubelet-endpointslice=true` feature is available. please support setting 
```
        - --kubelet-endpoints=false
        - --kubelet-endpointslice=true
```
to the operator in values

also when i tried enabling it, operator failed to run due to 
```
ts=2024-10-03T22:27:05.878020844Z level=warn caller=/home/runner/work/prometheus-operator/prometheus-operator/cmd/operator/main.go:556 msg="missing permission on resource 'endpointslices' (group: \"discovery.k8s.io\")" reason="missing \"create\" permission on resource \"endpointslices\" (group: \"discovery.k8s.io\") for namespace \"kube-system\""
ts=2024-10-03T22:27:05.878065181Z level=warn caller=/home/runner/work/prometheus-operator/prometheus-operator/cmd/operator/main.go:556 msg="missing permission on resource 'endpointslices' (group: \"discovery.k8s.io\")" reason="missing \"update\" permission on resource \"endpointslices\" (group: \"discovery.k8s.io\") for namespace \"kube-system\""
ts=2024-10-03T22:27:05.878072703Z level=warn caller=/home/runner/work/prometheus-operator/prometheus-operator/cmd/operator/main.go:556 msg="missing permission on resource 'endpointslices' (group: \"discovery.k8s.io\")" reason="missing \"delete\" permission on resource \"endpointslices\" (group: \"discovery.k8s.io\") for namespace \"kube-system\""
```

looks like in `clusterrole/kube-prometheus-stack-operator`
you only have 
```
- apiGroups:
  - discovery.k8s.io
  resources:
  - endpointslices
  verbs:
  - get
  - list
  - watch
```

please update it to
```
- apiGroups:
  - discovery.k8s.io
  resources:
  - endpointslices
  verbs:
  - get
  - create
  - update
  - delete
  - list
  - watch
```

### What's your helm version?

3.16.1

1.29.1

### Which chart?

kube-prometheus-stack

### What's the chart version?

65.0.0

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

command works, i need to enable new functionality

_No response_

DrFaust92:

- fixes #4897

Based one upstream https://github.com/prometheus-operator/prometheus-operator/pull/6882

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
@@ -103,7 +103,10 @@ rules:
   - endpointslices
   verbs:
   - get
+  - create
   - list
   - watch
+  - update
+  - delete
 {{- end }}
 {{- end }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -63,12 +63,8 @@ spec:
             - --kubelet-selector={{ .Values.prometheusOperator.kubeletService.selector }}
             {{- end }}
             {{- end }}
-            {{- if .Values.prometheusOperator.kubeletEndpointsEnabled }}
             - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
-            {{- end }}
-            {{- if .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
             - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
-            {{- end }}
             {{- if .Values.prometheusOperator.logFormat }}
             - --log-format={{ .Values.prometheusOperator.logFormat }}
             {{- end }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -63,6 +63,12 @@ spec:
             - --kubelet-selector={{ .Values.prometheusOperator.kubeletService.selector }}
             {{- end }}
             {{- end }}
+            {{- if .Values.prometheusOperator.kubeletEndpointsEnabled }}
+            - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
+            {{- end }}
+            {{- if .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
+            - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
+            {{- end }}
             {{- if .Values.prometheusOperator.logFormat }}
             - --log-format={{ .Values.prometheusOperator.logFormat }}
             {{- end }}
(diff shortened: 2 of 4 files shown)


charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
```diff
@@ -103,7 +103,10 @@ rules:
   - endpointslices
   verbs:
   - get
+  - create
```
DrFaust92, line 106: Ideally we can use the `kubeletEndpointsEnabled` and `kubeletEndpointSliceEnabled` to toggle granting these permissions. will let reviewer decide. i can add this logic if needed

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -63,6 +63,12 @@ spec:
             - --kubelet-selector={{ .Values.prometheusOperator.kubeletService.selector }}
             {{- end }}
             {{- end }}
+            {{- if .Values.prometheusOperator.kubeletEndpointsEnabled }}
```
jkroepke, line 66: if in combination with an boolean might be mis-leading.

Since `--kubelet-endpoints=` is set to true by default, I set the value to `false` keep the flag still to true, because `--kubelet-endpoints=false` would be omit.

not sure, if

`{{- if (print .Values.prometheusOperator.kubeletEndpointsEnabled) }}`

would help here, since it converts the value to a string and `"false"` is still true-ish, I guess?

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
jkroepke: The toggle would cover only write permissions, right?

jkroepke (changes requested):
LGTM.

Reading discussion upstream and it seems kinda useful for larger clusters. I as understand, OpenShift toggle the default to endpointslices. WDYT about? We can also leave the upstream defaults.

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
DrFaust92: Good point, maybe just omit the ifs entirely and go by the values.yaml values?

DrFaust92:
[jkroepke](https://github.com/jkroepke) I think (assume) to leave it same as upstream for the "common" case.

Regarding permissions, it seems that there isnt a reason to have any endpoint slice permissions if toggle isnt on. so it would cover all of it. and ill need to seperate `endpoints` permissions to its own block as well and keep it behind a toggle

charts/kube-prometheus-stack/templates/prometheus-operator/clusterrole.yaml
DrFaust92: No, I think all of the endpointslice permissions. unelss there is a reason not to (I dont see one)

DrFaust92:
@jkroepke any additional thoughts before i make changes? I think i wont add toggle to avoid edge cases/breaking too much

jkroepke:
I would says omit toggles overall, including at CLI flags. That would avoid the issue to set the CLI flags to false.

jkroepke (approved):

PR #4900, opened 2024-10-06
[kube-prometheus-stack] add tpl to prometheus service account

denzhel:
**What this PR does / why we need it**

We're moving our infra charts from Terraform to ArgoCD/ApplicationSets and we need the ability to template
some values with an environment name, accountID and etc ... 

In this case, we need the ability to template the service account and inject account id and environment name to the IAM role ARN 

- [x] Chart Version bumped


jkroepke (approved):

jkroepke:
Thanks

PR #4883, opened 2024-10-02
[kube-prometheus-stack] Revert "Add downward compat for Prom CRD (#4818)"

jkroepke:
This reverts commit 25f32d65

#### What this PR does / why we need it

See #4878

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #4878

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

Issue #4878, opened 2024-09-27
[kube-prometheus-stack] Lost Prometheus targets after upgrade to version 63


kallaics:
### Describe the bug a clear and concise description of what the bug is.

We have PodMonitor, ServiceMonitor and ScrapeConfig resources and after the upgrade these resources are completely disappeared from the targets list in Prometheus and not collect the metrics. I spent a morning to find any relevant change in the release notes, but I don't found anything. The only solution was the downgrade.
All affected resources are included through <resource name>Selector.matchLabels

Worked well on version:   62.7.0
Not worked on version: 63.0.0 and 63.0.1

Prometheus configuration

```yaml
prometheus:
  enabled: true
  ingress:
    enabled: true
    ingressClassName: nginx
    annotations:
      nginx.ingress.kubernetes.io/whitelist-source-range: <ip ranges>
    labels: {}
    hosts:
      - prometheus.<domain>.<tld>
    paths:
      - /
    pathType: ImplementationSpecific
    tls: []
  prometheusSpec:
    prometheusExternalLabelNameClear: true
    replicaExternalLabelNameClear: true
    podMonitorSelector:
      matchLabels:
        app.kubernetes.io/part-of: flux
    serviceMonitorSelector:
      matchLabels:
        monitoring: prometheus
    scrapeConfigSelector:
      matchLabels:
        monitoring: prometheus
    secrets:
      - ca-certs
    additionalArgs:
      - name: "web.enable-admin-api"
    resources:  
      requests:
        memory: 6Gi
```

Scrape configuration (I know it is alpha status, so we are testing on non-production environment only)

```yaml
apiVersion: monitoring.coreos.com/v1alpha1
kind: ScrapeConfig
metadata:
  name: <Kubernetes cluster name>
  namespace: monitoring
  labels:
    monitoring: prometheus-scrape
spec:
  scrapeInterval: 15s
  scrapeTimeout: 12s
  honorLabels: true
  metricsPath: '/federate'
  scheme: HTTPS
  tlsConfig:
    ca:
      secret:
        key: internal-ca.pem
        name: ca-certs
  basicAuth:
    username:
      key: username
      name: monitoring-federate-auth
    password:
      key: password
      name: monitoring-federate-auth
  params:
    'match[]':
      - '{__name__=~".+"}'
  staticConfigs:
    - labels:
        job: <Kubernetes cluster name>
        env: dev
        cluster_name: <Kubernetes cluster name>
      targets:
[...]

PR #4906, opened 2024-10-09

schnatterer:

```
failed to create typed patch object (..): .spec.scrapeConfigNamespaceSelector: field not declared in schema
```

```
prometheus:
  prometheusSpec:
    scrapeConfigNamespaceSelector: null
    scrapeConfigSelectorNilUsesHelmValues: null
    scrapeConfigSelector: null
```

At runtime, the operator yields some warnings about the missing fields, but apparently everything else works.

As discussed with @jkroepke I'm launching another try to include this feature :slightly_smiling_face: 

This is the same functionality as in #4818, which unfortunately was reverted in #4883, because it hit a bug in helm helm/helm#12879.

If we go for my original minimally invasive approach, there is no breaking change we won't cause #4878 again.

It won't work for sub charts (because of aforementioned bug in helm) but at least for direct users of kps chart.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -253,13 +253,13 @@ spec:
   scrapeConfigSelector:
     matchLabels:
       release: {{ $.Release.Name | quote }}
-{{ else }}
+{{ else if not (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil) }}
   scrapeConfigSelector: {}
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector }}
   scrapeConfigNamespaceSelector:
 {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector | indent 4) . }}
-{{ else }}
+{{ else if not (eq .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector nil) }}
   scrapeConfigNamespaceSelector: {}
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.storageSpec }}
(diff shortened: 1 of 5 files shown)


charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -253,13 +253,13 @@ spec:
...
       release: {{ $.Release.Name | quote }}
-{{ else }}
+{{ else if not (eq .Values.prometheus.prometheusSpec.scrapeConfigSelector nil) }}
   scrapeConfigSelector: {}
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector }}
   scrapeConfigNamespaceSelector:
 {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector | indent 4) . }}
-{{ else }}
+{{ else if not (eq .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector nil) }}
```
jkroepke, line 262: what not `ne` instead of `not (eq` ?

schnatterer: Good point @jkroepke, that's shorter and easier to read. I added unit test and they validate that both expressions achieve the same result.
Also added docs.

Should we implement this behavior for the same selectors as the original PR?
In my use case only the ones included in the PR caused the problem.

charts/kube-prometheus-stack/values.yaml
jkroepke, line 1: consider to revert all changes here.

schnatterer: You mean the whitespaces? Done.
My IDE somehow did that without my consent 😠

jkroepke (approved):

jkroepke:
@QuentinBisson please double check this, LGTM

QuentinBisson (approved):
LGTM as well

schnatterer:
Thank you @jkroepke and @QuentinBisson for your support!

PR #4915, opened 2024-10-15
[prometheus-blackbox-exporter] Adjust configPath handling

Jakob3xD:
**Issue**
I want to deploy the config map for the Blackbox config on my own, as I want to write some logic that adjust the ConfigMap more dynamically. Therefore I need to `null` the `config` value as otherwise the ConfigMap is created with the default chart values. When setting `config` to `null` the config path is set to `/etc/blackbox_exporter/config.yml` which does not exist.

**Changes**
I changed the config path logic to match the one of the config reloader. If no `configPath` is set, we use the default of `/config/blackbox.yaml` otherwise we use the `configPath` value. The default value of `/etc/blackbox_exporter/config.yml` gets removed as I don´t see the point in keeping it. If needed it can be set via `configPath`.

As it its a breaking change, I also bumped the version.

 None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
+++ b/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
@@ -228,7 +228,7 @@ containers:
   {{- end }}
   {{- end }}
   args:
-  - --config-file={{ .Values.configPath | default "/config/blackbox.yaml" }}
+  - --config.file={{ .Values.configPath | default "/config/blackbox.yaml" }}
   {{- with .Values.extraArgs }}
 {{ tpl (toYaml .) $ | indent 2 }}
   {{- end }}
--- a/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
+++ b/charts/prometheus-blackbox-exporter/templates/_helpers.tpl
@@ -228,15 +228,7 @@ containers:
   {{- end }}
   {{- end }}
   args:
-  {{- if .Values.config }}
-  {{- if .Values.configPath }}
-  - "--config.file={{ .Values.configPath }}"
-  {{- else }}
-  - "--config.file=/config/blackbox.yaml"
-  {{- end }}
-  {{- else }}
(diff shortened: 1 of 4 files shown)


charts/prometheus-blackbox-exporter/templates/_helpers.tpl
```diff
@@ -228,15 +228,7 @@ containers:
...
-  {{- if .Values.config }}
-  {{- if .Values.configPath }}
-  - "--config.file={{ .Values.configPath }}"
-  {{- else }}
-  - "--config.file=/config/blackbox.yaml"
-  {{- end }}
-  {{- else }}
-  - "--config.file=/etc/blackbox_exporter/config.yml"
-  {{- end }}
+  - --config.file={{ .Values.configPath | default "/config/blackbox.yaml" }}
```
monotek, line 231: As this is a major updat anyway we might think about to rename the var to `.Values.configFile` too.
Imho this fits better as the var currently contains a filename and not only the path.
Thoughts?

monotek (changes requested):

charts/prometheus-blackbox-exporter/templates/_helpers.tpl
Jakob3xD: I am really sorry for the late response. I lost track of the PR.

You are correct, that naming it configPath would match its actual purpose.

monotek: But the file does exist in the docker image: https://github.com/prometheus/blackbox_exporter/blob/master/Dockerfile#L9C1-L9C49

Jakob3xD: You are right, I was looking at the wrong container (config reloader).

I'll close the PR. However, I am asking if config reloader should be removed if no config and no config Path is set, as it wouldn't do anything.

PR #4941, opened 2024-10-26
[prometheus-postgres-exporter] support custom TLS config

### LINKED ISSUE #4940: [prometheus-postgres-exporter] Support custom TLS configuration

The current version of the prometheus-postgres-exporter helm chart does not support TLS encryption between prometheus and the exporter. It is currently not possible to define a custom TLS configuration as part of the serviceMonitor resurce. 

In my opinion is this a security issue, because the metrics can be accessed from other application running inside the cluster. Only network policies can currently avoid accessing the metric endpoint. For this reason should be the helm chart adapted, that users can without implementing network policies use TLS authentication and encryption for the metrics endpoint.

I would like to have some configuration options in the `values.yaml` to customize the TLS configuration used py prometheus to scrape the metrics endpoint. 

Inspect more details via `kubectl explain servicemonitor.spec.endpoints.tlsConfig` and `kubectl explain servicemonitor.spec.endpoints.scheme`.

No alternative.

### Additional context.

As far as I know. It is currently not possible to start the web server of prometheus-postgres-exporter with TLS encryption. An issue regarding the missing feature of TLS encryption as part of the application is still [open](https://github.com/prometheus-community/postgres_exporter/issues/454) since some years. 

I think the feature to configure a custom TLS configuration as part of the serviceMonitor should nevertheless be implemented.

volker-raschek:
The following patch extends the serviceMonitor resource to allow users to apply a custom TLS configuration for TLS authentication and encryption between prometheus and the postgres-exporter. Furthermore, the protocol scheme can now be customized. The default is `http` but needs to be changed to `https`, when TLS authentication/encryption should be applied.

A user-specific TLS configuration can be applied as follows.

```yaml
serviceMonitor:
  enabled: true
  scheme: https
  tlsConfig:
    caFile: /etc/prometheus/tls/ca/ca.crt
    certFile: /etc/prometheus/tls/app2app/tls.crt
    keyFile: /etc/prometheus/tls/app2app/tls.key
    insecureSkipVerify: false
    serverName: prometheus-postgres-exporter
```

Important Note: The `serverName` attribute must correspond to the CommonName or a Subject Alternative Name (SAN) of the TLS certificate. If this is not the case, prometheus will reject the connection trying to match the IP address of the pod with the CommonName / SAN.

The client certificate and private key as well as the certificate of the certificate authorithy must be mounted additionally via the `extraVolumes` and `extraVolumeMounts` option. This configuration is not standard and must also be implemented by the user if TLS client authentication is required.

- fixes #4940

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
@@ -20,6 +20,9 @@ spec:
 {{- if .Values.serviceMonitor.telemetryPath }}
     path: {{ .Values.serviceMonitor.telemetryPath }}
 {{- end }}
+{{- if .Values.serviceMonitor.scheme }}
+    scheme: {{ .Values.serviceMonitor.scheme }}
+{{- end }}
 {{- if .Values.serviceMonitor.timeout }}
     scrapeTimeout: {{ .Values.serviceMonitor.timeout }}
 {{- end }}
@@ -30,6 +33,10 @@ spec:
 {{- if .Values.serviceMonitor.relabelings }}
     relabelings:
 {{ toYaml .Values.serviceMonitor.relabelings | nindent 4 }}
+{{- end }}
+{{- with .Values.serviceMonitor.tlsConfig }}
+    tlsConfig:
+{{- toYaml . | nindent 6 }}
 {{- end }}
   jobLabel: {{ template "prometheus-postgres-exporter.fullname" . }}
   namespaceSelector:
--- a/charts/prometheus-postgres-exporter/values.yaml
+++ b/charts/prometheus-postgres-exporter/values.yaml
@@ -45,6 +45,10 @@ serviceMonitor:
   # metricRelabelings: []
   # Set relabel_configs as per https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
   # relabelings: []
+  # HTTP scheme to use for scraping. For example `http` or `https`. Default is `http`.
+  # scheme: http
+  # TLS configuration to use when scraping the metric endpoint by Prometheus.
+  # tlsConfig: {}
 
 prometheusRule:
   enabled: false
@@ -195,13 +199,25 @@ annotations: {}
 
 podLabels: {}
 
-# Configurable health checks
+# Configurable probes. If TLS client authentication is enabled at the exporter, TCP probe must be used.
 livenessProbe:
   initialDelaySeconds: 0
+  httpGet:
+    path: /
+    port: http
+    scheme: HTTP
+  # tcpSocket:
+  #   port: http
   timeoutSeconds: 3
 
 readinessProbe:
   initialDelaySeconds: 0
+  httpGet:
+    path: /
+    port: http
+    scheme: HTTP
+  # tcpSocket:
+  #   port: http
   timeoutSeconds: 1
 
 # Labels and annotations to attach to the deployment resource
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "v0.15.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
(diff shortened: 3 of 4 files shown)


zeritti:
Thank you, @volker-raschek, for your PR. Please, sign your commits.

zeritti:
Please, bump the version in Chart.yaml (this would be a minor change).

charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
```diff
@@ -30,6 +33,10 @@ spec:
 {{- if .Values.serviceMonitor.relabelings }}
     relabelings:
 {{ toYaml .Values.serviceMonitor.relabelings | nindent 4 }}
+{{- end }}
+{{- with .Values.serviceMonitor.tlsConfig }}
+    tlsConfig:
+{{- . | toYaml | nindent 6 }}
```
zeritti, line 39: This way is more common:

```suggestion
{{- toYaml . | nindent 6 }}
```

zeritti (changes requested):

charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
volker-raschek: I will change that.

volker-raschek:
Hi @zeritti,
I have currently a problem. I've tried to apply the configuration but receive from kubelet an error of the readyness and livenessProbe. The HTTP request returns an unexpected HTTP response, which leads to a CrashbackLoop.

```bash
$ kubectl describe pod prometheus-postgres-exporter-6899b46b7c-7lz9p
...
  Warning  Unhealthy  10m (x9 over 11m)      kubelet            Readiness probe failed: Get "https://10.244.7.17:9187/": remote error: tls: certificate required
  Warning  Unhealthy  10m (x6 over 11m)      kubelet            Liveness probe failed: Get "https://10.244.7.17:9187/": remote error: tls: certificate required
```

I'm not sure whether the httpGet request is correct if the kubelet lacks TLS authentication. 

Would it be possible to simply check the TCP socket instead of httpGet? I would add an option that leaves the default at httpGet and only checks the tcpSocket on request.

The colleagues at gitea do something similar: https://gitea.com/gitea/helm-chart/src/branch/main/values.yaml#L460

zeritti:
> Would it be possible to simply check the TCP socket instead of httpGet? I would add an option that leaves the default at httpGet and only checks the tcpSocket on request.

I'd suggest taking readiness and liveness probes' parameters out of the template as they are now and keep them in the values. In this way, they can be fully customized which is quite often needed. The template will just take each field as a whole, would include e.g. 

```yaml
{{- with .Values.livenessProbe }}
livenessProbe:
  {{- toYaml . | nindent 12 }}
{{- end }}
```

whilst values would include the current settings as defaults

```yaml
livenessProbe:
  initialDelaySeconds: 0 
  timeoutSeconds: 3 
  httpGet:
    path: /
    port: http
    scheme: HTTP
```

In this way one could set any probe type.

volker-raschek:
Hi @zeritti,
I've updated my patch. Please take a review :)

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -195,13 +199,27 @@ annotations: {}
 
 podLabels: {}
 
-# Configurable health checks
+# Configurable probes. Comment in the complete sample definition to deactivate it. Replace the httpGet probe
+# with a TCP probe when the exporter is started with TLS authentication/encryption. Otherwise the application is in a
+# `CrashBackLoop`, because kubelet receives the error message that a TLS certificate is required for authentication.
```
zeritti, line 204: Let us keep it brief, please.

```suggestion
# Configurable probes. If TLS client authentication is enabled at the exporter, TCP probe must be used.
```

zeritti (changes requested):

charts/prometheus-postgres-exporter/values.yaml
volker-raschek: Applied

zeritti (approved):
LGTM

PR #4947, opened 2024-10-27
[prometheus-sql-exporter] Fix labels/matchLabels by using include instead of duplicating declaration in each file

wilfriedroset:

Instead of duplicating the labels & matchLabels in each file we are not relying on include.

N/A

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-sql-exporter/Chart.yaml
+++ b/charts/prometheus-sql-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus SQL Exporter
 name: prometheus-sql-exporter
-version: 0.1.3
+version: 0.2.0
 appVersion: v0.5.8
 home: https://github.com/justwatchcom/sql_exporter
 sources:
--- a/charts/prometheus-sql-exporter/Chart.yaml
+++ b/charts/prometheus-sql-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus SQL Exporter
 name: prometheus-sql-exporter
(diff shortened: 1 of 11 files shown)


DrFaust92 (approved):

zeritti:
Thank you, @wilfriedroset, for your PR. 

The proposed change not only changes the way labels get generated but also replaces their keys, e.g. `app` gets replaced with `app.kubernetes.io/name`, i.e. the old-style labels will be replaced. It is a breaking change as workloads' selector labels cannot be changed. 

Although the chart still is in development (major 0) which allows breaking changes, I'd recommend making an upgrade note, see e.g. [here](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-elasticsearch-exporter#to-600). You might want to consider taking the chart out of initial development, too. Otherwise, the minor version number should be raised.

wilfriedroset:
Thank you @zeritti for your feedbacks. You are obviously right about the breaking change. If you don't mind, I would like to keep the chart in development as I would like to have more community feedbacks. From my point of view I could be the only one using the chart and I might have missed something as obvious as the labels duplication.

I still have a couple of deployment to doing using the chart, if nothing blocks me I will get the chart out of development 😃 

WDYT?

zeritti:
That is quite alright with me, @wilfriedroset, thank you. Please, review NOTES.txt with respect to the changed `app` label and the exporter's port number.

wilfriedroset:
@zeritti the last commit should be good enough. thank you for your time.

zeritti (approved):

PR #4950, opened 2024-10-28
[kube-prometheus-stack] sets podAntiAffinity default to soft

Pandry:
There is a discrepancy between the podAntiAffinity value and what is written in the comments of the chart.  
I think that "soft" as a podAntiAffinity is a reasonable vaule (and the comments say that is the default value).  
Unfortunately, looking at the values, it does not appear that is the actual default value.  
This PR adds it as a default

N/A

#### Checklist

- [x] Chart Version bumped


QuentinBisson (approved):
LGTM

PR #4964, opened 2024-11-03
[prometheus-smartctl-exporter] Don't use `hostNetwork`

### LINKED ISSUE #4963: [prometheus-smartctl-exporter] HostNetwork defaults to true

In the smartctl-exporter Helm Chart the "hostNetwork" value is configurable but defaults to true.

### What's your helm version?

3.16.2

1.31.2

### Which chart?

smartctl-exporter

### What's the chart version?

0.10.0

### What happened?

_No response_

hostNetwork should be false per default

### How to reproduce it?

_No response_

_No response_

N/A

_No response_

judahrand:

I can't see any reason that this exporter requires `hostNetwork: true`. It should only be looking at local devices. This also means that issues with port usage can occur when also using `node-exporter` which does need `hostNetwork`. I think it'd be best to just just remove the line as this PR does.

- fixes #4963

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-smartctl-exporter/Chart.yaml
+++ b/charts/prometheus-smartctl-exporter/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.1
+version: 0.11.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/prometheus-smartctl-exporter/Chart.yaml
+++ b/charts/prometheus-smartctl-exporter/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.0
+version: 0.11.0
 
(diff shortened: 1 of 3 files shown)


kfox1111:
Depends on if prometheus is inside or outside the k8s network....

I think it probably should be made configurable then, rather then just remove it entirely.

judahrand:
> Depends on if prometheus is inside or outside the k8s network....
> 
> I think it probably should be made configurable then, rather then just remove it entirely.

If Prometheus is inside the cluster then couldn't/shouldn't the exporter be exposed via an ingress (or loadbalancer)?

Either way if you think that this should be configurable then I'll switch to that. Maybe best to keep the current default and then just bump a patch version?

kfox1111 (approved):
LGTM. Thanks! :)

kfox1111:
> If Prometheus is inside the cluster then couldn't/shouldn't the exporter be exposed via an ingress (or loadbalancer)?

ingress would expose them load balanced. You couldn't predictably target all of them for scraping.

> Either way if you think that this should be configurable then I'll switch to that. Maybe best to keep the current default and then just bump a patch version?

hmmm... Yeah, I guess if its a new default, it may be better to bump the minor version, or leave it as a patch version and keep it with the old default. Either way is ok I think.

zeritti (approved):
Thank you, @judahrand, for your PR. LGTM. I'd consider this change a feature rather than a patch, though.

judahrand:
> Thank you, @judahrand, for your PR. LGTM. I'd consider this change a feature rather than a patch, though.

I've updated the PR to bump a minor version instead so I think this should be good to go?

zeritti (approved):

PR #4969, opened 2024-11-04
[kube-prometheus-stack] - targetLabels support for kubelet

### LINKED ISSUE #4966: prometheus-kube-stack - support targetLabels for node-exporter and kubelet using operator

Currently we are not passing targetLabels to node-exporter serviceMonitor and also no way to create kubelet service with additional labels so that kubelet serviceMinotor can add them to targetLabels in prometheus.

Example:

`prometheus-node-exporter:
  service:
    labels:
      cluster: talos-cluster
  prometheus:
    monitor:
      targetLabels:
        - cluster
      relabelings:
        - sourceLabels: [__meta_kubernetes_endpoint_node_name]
          separator: ;
          targetLabel: nodename
          replacement: $1
          action: replace

Where as same is supported in kube-service-metrics exporter.

kube-state-metrics:
  customLabels:
    cluster: talos-cluster
  prometheus:
    monitor:
      targetLabels:
        - cluster
      relabelings:
        - sourceLabels: [__meta_kubernetes_endpoint_node_name]
          separator: ;
          targetLabel: nodename
          replacement: $1
          action: replace`

### What's your helm version?

version.BuildInfo{Version:"v3.16.2", GitCommit:"13654a52f7c70a143b1dd51416d633e1071faffb", GitTreeState:"clean", GoVersion:"go1.22.7"}

Client Version: v1.31.1 Kustomize Version: v5.4.2 Server Version: v1.31.1

### Which chart?

kube-prometheus-stack

### What's the chart version?

65.5.1

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

 helm upgrade -i prometheus . -n kube-system -f ./local/overrides.yaml

_No response_

kranthikirang:

These changes will support the kubelet to add labels for service and serviceMonitor targetLabels.

#4966 kubelet part


charts/kube-prometheus-stack/values.yaml
```diff
@@ -1464,6 +1464,11 @@ kubelet:
     additionalLabels: {}
     #  foo: bar
 
+    ## targetLabels for serviceMonitor
```
QuentinBisson, line 1467:
```suggestion
    ## defines the labels which are transferred from the associated Kubernetes Pod object onto the ingested metrics.
```

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -65,6 +65,9 @@ spec:
             {{- end }}
             - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
             - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
+            {{- if .Values.prometheusOperator.kubeletService.labels }}
```
QuentinBisson, line 68: I'm confused, reading the code of the operator, this field is in no way limited to the kubelet service. It should affect all resources created by the operator right?

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1464,6 +1464,11 @@ kubelet:
     additionalLabels: {}
     #  foo: bar
 
+    ## defines the labels which are transferred from the associated Kubernetes Pod object onto the ingested metrics.
```
QuentinBisson, line 1467:
```suggestion
    ## defines the labels which are transferred from the associated Kubernetes Service object onto the ingested metrics.
```

Sorry for the typo :D

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
QuentinBisson: I'm wondering if we could not just add an extraArgs for the operator because the value of `prometheusOperator.labels` is already taken 🤔 @jkroepke any thoughts?

kranthikirang: Now I am confused. According to the documentation you are absolutely correct. But I do not see such behavior. I see it only adds to kube-prometheus-kubelet service but no other and that is the only service seems to preserve even after uninstalling the helm chart. Double checked again now unless I am missing something.

image

to kubelet

image

jkroepke: It should be added to Secrets as well, e.g. config secrets for AM and Prometheus. In context of services, the operations manages only the kubelet service.

jkroepke: +1 for extraArgs option. 

Anyways, the use-case to put the helm release labels as options might be an interesting one in context of automatic cleanups. But I would expect that helm deletes the kubelet service + config secrets on each helm upgrade as well. Thats an dangerous approach.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1464,6 +1464,11 @@ kubelet:
     additionalLabels: {}
     #  foo: bar
 
+    ## defines the labels which are transferred from the associated Kubernetes Service object onto the ingested metrics.
+    ## https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitor
+    targetLabels: []
```
jkroepke, line 1469: What about other exporters?

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
kranthikirang: that deletion must be taken care by operator itself right. Helm has nothing to do with that.

charts/kube-prometheus-stack/values.yaml
jkroepke: Sorry, maybe I was not clear enough.

I was talking about other exporters inside kube-prometheus-stack. There are more than just kubelet:

https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/exporters

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
kranthikirang: I see the same labels are being added to created secrets 

image

image

image

kranthikirang:
I messed up with DCO check. I will create a new PR for all the exporter under kube-prometheus-stack.

PR #4973, opened 2024-11-05
[kube-prometheus-stack] - targetLabels for all exporters

### LINKED ISSUE #4966: prometheus-kube-stack - support targetLabels for node-exporter and kubelet using operator

kranthikirang:

This PR will add the support for targetLabels for all exporters in kube-prometheus-stack

- fixes #4966 kube-prometheus-stack exporters part

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -143,10 +143,8 @@ spec:
             - --web.key-file=/cert/{{ if .Values.prometheusOperator.admissionWebhooks.certManager.enabled }}tls.key{{ else }}key{{ end }}
             - --web.listen-address=:{{ .Values.prometheusOperator.tls.internalPort }}
             - --web.tls-min-version={{ .Values.prometheusOperator.tls.tlsMinVersion }}
-            {{- if .Values.prometheusOperator.extraArgs }}
-            {{- range .Values.prometheusOperator.extraArgs }}
-            - {{ . }}
-            {{- end }}
+            {{- with.Values.prometheusOperator.extraArgs }}
+            {{- tpl (toYaml .) $ | nindent 12 }}
             {{- end }}
           ports:
             - containerPort: {{ .Values.prometheusOperator.tls.internalPort }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -143,6 +143,11 @@ spec:
             - --web.key-file=/cert/{{ if .Values.prometheusOperator.admissionWebhooks.certManager.enabled }}tls.key{{ else }}key{{ end }}
             - --web.listen-address=:{{ .Values.prometheusOperator.tls.internalPort }}
             - --web.tls-min-version={{ .Values.prometheusOperator.tls.tlsMinVersion }}
+            {{- if .Values.prometheusOperator.extraArgs }}
+            {{- range .Values.prometheusOperator.extraArgs }}
+            - {{ . }}
+            {{- end }}
+            {{- end }}
           ports:
             - containerPort: {{ .Values.prometheusOperator.tls.internalPort }}
               name: https
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -65,9 +65,6 @@ spec:
             {{- end }}
             - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
             - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
-            {{- if .Values.prometheusOperator.kubeletService.labels }}
-            - --labels={{ .Values.prometheusOperator.kubeletService.labels }}
-            {{- end }}
             {{- if .Values.prometheusOperator.logFormat }}
             - --log-format={{ .Values.prometheusOperator.logFormat }}
             {{- end }}
--- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
@@ -65,6 +65,9 @@ spec:
             {{- end }}
             - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
             - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
+            {{- if .Values.prometheusOperator.kubeletService.labels }}
+            - --labels={{ .Values.prometheusOperator.kubeletService.labels }}
+            {{- end }}
             {{- if .Values.prometheusOperator.logFormat }}
             - --log-format={{ .Values.prometheusOperator.logFormat }}
             {{- end }}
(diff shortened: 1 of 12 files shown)


charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -65,6 +65,9 @@ spec:
             {{- end }}
             - --kubelet-endpoints={{ .Values.prometheusOperator.kubeletEndpointsEnabled }}
             - --kubelet-endpointslice={{ .Values.prometheusOperator.kubeletEndpointSliceEnabled }}
+            {{- if .Values.prometheusOperator.kubeletService.labels }}
+            - --labels={{ .Values.prometheusOperator.kubeletService.labels }}
```
jkroepke, line 69: As discussed with @QuentinBisson earlier `prometheusOperator.kubeletService.labels` assume, that only the `kubeletService` get additional labels. However all resources created by the operator. Reference `kubeletService` is not correct here.

I would recommend a more generic approach, e.g.:

https://github.com/prometheus-community/helm-charts/blob/f828f547380ce33ea5bd7d7b483bb573a8df1a70/charts/kube-state-metrics/templates/deployment.yaml#L71-L73

where users can pass any CLI arguments as they want.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1982,6 +2006,10 @@ kubeScheduler:
     additionalLabels: {}
     #  foo: bar
 
+    ## defines the labels which are transferred from the associated Kubernetes Service object onto the ingested metrics.
+    ## https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitor
+    targetLabels: []
```
jkroepke, line 2011: It would be make sense to allow end-users to define additional labels on that service? Otherwise, it might be pointless.

https://github.com/prometheus-community/helm-charts/blob/f828f547380ce33ea5bd7d7b483bb573a8df1a70/charts/kube-prometheus-stack/templates/exporters/kube-etcd/service.yaml#L6-L9

A value like `Values.kubeEtcd.service.labels`

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
kranthikirang: that make sense; Removed that change

charts/kube-prometheus-stack/values.yaml
kranthikirang: As per kube-prometheus-stack.labels function we can supply additional labels to services through .Values.commonLabels which will be passed to targetLabels (once they match) using serviceMonitor.

kranthikirang: So, basically instead of adding relabeling for each we just add serviceMonitor targetLabels to fetch the respective service label if they match. Would be easy to maintain overrides for multiple clusters.

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
jkroepke: @kranthikirang just to prevent an missunderstanding, but the extraArgs needs to be implemented for the operator

kranthikirang: `{{- if .Values.prometheusOperator.extraArgs }}
{{- range .Values.prometheusOperator.extraArgs }}
- {{ . }}
{{- end }}
{{- end }}`

Modified and tested. Please check now

kranthikirang: Also added an example in values file

jkroepke (commented):
LGTM.

I found one thing and please add a test case for the extra arg here: https://github.com/prometheus-community/helm-charts/blob/f828f547380ce33ea5bd7d7b483bb573a8df1a70/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml#L10 You can use the example from the values.yaml. 

Once done, I fine to merge it.

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -143,6 +143,11 @@ spec:
             - --web.key-file=/cert/{{ if .Values.prometheusOperator.admissionWebhooks.certManager.enabled }}tls.key{{ else }}key{{ end }}
             - --web.listen-address=:{{ .Values.prometheusOperator.tls.internalPort }}
             - --web.tls-min-version={{ .Values.prometheusOperator.tls.tlsMinVersion }}
+            {{- if .Values.prometheusOperator.extraArgs }}
+            {{- range .Values.prometheusOperator.extraArgs }}
+            - {{ . }}
+            {{- end }}
+            {{- end }}
```
jkroepke, line 150:
```suggestion
            {{- with.Values.prometheusOperator.extraArgs }}
            {{- tpl (toYaml .) $ | nindent 12 }}
            {{- end }}
```

That should be worked as well, or I'm wrong?

kranthikirang: Agreed. Will add above

kranthikirang:
@jkroepke  Can you please check now?

charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
```diff
@@ -143,6 +143,9 @@ spec:
             - --web.key-file=/cert/{{ if .Values.prometheusOperator.admissionWebhooks.certManager.enabled }}tls.key{{ else }}key{{ end }}
             - --web.listen-address=:{{ .Values.prometheusOperator.tls.internalPort }}
             - --web.tls-min-version={{ .Values.prometheusOperator.tls.tlsMinVersion }}
+            {{- with.Values.prometheusOperator.extraArgs }}
```
jkroepke, line 146:
```suggestion
            {{- with .Values.prometheusOperator.extraArgs }}
```

jkroepke (approved):
LGTM 👍

rouke-broersma:
We add the source labels only at federation scrape in the scrape config, not in the in-cluster metrics:

```yaml
staticConfigs:
  - targets:
    - {{ .Values.federatedDomain }}
    labels:
      metric_source: federation
      cluster: {{ .Values.clusterName }}
```

PR #4976, opened 2024-11-06
[kube-prometheus-stack] Fix Helm template error by using hasKey to ch…

traberph:
[kube-prometheus-stack] Fix Helm template error by using hasKey to check scrapeConfigNamespaceSelector presence

This line
`{{ ne .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector nil }}`

Causes following error
`error calling ne: uncomparable type map[string]interface {}: map[]`

Changing the `ne` to the `hasKey` is more robust and does not fail if the `scrapeConfigNamespaceSelector` is `{}` like in the original `values.yml`

(currently no issue open for this problem as i know)

#### Checklist

- [x ] Chart Version bumped
- [x ] Title of the PR starts with chart name

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 65.8.1
+version: 65.8.2
 appVersion: v0.77.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -253,13 +253,13 @@ spec:
   scrapeConfigSelector:
     matchLabels:
       release: {{ $.Release.Name | quote }}
-{{ else if hasKey .Values.prometheus.prometheusSpec "scrapeConfigSelector" }}
+{{ else if not (kindIs "invalid" .Values.prometheus.prometheusSpec.scrapeConfigSelector) }}
   scrapeConfigSelector: {}
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector }}
   scrapeConfigNamespaceSelector:
 {{ tpl (toYaml .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector | indent 4) . }}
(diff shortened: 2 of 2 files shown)


jkroepke:
@schnatterer Could you take a look here?

schnatterer:
Well, I can tell that the unit test still passes with the change, which is good:

```bash
$ helm unittest . -f  unittests/prometheus/scrape_config_namespace_selector_test.yaml

### Chart [ kube-prometheus-stack ] .

 PASS  test scrapeConfigNamespaceSelector       unittests/prometheus/scrape_config_namespace_selector_test.yaml

Charts:      1 passed, 1 total
Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshot:    0 passed, 0 total
Time:        56.225257ms
```

But TBH I did not understand which value caused the issue. Could you elaborate on that with an example or even a unit test, @traberph?

traberph:
Thanks for the fast reply.

@schnatterer  I stumbled about this error while upgrading the minor version of the helm chart in our k8s cluster.

The error we get is caused by the comparison of {} against nil as I understand.

I am currently on the road with limited Internet access.
I will give you more details later and try to setup a test.

ticup:
currently trying to install the kube-prometheus-stack is giving that error for us, even if we provide a value for it in a values file:

````
Error: INSTALLATION FAILED: template: kube-prometheus-stack/templates/prometheus/prometheus.yaml:262:11: executing "kube-prometheus-stack/templates/prometheus/prometheus.yaml" at <ne .Values.prometheus.prometheusSpec.scrapeConfigNamespaceSelector nil>: error calling ne: uncomparable type map[string]interface {}: map[]
````

**Comment by xannz** [comment]
Running into the same problem when trying the latest release. It seems to be because the helm version used by our argocd installation is to low. (Related issue: https://github.com/prometheus-community/helm-charts/issues/4958)

**Comment by traberph** [comment]
> (Related issue: #4958)

Upgrading Helm did fix the problem for us as well

Thanks for pointing out

**Comment by traberph** [comment]
@schnatterer I just tested different versions of Helm with a similar scenario.

Using `ne` is backward compatible to version `3.11`.
Using `hasKey` is compatible back to version `3.7` (I don't test older versions)

Template I used

```
{{- if ne .Values.test0 nil }}
{{- end }} 
{{- if hasKey .Values "test0" }}
{{- end }} 
{{- if ne .Values.test1 nil }}
{{- end }}
{{- if hasKey .Values "test1" }}
{{- end }} 
```

And the values

```
test0: {}
```

jkroepke:
Thats I prefer the kindIs invalid condition.

traberph:
@jkroepke I changed the lines again.
It should be compatible back to Helm 3.7

jkroepke (approved):

PR #5001, opened 2024-11-18
[prometheus] pdb template values

mentlak0:

- fixes:
- FIX to PDB template. Here we are seeing maxUnavailable defaulting to 1. If we try to only spec minAvailable, maxUnavailable is templating due to default value. K8s does not allow both of these specs. 
- This PR sees a change to allow specific values without issue. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/templates/pdb.yaml
+++ b/charts/prometheus/templates/pdb.yaml
@@ -11,5 +11,10 @@ spec:
   selector:
     matchLabels:
       {{- include "prometheus.server.matchLabels" . | nindent 6 }}
-  {{- toYaml $pdbSpec | nindent 2 }}
+  {{- if $pdbSpec.minAvailable }}
+  minAvailable: {{ $pdbSpec.minAvailable }}
+  {{- end }}
+  {{- if $pdbSpec.maxUnavailable }}
+  maxUnavailable: {{ $pdbSpec.maxUnavailable }}
+  {{- end }}
 {{- end }}
--- a/charts/prometheus/templates/pdb.yaml
+++ b/charts/prometheus/templates/pdb.yaml
@@ -22,5 +22,5 @@ spec:
   {{- end }}
   {{- if hasKey $pdbSpec "unhealthyPodEvictionPolicy" }}
   unhealthyPodEvictionPolicy: {{ $pdbSpec.unhealthyPodEvictionPolicy }}
-  {{- end }}}}
+  {{- end }}
 {{- end }}
--- a/charts/prometheus/templates/pdb.yaml
+++ b/charts/prometheus/templates/pdb.yaml
@@ -11,10 +11,16 @@ spec:
   selector:
     matchLabels:
       {{- include "prometheus.server.matchLabels" . | nindent 6 }}
-  {{- if $pdbSpec.minAvailable }}
+  {{- if not (or (hasKey $pdbSpec "minAvailable") (hasKey $pdbSpec "maxUnavailable")) }}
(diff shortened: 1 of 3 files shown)


zeritti:
> @gianrubio @naseemkullah @Xtigyro @zanhsieh @zeritti
> 
> Hello,
> 
> Would appreciate a review on this. Our prom instance keeps jumping nodes and I want to add a correct PDB

Please, sign your commits so that DCO check can pass.

charts/prometheus/templates/pdb.yaml
```diff
@@ -11,5 +11,10 @@ spec:
   selector:
     matchLabels:
       {{- include "prometheus.server.matchLabels" . | nindent 6 }}
-  {{- toYaml $pdbSpec | nindent 2 }}
+  {{- if $pdbSpec.minAvailable }}
+  minAvailable: {{ $pdbSpec.minAvailable }}
+  {{- end }}
+  {{- if $pdbSpec.maxUnavailable }}
+  maxUnavailable: {{ $pdbSpec.maxUnavailable }}
+  {{- end }}
```
zeritti, line 19: We can make PDB a bit friendlier to users who would like to set `minAvailable` since they have to set two fields: unset default `maxUnavailable` and then set `minAvailable`. I would suggest a change in this respect - commenting the current default out in the values file and rewriting the template while setting the current default. `unhealthyPodEvictionPolicy`should not be discarded.

```yaml
  {{- if not (or (hasKey $pdbSpec "minAvailable") (hasKey $pdbSpec "maxUnavailable")) }}
  maxUnavailable: 1
  {{- end }}
  {{- if hasKey $pdbSpec "minAvailable" }}
  minAvailable: {{ $pdbSpec.minAvailable }}
  {{- end }}
  {{- if hasKey $pdbSpec "maxUnavailable" }}
  maxUnavailable: {{ $pdbSpec.maxUnavailable }}
  {{- end }}
  {{- if hasKey $pdbSpec "unhealthyPodEvictionPolicy" }}
  unhealthyPodEvictionPolicy: {{ $pdbSpec.unhealthyPodEvictionPolicy }}
  {{- end }}
```

We'd have to add a comment in the values stating that default is set to `maxUnavailable=1` if neither `minAvailable` nor `maxUnavailable` is set. What dou you think?

zeritti (changes requested):

charts/prometheus/templates/pdb.yaml
zeritti: Thank you, @mentlak0. Please, comment out the default value in the values file L456.

mentlak0: @zeritti Done

charts/prometheus/templates/pdb.yaml
```diff
@@ -11,5 +11,16 @@ spec:
...
+  {{- if hasKey $pdbSpec "minAvailable" }}
+  minAvailable: {{ $pdbSpec.minAvailable }}
+  {{- end }}
+  {{- if hasKey $pdbSpec "maxUnavailable" }}
+  maxUnavailable: {{ $pdbSpec.maxUnavailable }}
+  {{- end }}
+  {{- if hasKey $pdbSpec "unhealthyPodEvictionPolicy" }}
+  unhealthyPodEvictionPolicy: {{ $pdbSpec.unhealthyPodEvictionPolicy }}
+  {{- end }}}}
```
zeritti, line 25:
```suggestion
  {{- end }}
```

PR #5021, opened 2024-11-25
[prometheus-postgres-exporter] Don't set initialDelaySeconds to the default values

milesarmstrong:

Stop setting `livenessProbe.initialDelaySeconds` and `readinessProbe.initialDelaySeconds` to the Kubernetes defaults of `0` in `values.yaml`

Setting these values to the Kubernetes defaults causes permanent diffs in GitOps tooling, e.g. ArgoCD (see https://github.com/argoproj/argo-cd/issues/8833 for an example) as defaults are pruned when reading back the resource from the API.

#### Checklist

- [x] Chart Version bumped


zeritti (approved):
Thank you, @milesarmstrong, for your PR, LGTM.

PR #5038, opened 2024-12-04
[kube-prometheus-stack] Add node-exporter attachMetadata configuration

daviderli614:
Declare the attachMetadata configuration in the `values.yaml` of kube-prometheus-stack, so that do not have to look for it in [node-exporter helm-chart](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-node-exporter/values.yaml#L204).

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 66.3.0
+version: 66.3.1
 appVersion: v0.78.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


GMartinez-Sisti:
Hi @daviderli614, please take a look at the checklist in the PR description. You're missing the chart bump.

daviderli614:
Thank you for your review, I changed the version from 66.3.0 to 66.3.1.
@GMartinez-Sisti

jkroepke (changes requested):
I don't get the reason for that change. 

We depend on node-export chart. If there changes on that are already done, we also benefit from it. We should not redeclare  a default value.

I guess the root cause is that kube-prometheus-stack is not used the lastest version of that chart.

daviderli614:
Oh, I found that I need to look for this field in the node-exporter helm chart. Showing the definition here helps the user configure it. We can close this pr, Thanks.

jkroepke:
@daviderli614 that makes sense to me, but than add the value as a comment. like the lines before.

jkroepke (approved):

GMartinez-Sisti (approved):

PR #5051, opened 2024-12-12

schnatterer:

```
failed to create typed patch object (..): .spec.automountServiceAccountToken: field not declared in schema
```

```
prometheus:
  prometheusSpec:
    automountServiceAccountToken: null
```

This is the same as #4906 only for the new field 
`prometheus.prometheusSpec.automountServiceAccountToken` that was introduced to the Prometheus CRD in 34b499cf2a368cab94d5c1b1a7784a7b6cef88df and then added to the chart in ff10663d1c9fe327afccbf70698649621761079e, which leads to the incompatibility.

@jkroepke Could I ask you again to review this?

#### Checklist

- [x] Chart Version bumped


charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -17,7 +17,9 @@ metadata:
 {{ toYaml .Values.prometheus.annotations | indent 4 }}
 {{- end }}
 spec:
+{{- if not (eq .Values.prometheus.prometheusSpec.automountServiceAccountToken nil) }}
```
jkroepke, line 20: If I remember from last PR, this eq nil check require a minimum helm version?

schnatterer: You're right! #4976 was the one about needing Helm 3.11.
I only remembered the subchart issue in #4818 😬
Compatibility issues everywhere!

Fixed in 7d8c6d1c.

jkroepke:
Seems like dashboards needs to be generated first.

jkroepke (approved):
LGTM

PR #5050, opened 2024-12-12
[kube-prometheus-stack] feature(kubelet): Add trackTimestampsStaleness

Breee:
#### What this PR does / why we need it

This PR enables `trackTimestampsStaleness` for the cadvisor  path to avoid stale Metrics provided by the cadvisor. 

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes # 

#### Special notes for your reviewer

@SuperQ  suggested this and might want to add comments. 
slack conversation: https://cloud-native.slack.com/archives/C167KFM6C/p1733917170837469 

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #5063, opened 2024-12-15
[kube-prometheus-stack] Feat: Improve cAdvisor metrics scrape

SuperQ:
Improve the collection of cAdvisor metrics by adjusting the scrape interval to match the kubelet hardcoded minimum housekeeping interval.
* Set the cAdvisor, and resource, interval to 10s by default.
* Use the user configured a kubelet metrics interval if not default.
* Enforce honorTimestamps for cAdvisor and resource if timetamp staleness is enabled.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
+++ b/charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
@@ -61,6 +61,8 @@ spec:
     path: /metrics/cadvisor
     {{- if .Values.kubelet.serviceMonitor.interval }}
     interval: {{ .Values.kubelet.serviceMonitor.interval }}
+    {{- else }}
+    interval: {{ .Values.kubelet.serviceMonitor.cAdvisorInterval }}
     {{- end }}
     {{- if .Values.kubelet.serviceMonitor.proxyUrl }}
     proxyUrl: {{ .Values.kubelet.serviceMonitor.proxyUrl }}
@@ -69,7 +71,11 @@ spec:
     scrapeTimeout: {{ .Values.kubelet.serviceMonitor.scrapeTimeout }}
     {{- end }}
     honorLabels: {{ .Values.kubelet.serviceMonitor.honorLabels }}
+    {{- if .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
+    honorTimestamps: true
+    {{- else }}
     honorTimestamps: {{ .Values.kubelet.serviceMonitor.honorTimestamps }}
+    {{- end }}
     trackTimestampsStaleness: {{ .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
     {{- include "kube-prometheus-stack.kubelet.authConfig" . | indent 4 }}
 {{- if .Values.kubelet.serviceMonitor.cAdvisorMetricRelabelings }}
@@ -112,6 +118,8 @@ spec:
     path: {{ .Values.kubelet.serviceMonitor.resourcePath }}
     {{- if .Values.kubelet.serviceMonitor.interval }}
     interval: {{ .Values.kubelet.serviceMonitor.interval }}
+    {{- else }}
+    interval: {{ .Values.kubelet.serviceMonitor.resourceInterval }}
     {{- end }}
     {{- if .Values.kubelet.serviceMonitor.proxyUrl }}
     proxyUrl: {{ .Values.kubelet.serviceMonitor.proxyUrl }}
@@ -120,7 +128,11 @@ spec:
     scrapeTimeout: {{ .Values.kubelet.serviceMonitor.scrapeTimeout }}
     {{- end }}
     honorLabels: {{ .Values.kubelet.serviceMonitor.honorLabels }}
+    {{- if .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
+    honorTimestamps: true
+    {{- else }}
     honorTimestamps: {{ .Values.kubelet.serviceMonitor.honorTimestamps }}
+    {{- end }}
     trackTimestampsStaleness: {{ .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
     {{- include "kube-prometheus-stack.kubelet.authConfig" .  | indent 4 }}
 {{- if .Values.kubelet.serviceMonitor.resourceMetricRelabelings }}
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 67.1.0
+version: 67.2.0
 appVersion: v0.79.0
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1324,7 +1324,7 @@ kubelet:
 
     ## If true, defines whether Prometheus tracks staleness of the metrics that have an explicit timestamp present in scraped data. Has no effect if `honorTimestamps` is false.
     ## We recommend enabling this if you want the best possible accuracy for container_ metrics scraped from cadvisor.
-    ##
+    ## For more details see: https://github.com/prometheus-community/helm-charts/pull/5063#issuecomment-2545374849
     trackTimestampsStaleness: true
 
     ## SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
@@ -1362,10 +1362,6 @@ kubelet:
     ##
     insecureSkipVerify: true
 
-    ## Enable scraping /metrics/cadvisor from kubelet's service
-    ##
-    cAdvisor: true
-
     ## Enable scraping /metrics/probes from kubelet's service
     ##
     probes: true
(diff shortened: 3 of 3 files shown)


SuperQ:
Followup to #5050 (CC @Breee).

For reference, this is the [hardcoded housekeeping interval](https://github.com/kubernetes/kubernetes/blob/cb93d6ee69b8d4ca8701336e4f7cb278751f34e4/pkg/kubelet/cadvisor/cadvisor_linux.go#L57-L59).

Note, this will likely cause `PrometheusDuplicateTimestamps` to fire for most users. We should probably remove/disable this alert.

We could disable it by default in the values.yaml, or we could remove it from the upstream kube-prometheus alert mixin.

jkroepke:
Hey @SuperQ 

I feel the information about trackTimestampsStaleness and cAdvisor interesting for some users (and for maintainers very important).

WDYT about leaving a note or link to the `trackTimestampsStaleness` that explain that a little bit? The link to the slack discussion will be unavailable in 90-180 days. one or two sentences are fine.

----

> Note, this will likely cause PrometheusDuplicateTimestamps to fire for most users. We should probably remove/disable this alert.


Does this include a lot of warnings `Error on ingesting samples with different value but same timestamp` on the prometheus logs as well? If so, I see some incoming GitHub issues that users may complain about this. :/

From my point of view, it should be sufficient to exclude the kubelet scrape job from PrometheusDuplicateTimestamps alert, or I'm wrong?

SuperQ:
> Does this include a lot of warnings `Error on ingesting samples with different value but same timestamp`

It shouldn't. This is more related to ingesting the same timestamps more than once. Maybe this is a Prometheus metric "bug", that we're counting the same condition for multiple reasons. I haven't investigated the code for this.

SuperQ:
Sure, so here's the details on `trackTimestampsStaleness`.

Normal metrics exposed to Prometheus do not include timestamps. Metrics are supposed to be [produced "on demand"](https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling), so the data timestamp is the same as the scrape timestamp. 

cAdvisor does not follow this best practice. Instead cAdvisor has its own internal data gathering loop, it exposes timestamps on `container_...` metrics for the exact time in which the data was collected.

Prometheus scrapes have millisecond time accuracy and are intentionally splayed over time avoid correlated network traffic. But user queries _do_ have a correlated timestamp, and systems like Grafana will snap the query to a round number like 00:30:00. This means that when you query at 00:30:00, but the metric you want was scraped at 00:29:55.231, Prometheus has to "look back" some amount of time to find that sample.

This lookback defaults to 5 minutes in order to handle a variety of scrape interval use cases. The down side to lookback is that if a metric is removed from an endpoint, or a target is down, Prometheus would continue to show data up to 5 minutes after it really existed. This was confusing for users.

In Prometheus 2.0 we improved on this situation by introducing a [staleness handler](https://prometheus.io/blog/2017/11/08/announcing-prometheus-2-0/#staleness-handling). When a metric label combo is removed from a scrape it will be marked as "stale" in the TSDB.

Prometheus, due to historical reasons, handles metrics that include the optional timestamp field without this staleness marker. 

Because cAdvisor has its own internal data gathering loop, it exposes timestamps on `container_...` metrics for the exact time in which the data was collected. This allows the data to be more accurate since it can be many seconds between when the data is collected and when Prometheus gets the data.

But now, without the staleness markers, Prometheus will continue to apply the lookback of up to 5 minutes to find metrics. This can be very confusing for users as they will "see" cAdvisor metrics that don't exist anymore. The common example of this is metrics like `container_memory_rss`. What users will see is that for 5 minutes after a Pod is deleted, the memory RSS will continue to exist.

The new track timestamp staleness feature was introduced to fix this behavior. Now with `trackTimestampsStaleness` enabled metrics like `container_memory_rss` will correctly be removed when the Pod is removed. 

The previous workaround was to disable "honorTimestamps" for cAdvisor. This worked to fix the staleness handling issue. But it causes the data to be slightly inaccurate as cAdvisor's internal loop is not in sync with Prometheus. So in order to fully enable `trackTimestampsStaleness`, we also need to force enable the `honorTimestamps` feature.

The second half of this change has to do with the fact that cAdvisor does not have a strict timer for internal data collection. It uses [variable timing from 10 to 15](https://github.com/kubernetes/kubernetes/blob/cb93d6ee69b8d4ca8701336e4f7cb278751f34e4/pkg/kubelet/cadvisor/cadvisor_linux.go#L57-L59) seconds in Kubernetes. So in order to capture every sample produced by cAvisor, we need to scrape slightly faster than the minimum housekeeping interval.

charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
```diff
@@ -69,7 +71,11 @@ spec:
     scrapeTimeout: {{ .Values.kubelet.serviceMonitor.scrapeTimeout }}
     {{- end }}
     honorLabels: {{ .Values.kubelet.serviceMonitor.honorLabels }}
+    {{ if .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
+    honorTimestamps: true
+    {{- else }}
     honorTimestamps: {{ .Values.kubelet.serviceMonitor.honorTimestamps }}
+    {{- end }}
     trackTimestampsStaleness: {{ .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
```
jkroepke, line 79: Could add the link to the very detailed comment (https://github.com/prometheus-community/helm-charts/pull/5063#issuecomment-2545374849) to the corresponding trackTimestampsStaleness property in the values.yaml?

charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
```diff
@@ -95,7 +101,11 @@ spec:
     scrapeTimeout: {{ .Values.kubelet.serviceMonitor.scrapeTimeout }}
     {{- end }}
     honorLabels: {{ .Values.kubelet.serviceMonitor.honorLabels }}
+    {{ if .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
```
jkroepke, line 104: Since `trackTimestampsStaleness` is not set for the probe endpoint, the condition feels pointless.

charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
```diff
@@ -69,7 +71,11 @@ spec:
...
```
jkroepke, line 74:
```suggestion
    {{- if .Values.kubelet.serviceMonitor.trackTimestampsStaleness }}
```

charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
```diff
@@ -112,6 +122,8 @@ spec:
     path: {{ .Values.kubelet.serviceMonitor.resourcePath }}
```
jkroepke, line 122: If cadvisor and /metrics/resource shares the same logic, the `if .Values.kubelet.serviceMonitor.trackTimestampsStaleness` should be added to the resource metrics endpoint, not to the probes endpoint.

jkroepke (changes requested):
Have some question here and I guess one go template condition is set to the wrong endpoint.

Thanks for the detailed comment. The link should be added to `kubelet.serviceMonitor.trackTimestampsStaleness` in the value.yaml!

charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml
SuperQ: Oops, that was supposed to be on resources.

jkroepke:
@SuperQ what about the PrometheusDuplicateTimestamps alert that you mention earlier? Should it be covered in this or an follow-up PR, if users complain about?

SuperQ:
I've been testing the 10s interval on one of my k3s test clusters all day and haven't been able to reproduce the duplicate timestamp issue there. Perhaps there is something wrong with my production setup that isn't normal. Maybe it's a cAdvisor under heavy load problem? I'm not sure at this point.

jkroepke (approved):
Thanks a lot!

festeveira:
Hello, after these changes I started repeatedly getting the following log messages in the prometheus container:

`time=2024-12-18T17:08:26.335Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.34:10250/metrics/cadvisor num_dropped=2646
time=2024-12-18T17:08:28.998Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.32:10250/metrics/cadvisor num_dropped=3396
time=2024-12-18T17:08:33.063Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.31:10250/metrics/cadvisor num_dropped=3172
time=2024-12-18T17:08:34.711Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.33:10250/metrics/cadvisor num_dropped=4662
time=2024-12-18T17:08:36.325Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.34:10250/metrics/cadvisor num_dropped=2437
time=2024-12-18T17:08:39.061Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.32:10250/metrics/cadvisor num_dropped=4196
time=2024-12-18T17:08:43.134Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.31:10250/metrics/cadvisor num_dropped=1796
time=2024-12-18T17:08:44.685Z level=WARN source=scrape.go:1877 msg="Error on ingesting out-of-order samples" component="scrape manager" scrape_pool=serviceMonitor/monitoring/prometheus-agent-kube-prom-kubelet/1 target=https://10.134.60.33:10250/metrics/cadvisor num_dropped=3620
`
Is this expected?

levirosal:
Hi @festeveira, I'm also having problems with this new flag receiving the same warning that @jkroepke mention. I opened an issue here in the repository.

Issue: https://github.com/prometheus-community/helm-charts/issues/5225

Warning:

`time=2025-01-21T04:52:14.768Z level=WARN source=scrape.go:1885 msg="Error on ingesting samples with different value but same timestamp" component="scrape manager" scrape_pool=serviceMonitor/observability-system/cloudwatch-exporter-prometheus-cloudwatch-exporter/0 target=http://172.30.9.19:9106/metrics num_dropped=13`

Cc.: @SuperQ

emilbillberg:
I am having these issues as well `Error on ingesting samples with different value but same timestamp` after upgrading

PR #5069, opened 2024-12-16
[prometheus-postgres-exporter] Use non default database in multi-targets

thomas-may:

This PR introduces the ability to **select the database to connect to in a multi-target configuration** and have the servicemonitor correctly scrape databases that do not match username.

The multi-target configuration section is completed with the new optional parameter `databaseName` as follows:

```yaml
serviceMonitor:
  enabled: true
   # Use multipleTarget mode
  multipleTarget:
    enabled: true
    targets:
      - endpoint: pg01.local
        name: pg01
      - endpoint: pg02.local
        name: pg02
        databaseName: mydb     # <== New optional parameter for selecting database
```

This configuration produces the following servicemonitor resource:

```yaml
kind: ServiceMonitor
# ... [redacted]
spec:
  endpoints:
  - metricRelabelings:
    # ... [redacted]
    params:
      auth_module:
      - client.pg01
      target:
      - pg01.local:5432
    path: /probe
    port: http
  - metricRelabelings:
    # ... [redacted]
    params:
      auth_module:
      - client.pg02
      target:
      - pg02.local:5432/mydb     # <== Target references the database in its path
    path: /probe
    port: http
  # ... [redacted]
```

The multi-target feature is currently in beta, but such configuration would be needed for any users relying on a Postgres user dedicated to monitoring purposes, without a database matching its name.

Currently, a workaround is to keep the servicemonitor of this chart disabled and define a custom servicemonitor.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-postgres-exporter/Chart.yaml
+++ b/charts/prometheus-postgres-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "v0.16.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 6.7.1
+version: 6.8.0
 home: https://github.com/prometheus-community/postgres_exporter
 sources:
 - https://github.com/prometheus-community/postgres_exporter
--- a/charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
+++ b/charts/prometheus-postgres-exporter/templates/servicemonitor.yaml
@@ -47,7 +47,11 @@ spec:
   {{- end }}
       params:
         target:
+        {{- if .databaseName }}
+          - {{ .endpoint }}:{{ .port | default 5432 }}/{{ .databaseName | default "" }}
+        {{- else }}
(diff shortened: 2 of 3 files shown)


charts/prometheus-postgres-exporter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "v0.16.0"
 description: A Helm chart for prometheus postgres-exporter
 name: prometheus-postgres-exporter
-version: 6.7.1
+version: 6.7.2
```
zanhsieh, line 5: Better bump to 6.8.0 since a new feature introduced.

```suggestion
version: 6.8.0
```

zanhsieh (changes requested):

charts/prometheus-postgres-exporter/Chart.yaml
thomas-may: Thank you for the review.
The version update to `6.8.0` seems indeed more appropriate.

zanhsieh (approved):

PR #5088, opened 2024-12-23
[prometheus-mongodb-exporter] Allow to configure pod security context

YvesZelros:

Allow to configure pod security context

Add value _**podSecurityContext**_ as it's present on most of other chart as _prometheus-postgres-exporter_ and _prometheus-elasticsearch-exporter_

Don't set default value to avoid breaking change.

@steven-sheehy
@zeritti


zeritti (changes requested):
Thank you, @YvesZelros, for your PR. Will you please add a CI test case in the `ci` directory for the new field?

zeritti (approved):

PR #5099, opened 2024-12-26
[prometheus-rabbitmq-exporter] Create secret to store credentials when provided

jchanam:
Connection credentials should be stored securely using secrets. Right now, the chart allows to provide your own secret, but if the credentials are given as values to the chart, they are exposed in the deployment env variables.

This PR adds a new secret, that will be created only if the user or password is provided and if the existing user/password secret is not given.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-rabbitmq-exporter/templates/secret.yaml
+++ b/charts/prometheus-rabbitmq-exporter/templates/secret.yaml
@@ -1,0 +1,22 @@
+{{- if or (and (.Values.rabbitmq.password) (not .Values.rabbitmq.existingPasswordSecret)) (and (.Values.rabbitmq.user) (not .Values.rabbitmq.existingUserSecret)) }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "prometheus-rabbitmq-exporter.fullname" . }}
+  labels:
+    app: {{ template "prometheus-rabbitmq-exporter.name" . }}
+    chart: {{ template "prometheus-rabbitmq-exporter.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  {{- if .Values.additionalLabels }}
+{{ toYaml .Values.additionalLabels | indent 4 }}
+  {{- end }}
+type: Opaque
+data:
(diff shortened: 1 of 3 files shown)


charts/prometheus-rabbitmq-exporter/templates/secret.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if or (and (.Values.rabbitmq.password) (not .Values.rabbitmq.existingPasswordSecret)) (and (.Values.rabbitmq.user) (not .Values.rabbitmq.existingUserSecret)) }}
```
desaintmartin, line 1: If `.Values.rabbitmq.password` is set but not `.Values.rabbitmq.existingPasswordSecret`, secret will not be created and Pod will never start. We should have a solution. Maybe we should prevent the two being set at the same time, or we align the condition in the two files, or we simply "ignore" the latter like it's done before this PR.

desaintmartin (changes requested):
Thanks for the contribution! This one has been on my list for a while but I never had the time to do it.

charts/prometheus-rabbitmq-exporter/templates/secret.yaml
jchanam: Hi!

Yes, you're right. If existing password secret is set, then we don't need to create the secret, as we'll be using the one provided in the values.

On that case, the secret where we're reading the value for the env is also the one provided, so I think the pod will be created and running.

This is an example of the env for the deployment on that case:

```yaml
          env:
            - name: RABBIT_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: "existingsecret"
                  key: "password"
            - name: RABBIT_USER
              valueFrom:
                secretKeyRef:
                  name: "existingsecret"
                  key: "username"
            - name: RABBIT_URL
              value: my-url.com
```

jchanam: About the options you mentioned, I followed what it was done before. If the existing secret is given, the password or user is ignored and the secret used.

desaintmartin: I am sorry, the diff bugged me, you are right! To be sure for the future, I'll add a few tests in a next PR.

desaintmartin (approved):

Issue #4972, opened 2024-11-05
[kube-prometheus-stack] Dashboards broken due to missing image, container labels


telmich:
### Describe the bug a clear and concise description of what the bug is.

Using a k8s 1.31 cluster with kube-prometheus-stack 65.5.1, cri-o-1.31.0 many dashboards that describe pod workloads are broken: memory/cpu/network utilisation is not shown anymore.

This bug might have been introduced in an earlier version, however above versions are confirmed in our cluster.

### What's your helm version?

version.BuildInfo{Version:"v3.16.2", GitCommit:"", GitTreeState:"", GoVersion:"go1.23.2"}

### What's your kubectl version?

kubectl version                        Client Version: v1.31.1 Kustomize Version: v5.4.2 Server Version: v1.31.1

### Which chart?

kube-prometheus-stack

### What's the chart version?

65.5.1

### What happened?

The dashboard, such as "Kubernetes / Compute Resources / Namespace (Pods)" shows data for most metrics.

When we change the query, for instance from "Memory Usage (w/o cache)" :

from

```
sum(container_memory_working_set_bytes{job="kubelet", metrics_path="/metrics/cadvisor", cluster="$cluster", namespace="$namespace", container!="", image!=""}) by (pod)
```

to

```
sum(container_memory_working_set_bytes{job="kubelet", metrics_path="/metrics/cadvisor", cluster="$cluster", namespace="$namespace"}) by (pod)
```
it works again.

In prometheus we can see that the metric does not have the image or container label.

### What you expected to happen?

The dashboards to work

### How to reproduce it?

See above

### Enter the changed values of values.yaml?

-

### Enter the command that you execute and failing/misfunctioning.

-

### Anything else we need to know?

There are other related reports on github about the labels missing, not sure if they are related to this issue:

https://github.com/kubernetes/kubernetes/issues/111077

https://github.com/k0sproject/k0s/issues/2187

kjg-ungleich:
i have same issue.

SteamMachinist:
I have exactly the same issue and same behavior when changing the query

telmich:
I've checked the dashboards a couple of days ago and the image !=.. and container != labels seem to be hardcoded (not controllable from the helm chart).

As so many people seem to be hit by this over and over again, i wonder if making these particular labels optional as part of the helm chart makes sense.

I also wonder if any of the helm chart developers is reading this?

telmich:
[...]

PR #5130, opened 2025-01-10
[kube-prometheus-stack] Improve various metric relabelings

SuperQ:

Update API metric relablings for Prometheus 3.0
    
Prometheus 3.0 normalizes bucket float values to include `.0` for
integers. Update the
* Add additional noisy apiserver metrics to the bucket filter.
  * `apiserver_request_sli_duration_seconds_bucket`
  * `apiserver_request_slo_duration_seconds_bucket`
  * `etcd_request_duration_seconds_bucket`
* Improve regex match to explicitly match `.`.

Improve cAdvisor metric filtering

Add new metricRelablings to the cAdvisor filter to reduce
default cardinality by eliminating noisy metrics.
* Drop container_memory_failures_total{scope="hierarchy"} metrics, we only need the container scope.
* Drop container_network_... metrics that match various interfaces that correspond to CNI and similar interfaces.

Add kubelet default metricRelabings

Reduce bucket cardinality of kubelet storage operations.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1417,6 +1417,17 @@ kubelet:
       - sourceLabels: [__name__]
         action: drop
         regex: 'container_(file_descriptors|tasks_state|threads_max)'
+      # Drop container_memory_failures_total{scope="hierarchy"} metrics,
+      # we only need the container scope.
+      - sourceLabels: [__name__, scope]
+        action: drop
+        regex: 'container_memory_failures_total;hierarchy'
+      # Drop container_network_... metrics that match various interfaces that
+      # correspond to CNI and similar interfaces. This avoids capturing network
+      # metrics for host network containers.
+      - sourceLabels: [__name__, interface]
+        action: drop
+        regex: 'container_network_.*;(cali|cilium|cni|lxc|nodelocaldns|tunl).*'
       # Drop container spec metrics that overlap with kube-state-metrics.
       - sourceLabels: [__name__]
         action: drop
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -1278,7 +1278,7 @@ kubeApiServer:
     metricRelabelings:
       # Drop excessively noisy apiserver buckets.
       - action: drop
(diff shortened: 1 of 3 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -1414,10 +1414,25 @@ kubelet:
...
+      # correspond to CNI and similar interfaces. This avoids capturing network
+      # metrics for host network containers.
+      - sourceLabels: [__name__, interface]
+        action: drop
+        regex: 'container_network_.*;(cali|cilium|cni|lxc|nodelocaldns|tunl).*'
       # Drop container spec metrics that overlap with kube-state-metrics.
       - sourceLabels: [__name__]
         action: drop
         regex: 'container_spec.*'
+      # Drop POD / non-container metrics where we only care about container level metrics.
```
jkroepke, line 1432: I'm not sure about this. There specific, but unknown condition where is the cadvisor does report container metrics without container metric.

There are various issues like https://github.com/prometheus-community/helm-charts/issues/4972 that indicate that situation. Ref: https://github.com/kubernetes/kubernetes/issues/89903

I found a old comment from you https://github.com/k3s-io/k3s/issues/5891#issuecomment-1855885837 that indicates that label without container should be dropped always. Not sure about k3s, but some users report this behavior on AWS EKS as well.

In case, we agree to not support container metrics without labels, it should be more visible documented.

SuperQ: Sure, I can remove this one. Having these extra metrics makes `sum without (container)` more complicated, but people have been living with this for a while.

jkroepke: I would post-pone this clarify it in a distinct PR/Issue. Grafana Dashboards are not working well without container labels.

SuperQ: I had already removed this part of the change.

jkroepke (commented):
I looked through the Alerts and Dashboard used by kube-prometheus-stack and I none of the removed metrics are used everywhere. 

However, based on https://xkcd.com/1172/ I would bump the major version und document, that metrics listend below are dropped by default.

jkroepke (approved):
Thanks a lot!

PR #5132, opened 2025-01-10
[prometheus-windows-exporter] Extract the prometheus.io/scrape service annotation

### LINKED ISSUE #5129: [prometheus-windows-exporter] Unable to disable `prometheus.io/scrape` annotation on the service

The [template file](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-windows-exporter/templates/service.yaml) for the Service sets the `prometheus.io/scrape` annotation unless `promtheus.monitor` or `prometheus.podMonitor` are enabled.

I don't want the annotation, nor do I want a PodMonitor or ServiceMonitor.

### What's your helm version?

v3.16.4

v1.32.0

### Which chart?

prometheus-windows-exporter

### What's the chart version?

0.7.1

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

anything. helm template

_No response_

petewall:

This change moves the `prometheus.io/scrape` service annotation into the values.yaml file where it can be disabled if not desired. 

Also add comments in the values.yaml file for the service section.

All of these changes make this match how things look and work in the Node Exporter helm chart.

- fixes #5129 

My testing to ensure that things behave correctly:
```bash
# Default
helm template test . | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: prometheus.io/scrape: "true"

# Extra annotations
helm template test . --set 'service.annotations.extra=1234' | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: prometheus.io/scrape: "true", extra: "1234"

# Disabling the prometheus.io/scrape annotation
helm template test . --set 'service.annotations.prometheus\.io/scrape=null' | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: null

# ServiceMonitor still removes the prometheus.io/scrape annotation
helm template test . --set 'prometheus.monitor.enabled=true' | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: {}

# PodMonitor still removes the prometheus.io/scrape annotation
helm template test . --set 'prometheus.podMonitor.enabled=true' | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: {}

# PodMonitor with extra works
helm template test . --set 'prometheus.podMonitor.enabled=true' --set 'service.annotations.extra=1234' | yq 'select(.kind == "Service") | .metadata.annotations'
# Result: extra: 1234
```

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/prometheus-windows-exporter/Chart.yaml
+++ b/charts/prometheus-windows-exporter/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 0.7.2
+version: 0.8.0
 appVersion: 0.29.2
 home: https://github.com/prometheus-community/windows_exporter/
 sources:
--- a/charts/prometheus-windows-exporter/Chart.yaml
+++ b/charts/prometheus-windows-exporter/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 0.7.1
+version: 0.7.2
 appVersion: 0.29.2
 home: https://github.com/prometheus-community/windows_exporter/
 sources:
--- a/charts/prometheus-windows-exporter/templates/service.yaml
+++ b/charts/prometheus-windows-exporter/templates/service.yaml
@@ -7,11 +7,7 @@ metadata:
     {{- include "prometheus-windows-exporter.labels" $ | nindent 4 }}
   {{- with .Values.service.annotations }}
   annotations:
-    {{- if or $.Values.prometheus.monitor.enabled $.Values.prometheus.podMonitor.enabled }}
-      {{- unset . "prometheus.io/scrape" | toYaml | nindent 4 }}
-    {{- else }}
-      {{- toYaml . | nindent 4 }}
-    {{- end }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
   type: {{ .Values.service.type }}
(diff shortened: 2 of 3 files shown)


jkroepke:
Hi @petewall 

I understand this kind of issue and the PR how you try to resolve it. It works in context of this chart, but it wont work, if the chart is embedded as sub chart, like in k8s-monitoring-helm.

There is a known helm issue, that removing the annotation via `service.annotations.prometheus\.io/scrape=null` does not work. Please keep that in mind and re-evalulate, if the proposed solution is still valid in context of k8s-monitoring-helm

petewall:
You're correct in that with Helm today, I can't "null" the value to remove it. But I can override it by setting this:
```
service:
  annotations:
    prometheus.io/scrape: false
```

which will disable the annotation based scraping configurations for prometheus and other scrapers.

I'm doing that with Node Exporter:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/charts/feature-cluster-metrics/values.yaml#L549-L551

Which is setting the value to false:
https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/docs/examples/features/cluster-metrics/default/output.yaml#L2556-L2557

When Helm makes it so we can remove values in subcharts, then i'll go that route, but this is a workable solution for now.

jkroepke:
Please go ahead with `annotations: {}`. It change might be breaky, but it's a cleaner solution at the end which i favorite. Please increase the minor version in that case.

petewall:
> Please go ahead with annotations: {} (by default, no prometheus.io/scrape annotation is set
As in, the values file remains:
```yaml
service:
    annotations: {}
```

And the service definition just applies those annotations? Would you keep the `unset` in the case of service monitors or pod monitors?

jkroepke:
Good catch! Yeah, the unset can be cleaned up as well. If the end user explicit set `prometheus.io/scrape` and enable the monitor, it might be expected.

jkroepke (approved):
LGTM

Issue #4770, opened 2024-08-02
[kube-state-metrics] DaemonSet option missing


wustus:
### Is your feature request related to a problem ?

The kube-state-metrics Repository [describes a Deployment strategy](https://github.com/kubernetes/kube-state-metrics?tab=readme-ov-file#daemonset-sharding-for-pod-metrics) where the application is rolled out as a DaemonSet and each instance is only responsible for a single node of the Kubernetes cluster.

### Describe the solution you'd like.

A `controller` key under `.Values.global` might be nice. The accepted values could be `deployment | daemonset | statefulset`.
An extra template: `daemonset.yaml` could be created which is applied if `.Values.global.controller` is `daemonset`.
`templates/deployment.yaml` would stay mostly the same, but instead of checking `{{- if .Values.autosharding.enabled }}` we could check if the controller is `statefulset`.

### Describe alternatives you've considered.

NONE

### Additional context.

We're running the kube-state-metrics application as a Deployment in our Kubernetes cluster. The metrics are scraped by some Grafana Agent. The agent exports an `x_host` label based on the `pod.nodeName` value. This can be misleading because the kube-state-metrics are collected cluster-wide. Meaning that metrics created on node `n3.cluster.com`, are still labeled as `x_host: n1.cluster.com` if the instance is running on node `n1.cluster.com`.
I'd still like to be able to filter metrics by node.
[The kube-state-metrics repository describes the solution](https://github.com/kubernetes/kube-state-metrics?tab=readme-ov-file#daemonset-sharding-for-pod-metrics) as running the application as a DaemonSet and limiting the responsibility of each instance to a node.
This would work well in our case, since the created Pods are distributed across all our nodes and will then produce the needed labels.

wustus:
This PR addresses a missing feature explicitly described in the kube-state-metrics README. Unless an alternative implementation exists, this should not be considered stale.

PR #5142, opened 2025-01-14
[kube-state-metrics] add Controller Selection

wustus:

This PR aims to implement the possibility of deploying the image as a `DaemonSet`. Currently only `Deployment` and `StatefulSet` are supported. 
When scraping metrics exported by `kube-state-metrics`, they will be labeled to originate from the node the pod is running on. In our case some `grafana-agent` collects metrics exposed by pods and adds an `x_host` label according to the node this pod is running on. By adding the possibility of running `kube-state-metrics` as a `DaemonSet`, metrics are correctly labeled with their node-origin.

Implements #4770

 - I tried to stay backward compatible by adding `{{- if eq (.Values.controller | default "deployment") "deployment" }}` in a few places.
 - `daemonset.yaml` is a copy of `deployment.yaml` with removal of `.Values.autosharding.enabled` and additional setting of the node name env variable according to [this](https://github.com/kubernetes/kube-state-metrics?tab=readme-ov-file#daemonset-sharding-for-pod-metrics).

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 5.28.0
+version: 5.28.1
 appVersion: 2.14.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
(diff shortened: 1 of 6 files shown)


wustus:
This PR addresses a missing feature explicitly described in the kube-state-metrics README. Unless an alternative implementation exists, this should not be considered stale.

PR #5165, opened 2025-01-14
[prometheus-yet-another-cloudwatch-exporter] fix yace chart name in readme

TanZng:
#### Which issue this PR fixes

- fixes error
```
Error: chart "prometheus-yet-another-cloudwatch-exporter" matching 0.39.0 not found in prometheus-community index. (try 'helm repo update'): no chart name found
```

read me was incorrect

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #5167, opened 2025-01-14
[prometheus-yet-another-cloudwatch-exporter] consistent naming of chart (make it match directory)

asherf:
see: https://github.com/prometheus-community/helm-charts/pull/5165#issuecomment-2590325789

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-yet-another-cloudwatch-exporter/Chart.yaml
+++ b/charts/prometheus-yet-another-cloudwatch-exporter/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-name: yet-another-cloudwatch-exporter
+name: prometheus-yet-another-cloudwatch-exporter
 description: Yace - Yet Another CloudWatch Exporter
 type: application
-version: 0.39.1
+version: 0.39.2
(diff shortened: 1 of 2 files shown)


cristiangreco (approved):

jkroepke:
It's a very strong breaking change, I hope you are aware that!

Because the full name template based on the chart name.

https://github.com/prometheus-community/helm-charts/blob/4adffa127408c0b9f3cf861d7a85c3778b6b707c/charts/prometheus-yet-another-cloudwatch-exporter/templates/_helpers.tpl#L5

cristiangreco:
@jkroepke the chart had been added just a day before, is that a big problem in your opinion?

We should be all settled now.

jkroepke:
What about the upgrade path from https://github.com/nerdswords/helm-charts/tree/main/charts/yet-another-cloudwatch-exporter

cristiangreco:
> in my case, i have to set nameOverride and fullnameOverride to avoid name changing on all resources.

We could add these info to the README? Could you please help with a PR? 🙏

Issue #4507, opened 2024-05-02
[prometheus-kube-stack] As per chart version 48.0.0 release notes we are no more required to upgrade CRD'S maually before upgrading stack to newer major version but in release notes of 49.0.0 again it is mentioned to do those manul steps to update CRD's using kubectl apply


kunalmehta-eve:
### Describe the bug a clear and concise description of what the bug is.

As per chart version 48.0.0 release notes we are no more required to upgrade CRD'S maually before upgrading stack to newer major version but in release notes of 49.0.0 again it is mentioned to do those manul steps to update CRD's using kubectl apply

### What's your helm version?

3.14.4

### What's your kubectl version?

1.28.5

### Which chart?

prometheus-kube-stack

### What's the chart version?

49.0.0

### Enter the command that you execute and failing/misfunctioning.

xx

GMartinez-Sisti:
This is not required if the CRDs are being managed by the helm chart, but since we don't know what users are doing they need to be included.

I would like to take this a step further and create a `pre-install` job hook that will install the CRDs for the correct version:

- this is now becoming a common pattern since the application and CRDs should match, so the application installs the correct CRDs on start, or in case of helm on deploy
- CRDs are better not being managed by GitOps tools to ensure they are not removed by mistake, we usually never want to remove CRDs, and if so should be a manual operation since it's dangerous
- scales with environments and GitOps practices by preventing manual steps on updates

This would be of course an optional configuration, disabled by default but I think it would benefit most users. I'm actually doing this right now on my job, so the overall work is already done.

@jkroepke since you were involved with the latest CRD changes I'd like to hear your input.

jkroepke:
> As per chart version 48.0.0 release notes we are no more required to upgrade CRD'S maually before upgrading stack to newer major

Could you please explain, where you got the information?

> This is not required if the CRDs are being managed by the helm chart

By design, Helm will not managed CRDs, because the helm maintainers has the opinion that it isn't possible to just run kubectl apply to managed them. Long story here: https://github.com/helm/community/blob/main/hips/hip-0011.md

> since you were involved with the latest CRD changes I'd like to hear your input.

I know that pattern from [velero](https://github.com/vmware-tanzu/helm-charts/tree/main/charts/velero/templates/upgrade-crds). Personally, I'm would never use such functionality, but a lot of users can benefit from it. Go for it. Luckly, hooks are not part of release, I did not except release size issues here.

PR #5175, opened 2025-01-15
[kube-prometheus-stack] upgrade CRDs using helm

jkroepke:

Since helm does not support upgrading CRDs, it's an manual task for our end-users. However the velero chart https://github.com/vmware-tanzu/helm-charts/tree/main/charts/velero/templates/upgrade-crds implement a upgrade job using helm hooks that works for at least 1 year.

- fixes #

I would leave it disabled for now in order to have the ability to test the job on real upgrades.

#### Checklist

- [x] Chart Version bumped


jkroepke:
@QuentinBisson @GMartinez-Sisti this is super WIP (but working) what did you think? It based on https://github.com/vmware-tanzu/helm-charts/tree/main/charts/velero/templates/upgrade-crds?rgh-link-date=2025-01-15T09%3A52%3A20Z

GMartinez-Sisti:
> @QuentinBisson @GMartinez-Sisti this is super WIP (but working) what did you think? It based on https://github.com/vmware-tanzu/helm-charts/tree/main/charts/velero/templates/upgrade-crds?rgh-link-date=2025-01-15T09%3A52%3A20Z

I'm all up for it! 🙌  -> https://github.com/prometheus-community/helm-charts/issues/4507

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - alertmanagerconfigs.monitoring.coreos.com
+      - alertmanagers.monitoring.coreos.com
+      - podmonitors.monitoring.coreos.com
+      - probes.monitoring.coreos.com
+      - prometheusagents.monitoring.coreos.com
+      - prometheuses.monitoring.coreos.com
+      - prometheusrules.monitoring.coreos.com
+      - scrapeconfigs.monitoring.coreos.com
+      - servicemonitors.monitoring.coreos.com
+      - thanosrulers.monitoring.coreos.com
```
GMartinez-Sisti, line 34: If possible we should make this dynamic so we don't miss new CRDs.

jkroepke: Done

jkroepke:
> I'm all up for it! 🙌 -> #4507

Please take note that the CRDs needs to be pre-compressed, All CRDs in gzip format are approx. 300KB, while xz just using 80KB. 

The creation of the tar.xz is part of hack scripts, but how we protect us against suspicious tar files?

GMartinez-Sisti:
> > I'm all up for it! 🙌 -> #4507
> 
> Please take note that the CRDs needs to be pre-compressed, All CRDs in gzip format are approx. 300KB, while xz just using 80KB.
> 
> The creation of the tar.xz is part of hack scripts, but how we protect us against suspicious tar files?

Checksum? We can add a checksum file next to it, then run the same on ci and check the match.

jkroepke:
I give-up after an hour to setup an reproducible tar archive which can be generated on any os (e.g. mac and linux)

GMartinez-Sisti:
> I give-up after an hour to setup an reproducible tar archive which can be generated on any os (e.g. mac and linux)
> 
> For example, I can't find a tar option on mac where I can drop the mtime of the files.
> 
> GNU tar adds some pax header while bsd tar does not do it.

Yeah, it’s hard to find a way with native commands, I usually fix these issues by running the commands in a docker container.

jkroepke:
What about check, 

> > I give-up after an hour to setup an reproducible tar archive which can be generated on any os (e.g. mac and linux)
> > For example, I can't find a tar option on mac where I can drop the mtime of the files.
> > GNU tar adds some pax header while bsd tar does not do it.
> 
> Yeah, it’s hard to find a way with native commands, I usually fix these issues by running the commands in a docker container.

please offer a solution. 

While tar isn't the big issue, xz might be. Seems not possible to get the same archive from macos/linux using xz compression. Even with gnu tar on mac, the archive is different. gz is not an option, because the compression rate is really, really bad.

Tests with zstd works fine, however busybox does not support zstd.

Then, I tried docker. The ubuntu docker image does not have xz pre installed. starting with apt-get first feels terrible slow. busybox have xz embedded, but does not offer capabilities to strip the mtime.

GMartinez-Sisti:
> please offer a solution.
> 
> While tar isn't the big issue, xz might be. Seems not possible to get the same archive from macos/linux using xz compression. Even with gnu tar on mac, the archive is different. gz is not an option, because the compression rate is really, really bad.
> 
> Tests with zstd works fine, however busybox does not support zstd.
> 
> Then, I tried docker. The ubuntu docker image does not have xz pre installed. starting with apt-get first feels terrible slow. busybox have xz embedded, but does not offer capabilities to strip the mtime.

Sure, I can test this during the weekend.

GMartinez-Sisti:
> Sure, I can test this during the weekend.

So, I've been trying multiple ways to achieve this, and I'm not able to achieve a similar compressed archive from both the systems. So maybe we can do a different approach and instead of doing a checksum on the archive file we can do a checksum on the contents of the archive?

I tested this new approach with the current tar on the repo, then tried the script with an updated compressed file created by both linux and macos.

<details>
  <summary>Click to expand diff.</summary>
  
  ```diff
diff --git a/charts/kube-prometheus-stack/hack/update_crds.sh b/charts/kube-prometheus-stack/hack/update_crds.sh
index 082e4cb6..b64dfca8 100755
--- a/charts/kube-prometheus-stack/hack/update_crds.sh
+++ b/charts/kube-prometheus-stack/hack/update_crds.sh
@@ -1,15 +1,18 @@
 #!/bin/bash
 
-set -e
+set -euo pipefail
 
-SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+TMP_DIR=$(mktemp -d)
 
+# Get current chart version
 if [[ $(uname -s) = "Darwin" ]]; then
-    VERSION="$(grep ^appVersion "${SCRIPT_DIR}/../Chart.yaml" | sed 's/appVersion: //g')"
+  VERSION="$(grep ^appVersion "${SCRIPT_DIR}/../Chart.yaml" | sed 's/appVersion: //g')"
 else
-    VERSION="$(grep ^appVersion "${SCRIPT_DIR}/../Chart.yaml" | sed 's/appVersion:\s//g')"
+  VERSION="$(grep ^appVersion "${SCRIPT_DIR}/../Chart.yaml" | sed 's/appVersion:\s//g')"
 fi
 
+# Set file map
 FILES=(
   "crd-alertmanagerconfigs.yaml :  monitoring.coreos.com_alertmanagerconfigs.yaml"
   "crd-alertmanagers.yaml       :  monitoring.coreos.com_alertmanagers.yaml"
@@ -23,42 +26,60 @@ FILES=(
   "crd-thanosrulers.yaml        :  monitoring.coreos.com_thanosrulers.yaml"
 )
 
+# Download latest files
 for line in "${FILES[@]}"; do
-    DESTINATION=$(echo "${line%%:*}" | xargs)
-    SOURCE=$(echo "${line##*:}" | xargs)
+  DESTINATION=$(echo "${line%%:*}" | xargs)
+  SOURCE=$(echo "${line##*:}" | xargs)
 
-    URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/$VERSION/example/prometheus-operator-crd/$SOURCE"
+  URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/${VERSION}/example/prometheus-operator-crd/${SOURCE}"
 
-    echo -e "Downloading Prometheus Operator CRD with Version ${VERSION}:\n${URL}\n"
+  echo -e "Downloading Prometheus Operator CRD with Version ${VERSION}:\n${URL}\n"
 
-    echo "# ${URL}" > "${SCRIPT_DIR}/../charts/crds/crds/${DESTINATION}"
+  echo "# ${URL}" >"${SCRIPT_DIR}/../charts/crds/crds/${DESTINATION}"
 
-    if ! curl --silent --retry-all-errors --fail --location "${URL}" >> "${SCRIPT_DIR}/../charts/crds/crds/${DESTINATION}"; then
-      echo -e "Failed to download ${URL}!"
-      exit 1
-    fi
+  if ! curl --silent --retry-all-errors --fail --location "${URL}" >>"${SCRIPT_DIR}/../charts/crds/crds/${DESTINATION}"; then
+    echo -e "Failed to download ${URL}!"
+    exit 1
+  fi
 done
 
-_TAR=$(which tar 2>/dev/null)
+# Calculate checksum for existing tar
+SHA256_CURL_FILES="$(find "${SCRIPT_DIR}/../charts/crds/crds/"* -print0 | sort -z | xargs -0 cat | sha256sum)"
+echo -e "SHA256_CURL_FILES:\t${SHA256_CURL_FILES}"
 
 cd "${SCRIPT_DIR}/../charts/crds/crds/"
 
-case $($_TAR --help) in
-  *GNU*)
-    find crd-*.yaml -print0 | sort -z | env XZ_OPT=-9 $_TAR --sort=name --format=ustar \
-      --mtime="@0" \
-      --numeric-owner --owner=0 --group=0 \
-      --mode='go+u,go-w' \
-      --no-xattrs --no-acls --no-selinux \
-      --no-recursion --null --files-from - \
-      -cJpf ../files/crds.tar.xz
-    ;;
-  *)
-    find crd-*.yaml -exec touch -d 1970-01-01T00:00:00Z {} + -print0 | sort -z | env COPYFILE_DISABLE=1 $_TAR --format=ustar \
-      --numeric-owner --uid=0 --gid=0 \
-      --no-xattrs \
-      --no-recursion --null --files-from - \
-      --options xz:compression-level=9 \
-      -cJf ../files/crds.tar.xz
-    ;;
+# Calculate checksum for existing tar
+mkdir "${TMP_DIR}\extracted"
+tar -xf ../files/crds.tar.xz -C "${TMP_DIR}\extracted"
+SHA256_EXTRACTED_FILES="$(find crd-*.yaml -print0 | sort -z | xargs -0 cat | sha256sum)"
+echo -e "SHA256_EXTRACTED_FILES:\t${SHA256_EXTRACTED_FILES}"
+rm -rf "${TMP_DIR}"
+
+# Check content match between downloaded and extracted files
+if [[ "${SHA256_CURL_FILES}" == "${SHA256_EXTRACTED_FILES}" ]]; then
+  echo -e "\nChecksums matched, skipping compression."
+  exit 0
+fi
+
+# Create new tar
+TAR_BIN=$(which tar 2>/dev/null)
+case $("${TAR_BIN}" --help) in
+*GNU*)
+  find crd-*.yaml -print0 | sort -z | env XZ_OPT=-9 "${TAR_BIN}" --sort=name --format=ustar \
+    --mtime="@0" \
+    --numeric-owner --owner=0 --group=0 \
+    --mode='go+u,go-w' \
+    --no-xattrs --no-acls --no-selinux \
+    --no-recursion --null --files-from - \
+    -cJpf ../files/crds.tar.xz
+  ;;
+*)
+  find crd-*.yaml -exec touch -d 1970-01-01T00:00:00Z {} + -print0 | sort -z | env COPYFILE_DISABLE=1 "${TAR_BIN}" --format=ustar \
+    --numeric-owner --uid=0 --gid=0 \
+    --no-xattrs \
+    --no-recursion --null --files-from - \
+    --options xz:compression-level=9 \
+    -cJf ../files/crds.tar.xz
+  ;;
 esac

  ```

</details>

Please let me know what you think.

jkroepke:
@GMartinez-Sisti After reading your message, I though do we really tar? We do not need file metadata and file metadata are the issue for the checksums. 

Instead, I'm using cat to concat all CRDs files and pipe the data directly through a compress program. 

I also switched to bzip2. it seems that depends on the xz version, the output is different. After tryout gzip (320KB), I looked into bz2 (120KB) which results into stable checksum and should solve the issue for now. 

If users report issue in that context, I will look into this again. Maybe zstd would be the better that to go, but its more complicated to implement, since busybox does not ship zstd.

jkroepke:
Ready to review!

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.upgradeJob.enabled }}
```
QuentinBisson, line 1: I think I would move upgradeJob under crds.upgradeJob to avoid potential clashes in the future. WDYT?

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "kube-prometheus-stack.upgradeCRD.name" . }}
+  namespace: {{ template "kube-prometheus-stack.namespace" . }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade,pre-rollback
```
QuentinBisson, line 8: Does that mean that this will rollback the crds successfully? That would be pretty neat but if not, it probably needs to be documented in the upgrade process

charts/kube-prometheus-stack/charts/crds/templates/upgrade/job.yaml
```diff
@@ -0,0 +1,142 @@
...
+      {{- with .Values.upgradeJob.podLabels }}
+      labels:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.upgradeJob.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+    {{- if .Values.global.imagePullSecrets }}
```
QuentinBisson, line 32: Nit but can you indent this with the imagePullSecrets and not the spec?

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
QuentinBisson: Oh it is in the values but it's not used correctly here

jkroepke: crd is the name of the sub chart. the values are define on top level pass to the sub chart by prefix all values with the chart name.

It works similar with grafana and node-exporter values where all grafana values prefixes with `grafana`

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
jkroepke: theoretically yes.

charts/kube-prometheus-stack/charts/crds/templates/upgrade/clusterrole.yaml
QuentinBisson: Damn I'm tired, sorry about this

jkroepke:
> Overall I quite like the job approach but i'm not really fond of the archive though. Is there a specific reason why we cannot use curl to fetch the crds from this chart app version instead?

Yeah, sideloading is not good in enterprise context where internet is restricted. That lead to unexpected situations.

aidanleuck:
This is awesome! I started looking at ways I could implement something similar in our wrapper chart. We use Terraform to deploy this chart and manage 20-30 Kubernetes clusters. Performing the CRD upgrades manually when we needed to take a major version upgrade had become tedious. This will help out our org a ton and resolve our biggest pain point with the deployment.

jkroepke:
@aidanleuck great to know. Once implemented, we are happy about feedback.

muffl0n:
Really looking forward to this, too! If you need any early feedback, let me know!

aidanleuck:
> > Overall I quite like the job approach but i'm not really fond of the archive though. Is there a specific reason why we cannot use curl to fetch the crds from this chart app version instead?
> 
> Yeah, sideloading is not good in enterprise context where internet is restricted. That lead to unexpected situations.
> 
> One possible solution would be that we provide the CRDs as OCI image. Then, we can mount the image as PVC. but It will take some time: https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/

I like mounting an OCI image, but as you mentioned this is a cutting-edge feature and it won't work with older clusters. If I understand this PR correctly, the busybox image is doing the decompression of the CRD archive, the results of which are placed in a PVC that the kubectl pod uses to apply. Woud it be simpler if you created a Dockerfile and extend the kubectl image? That way you don't need a shared PVC at all and you just use a single container image to run the job.

CI/Deployment you can build the custom kubectl image and the CRDS bundled together. You can tag the image with the same version as the release. This will get rid of the tar archive issue as well.

Here's the pseudocode
Dockerfile
```
FROM registry.k8s.io/kubectl
curl crd1.. -o /var/crds
curl crd2... -o /var/crds
```

job.yaml
```
workingDir: /var/
command:    
   - kubectl
args:
  - apply
  - --server-side
  - --filename
  - /var/crds/*.yaml
```
Let me know if I am off base here and am missing something, but this seems like a simpler solution to the problem.

jkroepke:
> Would it be simpler if you created a Dockerfile and extend the kubectl image?

This repo doesn't have a process for build images. It's unknown to me, but I would like to hear opionon form @GMartinez-Sisti @QuentinBisson  first.

There is not PVC between both containers, it's just an empty dir. I'm more looking forward to an upstream solution: https://github.com/prometheus-operator/prometheus-operator/issues/7270 - the bz2 archive is more or less an workaround unless the promtheus-operator provide a solution here.

aidanleuck:
> > Would it be simpler if you created a Dockerfile and extend the kubectl image?
> 
> 
> 
> This repo doesn't have a process for build images. It's unknown to me, but I would like to hear opionon form @GMartinez-Sisti @QuentinBisson  first.
> 
> 
> 
> There is not PVC between both containers, it's just an empty dir. I'm more looking forward to an upstream solution: https://github.com/prometheus-operator/prometheus-operator/issues/7270 - the bz2 archive is more or less an workaround unless the promtheus-operator provide a solution here.

Thanks for the context on why this direction was chosen. You are totally right I saw a volume mount and just assumed it was a PVC. I agree the upstream solution looks like a better solution long term than building and maintaining a docker image or bz2 archive. I have some experience with building docker images via GitHub actions so if that is the solution you guys decide on I will be happy to help out.

QuentinBisson:
@jkroepke I would be fine with merging this as is if you resolve the conflict because this is a huge improvement :)

QuentinBisson (approved):

jkroepke:
@QuentinBisson I would not bump a major now, but maybe keep this in mind, if there are new CRDs.

QuentinBisson:
I agree, especially since this is disabled by default :)

anthonyalayo:
I attempted to use this feature, but I was hit with conflicts when the upgrade job ran. Conflicts with keys such as `.metadata.annotations.controller-gen.kubebuilder.io/version` or `.metadata.annotations.operator.prometheus.io/version` blocked the merge. I had to do the manual kubectl apply via CLI with `--force-conflicts`. 

Has anyone run into this issue? if so, how did you solve it? For now, I had to disable the upgradeJob.

aidanleuck:
@anthonyalayo @peterbaumert A subsequent PR adds the ability to run --force-conflicts when the upgrade job runs. Setting upgradeJob.forceConflicts to true should fix this issue

PR #5177, opened 2025-01-15
[kube-prometheus-stack] Add Grafana Route default values

nilsjaekel:
Updated the chart's values.yaml with the default values ​​for the Grafana route configuration.
Bumped the version of the chart from 68.1.0 -> 68.2.0

- fixes no issues

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -23,7 +23,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 68.1.0
+version: 68.2.0
 appVersion: v0.79.2
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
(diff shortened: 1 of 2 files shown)


jkroepke:
is there a reason for importing default values?

nilsjaekel:
> is there a reason for importing default values?

This would make it consistent with the other tools like alertmanager, prometheus and thanosRuler where the values ​​for the routes are already included. Other than that there is no real reason :)

jkroepke:
The difference is that alertmanager, prometheus and thanosRuler are part of this chart while grafana not. I guess nobody will every maintain the upstream default value and take not, if changes on grafana side.

nilsjaekel:
Oh, I have overseen this point. I can understand your concern. 

While I doubt route values ​​will change anytime soon, wouldn't it be better to remove more Grafana values, as most of them seem to have been copied from upstream defaults at some point?
I've just checked some values i.e. grafana.serviceAccount.autoMount in this chart that got deprecated in the grafana upstream chart. 

But back to main topic:
Even though I would be happy to contribute to the project with this PR, I can understand if there is no point in merging this PR.

jkroepke:
> While I doubt route values ​​will change anytime soon, wouldn't it be better to remove more Grafana values, as most of them seem to have been copied from upstream defaults at some point?

From my point of view, all default grafana values might be removed and non default values should be applied only with few exception, for example grafana.sidecar.

Thanks for you PR. but it seems more worth to clean it up.

PR #5187, opened 2025-01-16
[kube-prometheus-stack] refine filters on kubelet alerts

TheRealNoob:
What this PR does / why we need it

Filter these alerts to nodes which are schedulable. We have received this alert many times when a node is under maintenance. I'm sure the same is true for others and I can't think of a case when someone would want to be alerted on a Cordoned node, I think it's safe to assume it's under maintenance and doesn't require an alert.

Didn't open one

#### Checklist

- [x] Chart Version bumped


TheRealNoob:
hmmmm I tried running the hack script but it reverts my change.....

TheRealNoob:
I see, I have to commit to another repo and then import here.  Okay I will do that and come back.  I will close this MR now because I want to make a second change, so I will re-open this merging from a new branch and not from master.

Issue #4292, opened 2024-02-26
Proposal: Use Renovate instead Dependabot as dependency manager


jkroepke:
Currently, https://github.com/prometheus-community/helm-charts is using dependabot for tracking dependency.

However, dependabot is unable to track to bump dependencies inside helm values.

While there is no native integration in Renovate, renovate can instrument with regex to take care of dependencies.

Example renovate config:
https://github.com/jkroepke/helm-charts/blob/b513065a173454484eecf412aa85032f905baf8d/renovate.json#L11-L19

Example: Dependency

https://github.com/jkroepke/helm-charts/blob/b513065a173454484eecf412aa85032f905baf8d/charts/amazon-eks-pod-identity-webhook/Chart.yaml#L6-L7

Example PR:

https://github.com/jkroepke/helm-charts/pull/51

To avoid duplicate config, I would recommend to disable dependabot here.

Pulling @prometheus-community/helm-charts-maintainers @prometheus-community/helm-charts-admins for additional opinions here.

I guess we also need an org admin for setup the renovate integration.

monotek:
Does renovate also checks github action versions?
If not i would also be ok to use both and use renovate only for docker image updates.
I guess the hard part will be the regex as not all charts use "image": Imhos some are using "tag" too....

jkroepke:
> Does renovate also checks github action versions?

Yes. 

Example: https://github.com/jkroepke/openvpn-auth-oauth2/pull/135

jkroepke:
There only downside here is that renovate is unable to bump the chart version, expect for bumping chart dependencies. In this case, I update the version via suggestions an apply them. 

The next benefit, as chart maintainer, I can update versions on my own chart, because the contribution is external and I'm able to merge it.

jkroepke:
@monotek Maybe it just works now. 

https://github.com/renovatebot/renovate/issues/8231 - the issue was resolved 5 weeks ago.

monotek:
How do we want to proceed?
Will you create a pr and add the needed renvovate comments to all charts?

jkroepke:
> How do we want to proceed?

We have to ask an org owner to configure the Renovate App first (https://github.com/apps/renovate)

monotek:
@SuperQ @brancz 

Could you enable renovcate app for us please?

jkroepke:
@monotek Secondly, I would recommend to add the renovate.json to the code owners files, like we do the .github/workflows/.

[...]

PR #5196, opened 2025-01-17
[ci] add github workflow to the bump version of an Chart version

jkroepke:

xref: #4292 

This PR continues the integration of Renovate into this repository. Unfortunately, Renovate cannot automatically bump the Helm chart version for changes made in the values files.

To address this, a support workflow has been added, which uses GitHub Actions to bump the version.

Additionally, the renovate.json file is configured to group version bumps by chart directory. Custom managers have been added from our internal company repository. These custom managers allow developers to enable automated dependency updates by adding Renovate annotations, using code comments. The regex patterns include common use cases, such as the following example:
https://github.com/prometheus-community/helm-charts/blob/7c7bc2e63639ff9cfb8ebae1903efeb4037b3923/charts/kube-prometheus-stack/values.yaml#L708-L711

Once worked, the next iteration is extend the renovate-bump-chart-version workflow to run the hack/update*sh scripts.

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/.github/workflows/renovate-bump-chart-version.yaml
+++ b/.github/workflows/renovate-bump-chart-version.yaml
@@ -1,0 +1,92 @@
+name: Renovate Bump Chart Version
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'charts/**/*'
+
+jobs:
+  renovate-bump-chart-version:
+    name: Renovate Bump Chart Version
+    runs-on: ubuntu-latest
+    if: github.actor == 'renovate[bot]'
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.6.1
+
+      - name: Detect changed charts
+        id: list-changed
+        run: |
+          changed="$(ct list-changed --config .github/linters/ct.yaml)"
+          if [[ -n "$changed" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/renovate-bump-chart-version.yaml
```diff
@@ -0,0 +1,92 @@
...
+    paths:
+      - 'charts/**/*'
+
+jobs:
+  renovate-bump-chart-version:
+    name: Renovate Bump Chart Version
+    runs-on: ubuntu-latest
+    if: github.actor == 'renovate[bot]'
+    permissions:
+      contents: write
```
GMartinez-Sisti, line 16: The logic looks good, I'm not sure if renovate will actually have permissions to write. We'll have to see 🤞

GMartinez-Sisti (approved):

.github/workflows/renovate-bump-chart-version.yaml
jkroepke: thats not how it works.

The if condition means, run the job only, if the PR comes from renovate. Inside the github action, the action user will still do things, not renovate.

GMartinez-Sisti: You are right, I was mistaking it with dependabot permissions that are special ([docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)):
> Workflow runs triggered by Dependabot pull requests run as if they are from a forked repository, and therefore use a read-only GITHUB_TOKEN. These workflow runs cannot access any secrets. For information about strategies to keep these workflows secure, see [Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions).

Issue #4992, opened 2024-11-11
[kube-state-metrics] Readiness probe not working with rbac proxy


plnordquist:
### Describe the bug a clear and concise description of what the bug is.

After upgrading to chart version v5.27.0, the kube-state-metrics pod is never marked as ready if the kube-rbac-proxy feature is enabled.  The deployment configures the telemetry host to 127.0.0.1 when `kubeRBACProxy.enabled` is true and the readiness probe uses the pod IP to attempt to reach the telemetry host so it is not reachable.

### What's your helm version?

version.BuildInfo{Version:"v3.16.2", GitCommit:"13654a52f7c70a143b1dd51416d633e1071faffb", GitTreeState:"clean", GoVersion:"go1.22.7"}

### What's your kubectl version?

Client Version: v1.31.2 Kustomize Version: v5.4.2 Server Version: v1.31.2

### Which chart?

kube-state-metrics

### What's the chart version?

5.27.0

### Enter the changed values of values.yaml?

```
kubeRBACProxy:
  enabled: true
```

### Enter the command that you execute and failing/misfunctioning.

helm install -n monitoring ksm prometheus-community/kube-state-metrics --set "kubeRBACProxy.enabled=true"

HaveFun83:
same here

fahedouch:
same here. The issue is the following: when `rbacproxy`enabled, the `telemetry` server port is binded comes to the loopback IP adresse (`127.0.0.1`) which prevents the kubelet from checking the readiness of the pod, the solution is to bind the socket to 0.0.0.0 or to vary the host via the values

r0bj:
The purpose of using kube-rbac-proxy is to prevent unauthorized access to the kube-state-metrics endpoints. However, binding it to 0.0.0.0 so that the kubelet can use it for probes defeats this purpose because it allows direct access to the kube-state-metrics port, bypassing kube-rbac-proxy entirely. So, we should bind both ports of kube-state-metrics to 127.0.0.1, which unfortunately would cause the kubelet probes to stop working. One solution would be to disable probes when using kube-rbac-proxy, similar to how it is done in kube-prometheus: https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/kubeStateMetrics-deployment.yaml

PR #4982, opened 2024-11-09
[kube-state-metrics] bump to 2.14.0

DrFaust92:
#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #5215, opened 2025-01-19
[kube-state-metrics] make livenessProbe and readinessProbe optional

r0bj:
This attempt addresses the issue of probes failing, which occurred after the changes in PR https://github.com/prometheus-community/helm-charts/pull/4982. By making the livenessProbe and readinessProbe optional, users can disable them when using kube-rbac-proxy.

- fixes # https://github.com/prometheus-community/helm-charts/issues/4992

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 5.28.0
+version: 5.29.0
 appVersion: 2.14.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -178,6 +178,7 @@ spec:
           successThreshold: {{ .Values.startupProbe.successThreshold }}
           timeoutSeconds: {{ .Values.startupProbe.timeoutSeconds }}
         {{- end }}
+        {{- if .Values.livenessProbe.enabled }}
         livenessProbe:
           failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
           httpGet:
@@ -196,6 +197,8 @@ spec:
           periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
           successThreshold: {{ .Values.livenessProbe.successThreshold }}
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+        {{- end }}
+        {{- if .Values.readinessProbe.enabled }}
         readinessProbe:
           failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
           httpGet:
@@ -214,6 +217,7 @@ spec:
           periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
           successThreshold: {{ .Values.readinessProbe.successThreshold }}
           timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+        {{- end }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
 {{- if .Values.containerSecurityContext }}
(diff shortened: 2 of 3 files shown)


r0bj:
@SuperQ Probes are failing because the port is bound to 127.0.0.1, so the kubelet cannot reach it. There are three potential solutions, each with its trade-offs:
1. Bind both main and telemetry ports to 0.0.0.0.
In this case, the kubelet can access the ports via probes, but doing so defeats the purpose of using kube-rbac-proxy, which is intended to prevent unauthorized access to the kube-state-metrics endpoints.

2. Bind both main and telemetry ports to 127.0.0.1.
This configuration hides the kube-state-metrics endpoints, allowing access only through the kube-rbac-proxy-protected ports. However, the kubelet will be unable to reach the kube-state-metrics ports, causing probes to fail.

3. Bind both main and telemetry ports to 127.0.0.1 and use the host network.
With this approach, the kubelet can reach the kube-state-metrics endpoints, so probes work correctly, and the endpoints remain protected. The downside is that running pods with the host network may not be allowed due to namespace security restrictions (for example, the baseline and restricted pod security standards as documented in the Pod Security Standards: https://kubernetes.io/docs/concepts/security/pod-security-standards/).

Given these options, perhaps we should provide a configuration option that allows users to choose their preferred trade-off based on their environment and security requirements.

fahedouch:
@r0bj
Additional option: 
Expose an endpoint on the `kube-rbac-proxy` to check the readiness/liveness of `kube-state-metrics`. This makes a lot of sense to me as the proxy is the only way to health check the `kube-state-metrics` port.

r0bj:
> Additional option:
Expose an endpoint on the kube-rbac-proxy to check the readiness/liveness of kube-state-metrics. This makes a lot of sense to me as the proxy is the only way to health check the kube-state-metrics port.

Yes, this option seems reasonable. I’ve just checked, and kube-rbac-proxy has a parameter that may be helpful for this purpose:
```
--ignore-paths strings                        Comma-separated list of paths against which kube-rbac-proxy pattern-matches the incoming request. If the requst matches, it will proxy the request without performing an authentication or authorization check. Cannot be used with --allow-paths.
```

r0bj:
This is my proposal for implementing the option mentioned by @fahedouch — to use the ports exposed by kube-rbac-proxy for probes:: https://github.com/prometheus-community/helm-charts/pull/5234

dotdc:
Closing in favor of https://github.com/prometheus-community/helm-charts/pull/5234

refucktor:
@r0bj do you mind to share what is the final combination that works to have the readiness/liveness probes enabled? Thanks in advance 👍🏻

r0bj:
@refucktor I have successfully configured readiness and liveness probes with kube-rbac-proxy enabled after upgrading the kube-state-metrics Helm chart to version 5.29.0. Here are the values I used:
```
  kubeRBACProxy:
    enabled: true
    extraArgs:
    - --ignore-paths=/livez,/readyz
```

refucktor:
@r0bj thanks for the reply, unfortunately still not working for me, specifically the readiness fails at `https://<pod_ip>:8081/readyz`, here is my current config


```hcl
resource "helm_release" "kube_state_metrics" {
  name             = "kube-state-metrics"
  namespace        = var.namespaces.monitoring
  create_namespace = false

  chart      = "kube-state-metrics"
  version    = "5.30.0"
  repository = "https://prometheus-community.github.io/helm-charts"

  max_history     = 3
  atomic          = true
  cleanup_on_fail = true
  recreate_pods   = true
  skip_crds       = true
  upgrade_install = true
  timeout         = 60 * 3

  values = [yamlencode({
    prometheusScrape = true

    revisionHistoryLimit = 3


    kubeRBACProxy = {
      enabled = true

      extraArgs = [
        "--ignore-paths=/livez,/readyz",
      ]

      resources = {
        limits   = { cpu = "100m", memory = "64Mi" }
        requests = { cpu = "10m", memory = "32Mi" }
      }
    }

    resources = {
      limits   = { cpu = "200m", memory = "300Mi" }
      requests = { cpu = "100m", memory = "250Mi" }
    }
  })]
}

```

r0bj:
@refucktor The /readyz endpoint is available when .Values.selfMonitor.enabled is set to true.

PR #5231, opened 2025-01-23
[kube-state-metrics] tpl affinity

TheRealNoob:

Add `tpl` templating to affinity.  Useful when deploying multiple instances of this chart, or when you don't want to hardcode selector labels.

Didn't open one.

#### Checklist

- [x] Chart Version bumped


charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -307,7 +307,11 @@ spec:
       {{- end }}
       {{- if .Values.affinity }}
       affinity:
-{{ toYaml .Values.affinity | indent 8 }}
+        {{- if kindIs "map" .Values.affinity }}
+        {{- toYaml .Values.affinity | nindent 8 }}
+        {{- else }}
+        {{- tpl .Values.affinity $ | nindent 8 }}
+        {{- end }}
```
dotdc, line 314: Could you try this instead?
```suggestion
        {{- tpl (toYaml .Values.affinity) $ | indent 8 }}
```

TheRealNoob: unfortunately that doesn't work.  the `toYaml` converts the whole thing into a single-line string.  it would work if you removed toYaml which I don't think is strictly necessary other than checking if the passed YAML is valid format.

values2.yaml
```
affinity: |
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            {{- include "kube-state-metrics.selectorLabels" . | nindent 12 }}
        topologyKey: kubernetes.io/hostname
```

```
s$ helm template state-metrics . -f values2.yaml
...
      affinity: "podAntiAffinity:\n  requiredDuringSchedulingIgnoredDuringExecution:\n    - labelSelector:\n        matchLabels:\n          \n          app.kubernetes.io/name: kube-state-metrics\n          app.kubernetes.io/instance: state-metrics\n      topologyKey: kubernetes.io/hostname\n"
```

dotdc: This is probably because you changed the `affinity` type to string (`""`) in the original `values.yaml`.
Revert to `{}` and it should work:
```
$ helm template state-metrics . -f values2.yaml | grep -C5 requiredDuringSchedulingIgnoredDuringExecution
            drop:
            - ALL
          readOnlyRootFilesystem: true
      affinity:        |
          podAntiAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              - labelSelector:
                  matchLabels:            
                    app.kubernetes.io/name: kube-state-metrics
                    app.kubernetes.io/instance: state-metrics
                topologyKey: kubernetes.io/hostname

```

Not a big deal, but you should also replace `nindent 12` to `indent 12` in your `values2.yaml` file.

TheRealNoob: if I'm understanding you right, this is what I'm getting when I implement it.  I did away with values2 for the purpose of this demonstration and just merged it into values.yaml for testing.

https://gist.github.com/TheRealNoob/1f77d8de4faaafd633d19fc7dd318a12

dotdc: You removed the pipe after `affinity:` while moving the block from one file to the other.
`affinity: |`

TheRealNoob: I did that to demonstrate that `tpl (toYaml .Values.affinity) $` does not work with a dictionary.  It does when `affinity` is a string.

TheRealNoob: `tpl` expects a string as input.  when affinity is a dictionary, `toYaml` keeps is as such, which causes the tpl to fail.

TheRealNoob: additionally, if affinity is a string, then toYaml casts it as a single-line string.  That's what I attempted to demonstrate [here](https://github.com/prometheus-community/helm-charts/pull/5231#discussion_r1930883438)

TheRealNoob: just to clarfiy I would like if your solution worked.  I prefer it, it's cleaned and easier to understand.  But I don't see how to make it work.

dotdc: Download a fresh copy of the `main` branch of this repository and `cd` into it.
Replace the [line 310](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-state-metrics/templates/deployment.yaml#L310) of charts/kube-state-metrics/templates/deployment.yaml with:
`{{- tpl (toYaml .Values.affinity) $ | indent 6 }}`

```
cd charts/kube-state-metrics
```

Create a file with your configuration:

values2.yaml

```
affinity: |
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            {{- include "kube-state-metrics.selectorLabels" . | indent 12 }}
        topologyKey: kubernetes.io/hostname
```

Then:
```
helm template state-metrics . -f values2.yaml
```

Doest it work?

TheRealNoob: hmm that does work.  let me look into it a little more and get back to you.

TheRealNoob: yeah, it's the same story.  it works when affinity is a string but not a dictionary, meaning it's not backwards compatible with the current values people may have defined.  Following your example, try changing values2 to this and run it again.

```
affinity:
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app.kubernetes.io/name: kube-state-metrics
            app.kubernetes.io/instance: state-metrics
        topologyKey: kubernetes.io/hostname
```

dotdc: My bad, didn't realized that this was now a multi-line strings...
Need time to reconsider, it feels a bit weird to support multiple types for affinity.

dotdc: We already have something similar [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-sql-exporter/templates/deployment.yaml#L55).
I guess it's fine.

charts/kube-state-metrics/values.yaml
```diff
@@ -304,8 +304,15 @@ containerSecurityContext:
...
+## Can be defined as either a dict or string.  String is useful for `tpl` templating.
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
-affinity: {}
+affinity: ""
+  # podAntiAffinity:
+  #   requiredDuringSchedulingIgnoredDuringExecution:
+  #     - labelSelector:
+  #         matchLabels:
+  #           {{- include "kube-state-metrics.selectorLabels" . | nindent 12 }}
+  #       topologyKey: kubernetes.io/hostname
```
dotdc, line 315: The shouldn't work, we either need the multi-line strings notation or remove the templating.

charts/kube-state-metrics/values.yaml
```diff
@@ -304,8 +304,15 @@ containerSecurityContext:
...
 ## Affinity settings for pod assignment
+## Can be defined as either a dict or string.  String is useful for `tpl` templating.
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
-affinity: {}
+affinity: ""
+  # podAntiAffinity:
+  #   requiredDuringSchedulingIgnoredDuringExecution:
+  #     - labelSelector:
+  #         matchLabels:
+  #           {{- include "kube-state-metrics.selectorLabels" . | nindent 12 }}
```
dotdc, line 314:
```suggestion
  #           {{- include "kube-state-metrics.selectorLabels" . | nindent 10 }}
```

TheRealNoob: implemented

charts/kube-state-metrics/values.yaml
TheRealNoob: updated defaults.  what do you think?

charts/kube-state-metrics/values.yaml
```diff
@@ -304,8 +304,17 @@ containerSecurityContext:
 nodeSelector: {}
 
 ## Affinity settings for pod assignment
+## Can be defined as either a dict or string.  String is useful for `tpl` templating.
```
dotdc, line 307:
```suggestion
## Can be defined as either a dict or string. String is useful for `tpl` templating.
```

charts/kube-state-metrics/values.yaml
```diff
@@ -304,8 +304,17 @@ containerSecurityContext:
...
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
 affinity: {}
+#
```
dotdc, line 310:
```suggestion
```

dotdc (approved):

dotdc:
Thank you for your contribution @TheRealNoob !

PR #5234, opened 2025-01-24
[kube-state-metrics] Use kube-rbac-proxy ports for probes when kube-rbac-proxy is enabled

r0bj:
When kubeRBACProxy is enabled, probes from the kubelet fail because the kube-state-metrics ports are bound to 127.0.0.1. This PR configures the probes to use the ports protected by kube-rbac-proxy, ensuring that health checks succeed when kube-rbac-proxy is enabled.

For more context, please refer to [this comment](https://github.com/prometheus-community/helm-charts/pull/5215#issuecomment-2602933330).

- fixes https://github.com/prometheus-community/helm-charts/issues/4992

#### Checklist

- [x] Chart Version bumped


fahedouch:
thanks you for implementing this @r0bj. LGTM

fahedouch (approved):

dotdc (commented):
Hi @r0bj,
Thank you for your pull requests on this topic.
For the first part of the pull request, I think that you could use `extraArgs` and `kubeRBACProxy.extraArgs` directly.
If you need specific extra args for the `http` and `telemetry` containers, I would rather add something like `kubeRBACProxy.extraHttpArgs` and `kubeRBACProxy.extraTelemetryArgs`.
What do you think?

r0bj:
@dotdc Yes, that also makes sense. We can remove the conditional logic for including --ignore-paths in the kube-rbac-proxy containers and instead rely on .Values.kubeRBACProxy.extraArgs to pass `--ignore-paths=/livez,/readyz`. Would that approach work better?

dotdc:
Yes I would prefer, as it gives more flexibility.
Can you update the pull request?

r0bj:
> Yes I would prefer, as it gives more flexibility.
Can you update the pull request?

@dotdc Yes, I updated the pull request.

r0bj:
> @r0bj You removed flags in https://github.com/prometheus-community/helm-charts/commit/4118f785807c7732f1777e94969d91c782cc91dd without implementing kubeRBACProxy.extraHttpArgs and kubeRBACProxy.extraTelemetryArgs, is it intended?
> Did you notice it wasn't needed in the end?

Instead of creating additional values for each kube-rbac-proxy instance (e.g.,  `.Values.kubeRBACProxy.extraHttpArgs`,  `.Values.kubeRBACProxy.extraTelemetryArgs`), we can use the existing value `.Values.kubeRBACProxy.extraArgs`, which simplifies the configuration. For example, setting `.Values.kubeRBACProxy.extraArgs` to `--ignore-paths=/livez,/readyz` should work fine. Let me know if that works for you.

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -79,6 +79,9 @@ spec:
         {{-  if .Values.extraArgs  }}
         {{- .Values.extraArgs | toYaml | nindent 8 }}
         {{-  end  }}
+        {{- if .Values.kubeRBACProxy.enabled }}
+        - --host=127.0.0.1
+        {{-  end  }}
```
dotdc, line 84:
```suggestion
        {{- end }}
```

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -171,8 +174,13 @@ spec:
               value: {{ $header.value }}
             {{- end }}
             path: /healthz
+            {{- if .Values.kubeRBACProxy.enabled }}
+            port: {{ .Values.service.port | default 8080 }}
+            scheme: HTTPS
+            {{- else }}
             port: {{ $servicePort }}
             scheme: {{ upper .Values.startupProbe.httpGet.scheme }}
+            {{-  end  }}
```
dotdc, line 183:
```suggestion
            {{- end }}
```

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -190,8 +198,13 @@ spec:
               value: {{ $header.value }}
             {{- end }}
             path: /livez
+            {{- if .Values.kubeRBACProxy.enabled }}
+            port: {{ .Values.service.port | default 8080 }}
+            scheme: HTTPS
+            {{- else }}
             port: {{ $servicePort }}
             scheme: {{ upper .Values.livenessProbe.httpGet.scheme }}
+            {{-  end  }}
```
dotdc, line 207:
```suggestion
            {{- end }}
```

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -208,8 +221,13 @@ spec:
               value: {{ $header.value }}
             {{- end }}
             path: /readyz
+            {{- if .Values.kubeRBACProxy.enabled }}
+            port: {{ .Values.selfMonitor.telemetryPort | default 8081 }}
+            scheme: HTTPS
+            {{- else }}
             port: {{ $telemetryPort }}
             scheme: {{ upper .Values.readinessProbe.httpGet.scheme }}
+            {{-  end  }}
```
dotdc, line 230:
```suggestion
            {{- end }}
```

dotdc (approved):

dotdc:
Thank you for your contribution @r0bj !

fahedouch:
thanks @r0bj.

Is there a planned release including this patch?

dotdc:
@fahedouch released in [kube-state-metrics-5.29.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-state-metrics-5.29.0)

HaveFun83:
> > @r0bj You removed flags in [4118f78](https://github.com/prometheus-community/helm-charts/commit/4118f785807c7732f1777e94969d91c782cc91dd) without implementing kubeRBACProxy.extraHttpArgs and kubeRBACProxy.extraTelemetryArgs, is it intended?
> > Did you notice it wasn't needed in the end?
> 
> Instead of creating additional values for each kube-rbac-proxy instance (e.g., `.Values.kubeRBACProxy.extraHttpArgs`, `.Values.kubeRBACProxy.extraTelemetryArgs`), we can use the existing value `.Values.kubeRBACProxy.extraArgs`, which simplifies the configuration. For example, setting `.Values.kubeRBACProxy.extraArgs` to `--ignore-paths=/livez,/readyz` should work fine. Let me know if that works for you.

This should be documented somewhere Within the values file. 
Or set per default when kubeRBACProxy is enabled?

fahedouch:
Hi @dotdc,
Is there a planned `kube-prometheus-stack` release including this patch ( [kube-state-metrics-5.29.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-state-metrics-5.29.0)) ?

refucktor:
After lots of trying I think there is a conceptual error, I even deploy the resources individually to confirm my case:

1. `kube-state-metrics` provides 2 types of server instances the `metrics` server and the `telemetry`, and there are proper arguments for it:

```yaml
...
  args:
    - "--host=127.0.0.1"
    - "--port=9090"
    - "--telemetry-host=127.0.0.1"
    - "--telemetry-port=9091"
```

2. on the `metrics` server they handle the following endpoints ([code](https://github.com/kubernetes/kube-state-metrics/blob/2a95d4649b2fea55799032fb9c0b571c4ba7f776/pkg/app/server.go#L432)): 
- `/metrics`
- `/livez`
- `/healthz`

Notice how there is no `/readyz` endpoint registered there

3. on the `telemetry` server, which is configured in a different port, they provide telemetry data of the software itself **and a `/readyz` endpoint** ([code](https://github.com/kubernetes/kube-state-metrics/blob/2a95d4649b2fea55799032fb9c0b571c4ba7f776/pkg/app/server.go#L378)):

- `/metrics`
- `/readyz`

4. on `kube-rbac-proxy` you can only configure one upstream, and it makes absolute sense to upstream to the `metrics` server endpoint. So when we configure the `kube-rbac-proxy` in this way:

```yaml
...
  args:
    - "--secure-listen-address=0.0.0.0:8080"
    - "--upstream=http://127.0.0.1:9090/"
    - "--ignore-paths=/healthz,/livez,/readyz"
...
```
ONLY the `/healthz` and `/livez` are forwarded to the upstream, the `/readyz` endpoint is in a port not considered by the `kube-rbac-proxy`

>--ignore-paths strings                        Comma-separated list of paths against which kube-rbac-proxy pattern-matches the incoming request. If the requst matches, it will proxy the request without performing an authentication or authorization check. Cannot be used with --allow-paths.

5. Here is my conclusion, we cannot have a readiness probe for `kube-state-metrics` when using `kube-rbac-proxy`, because it can only map ONE upstream, which is the `metrics` server, and the `/readyz` is provided by the `telemetry` server. Hence I only see 3 viable options

a) [Prefered] Open the telemetry port from the `kube-state-metrics`, then the readiness will bypass the proxy. The main idea of the rbac-proxy is to restrict access to the collected cluster metrics, which remains protected.
b) Use a different path in the readinessProbe of the kube-state-metrics container (either `/livez` or `/healthz`)

r0bj:
@refucktor The /readyz endpoint is available when `.Values.selfMonitor.enabled` is set to true. Here’s a working example for reference:
```
  kubeRBACProxy:
    enabled: true
    extraArgs:
    - --ignore-paths=/livez,/readyz

  selfMonitor:
    enabled: true
```

refucktor:
hey @r0bj, yeah I replied to you in the other thread, that's the perfect combination to have all probes enabled, no need for further modifications. Once again thx for your reply 🙌

PR #5245, opened 2025-01-28
[kube-prometheus-stack] rename additionalPrometheusRules prefix

TheRealNoob:

Rename prefix on `additionalPrometheusRules`.  I'm not sure why the prefix is currently set to the chart name and not the release name.  I feel like it should be the release name to avoid conflicts.  Thoughts?

I didn't open one.

Although I bumped the Major version, I don't see it as strictly necessary.  Yes the PrometheusRule objects will be destroyed and re-created under a new name, but to Operator that shouldn't make a difference.  Not unless you've written your own operator, or something unique like that.  But, it's just as easy to play it safe here and bump it anyway, so I have.

#### Checklist

- [x] Chart Version bumped


jkroepke (commented):
Hi, i'm on mobile right now, but maybe there is a git blame info to identity a reason for this. 

I have also the 63 char length in mind for kubernetes which is not covered yet.

From conflict point of view: the resources are namespaced scoped and not sure if they conflict.

TheRealNoob:
hmm that's a really good thought.  I had a look.  It looks like there's [only 2 commits on the history](https://github.com/prometheus-community/helm-charts/commits/main/charts/kube-prometheus-stack/templates/prometheus/additionalPrometheusRules.yaml) for this file.  The [first one](https://github.com/prometheus-community/helm-charts/commit/b643f58714a063e5a0ddf6832cd00ed29c0f4fab) was the migration 5 years from prometheus-operator to kube-prometheus-stack.  The [second one](https://github.com/prometheus-community/helm-charts/commit/09534e674f1fc2015a84e1ca268e09214a8bc347) is more interesting, but also 4 years ago.  It added a `metadata.name` to the List object and did implement `fullname` over `name`, but didn't touch the nested PrometheusRule objects.

Helm 3 came out 6 years ago and this history goes back nearly that far, so my guess would be it has been this way since the migration from helm 2.  Probably not a lot of people using this template.

TheRealNoob:
addressed the concerns about trunc 63.  I had thought about it as well, I just wanted to make sure maintainers would be open to this change.  I'm sure other changes need to be made in preparation for a Major bump?

TheRealNoob:
@jkroepke do you have any other concerns?  Are there any others that may need to be consulted?  If you think a major version bump makes the most sense, are there any other changes that need to be made?

charts/kube-prometheus-stack/templates/prometheus/additionalPrometheusRules.yaml
```diff
@@ -2,15 +2,15 @@
...
+  name: {{ printf "%s-%s" (include "kube-prometheus-stack.fullname" $) "additional-prometheus-rules" | trunc 63 | trimSuffix "-" }}
   namespace: {{ template "kube-prometheus-stack.namespace" . }}
 items:
 {{- if .Values.additionalPrometheusRulesMap }}
 {{- range $prometheusRuleName, $prometheusRule := .Values.additionalPrometheusRulesMap }}
   - apiVersion: monitoring.coreos.com/v1
     kind: PrometheusRule
     metadata:
-      name: {{ template "kube-prometheus-stack.name" $ }}-{{ $prometheusRuleName }}
+      name: {{ printf "%s-%s" (include "kube-prometheus-stack.fullname" $) $prometheusRuleName | trunc 63 | trimSuffix "-" }}
```
jkroepke, line 13: What did you think about if `$prometheusRule.name` is set, the property will be preferred over the default generated one?

If someone does not like the new name, the has at least the chance to mimic the old behavior.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -32,7 +32,7 @@ sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
 # in case of changes within CRDs, a major version bump is mandatory. See: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#upgrading-chart
-version: 68.3.3
+version: 69.0.0
```
jkroepke, line 35: Major is okay, but please add a notice in UPGRADING.md

jkroepke (changes requested):
LGTM in general. but I have some improvements here that may needs to be done here.

Also please resolve the conflicts.

charts/kube-prometheus-stack/templates/prometheus/additionalPrometheusRules.yaml
TheRealNoob: Well...but it's a mandatory field.  It's the `key` in the `key:value` dictionary.  At that point may as well remove `include "kube-prometheus-stack.fullname" $` right?  unless i'm missing something?

charts/kube-prometheus-stack/templates/prometheus/additionalPrometheusRules.yaml
```diff
@@ -2,15 +2,15 @@
 apiVersion: v1
 kind: List
```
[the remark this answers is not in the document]
TheRealNoob: Yeah I see no problem with that.  It will lead to a larger diff in things like `helm diff`, but that's already happening anyway so perfect timing.  Are you thinking just for purposes of simplifying the template?

TheRealNoob (commented):
@jkroepke sorry for taking so long to review this.  Things for hectic at work for a bit and when I didn't see any updates I assumed there was none.  Turns out I missed it.  Apologies for that.

charts/kube-prometheus-stack/Chart.yaml
TheRealNoob: will do

jkroepke (commented):
Overall, it looks good to me now.

Could you add some `additionalPrometheusRules` and `additionalPrometheusRulesMap` values to https://github.com/prometheus-community/helm-charts/blob/36a572a0f8db81cec4c9ea0ba619e45f01485707/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml

Then, this template should be covered by CI. At the moment, unit tests are not part of the CI yet.

TheRealNoob:
unittest added

PR #5276, opened 2025-02-05
kube-prometheus-stack bugfix wrong line break and add https redirect route

RenePinnow:

Bugfixing an issue by creating multiple routes for alertmanager, prometheus and thanosruler. In this case the linebreak does not work and the outcome is not a valid resource.

Adding new feature to create a route for https redirect see also https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#http-to-https-redirects

No open issue found that points to this bug and feature

@andrewgkew  @gianrubio @gkarthiks @GMartinez-Sisti @jkroepke @scottrigby @Xtigyro @QuentinBisson

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/route.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/route.yaml
@@ -2,7 +2,7 @@
   {{- $serviceName := printf "%s-%s" (include "kube-prometheus-stack.fullname" .) "prometheus" -}}
   {{- $servicePort := .Values.prometheus.ingress.servicePort | default .Values.prometheus.service.port -}}
   {{- range $name, $route := .Values.prometheus.route }}
-    {{- if $route.enabled -}}
+    {{- if $route.enabled }}
 ---
 apiVersion: {{ $route.apiVersion | default "gateway.networking.k8s.io/v1" }}
 kind: {{ $route.kind | default "HTTPRoute" }}
@@ -32,6 +32,13 @@ spec:
     {{- if $route.additionalRules }}
     {{- tpl (toYaml $route.additionalRules) $ | nindent 4 }}
     {{- end }}
+    {{- if $route.httpsRedirect }}
+    - filters:
+        - type: RequestRedirect
+          requestRedirect:
+            scheme: https
+            statusCode: 301
+    {{- else }}
     - backendRefs:
         - name: {{ $serviceName }}
           port: {{ $servicePort }}
@@ -44,5 +51,6 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     {{- end }}
+  {{- end }}
   {{- end }}
 {{- end }}
--- a/charts/kube-prometheus-stack/templates/prometheus/route.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/route.yaml
@@ -2,7 +2,7 @@
   {{- $serviceName := printf "%s-%s" (include "kube-prometheus-stack.fullname" .) "prometheus" -}}
   {{- $servicePort := .Values.prometheus.ingress.servicePort | default .Values.prometheus.service.port -}}
   {{- range $name, $route := .Values.prometheus.route }}
-    {{- if $route.enabled }}
+  {{- if $route.enabled }}
 ---
 apiVersion: {{ $route.apiVersion | default "gateway.networking.k8s.io/v1" }}
 kind: {{ $route.kind | default "HTTPRoute" }}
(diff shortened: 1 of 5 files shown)


charts/kube-prometheus-stack/templates/thanos-ruler/route.yaml
```diff
@@ -32,6 +32,13 @@ spec:
     {{- if $route.additionalRules }}
     {{- tpl (toYaml $route.additionalRules) $ | nindent 4 }}
     {{- end }}
+    {{- if $route.httpsRedirect }}
```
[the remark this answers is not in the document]
RenePinnow: @jkroepke 
Yes that is like expected. You see it also in the documentation from gateway api.

This is the example from the link I added in the description.

`apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: http-filter-redirect
spec:
  parentRefs:
  - name: redirect-gateway
    sectionName: http
  hostnames:
  - redirect.example
  rules:
  - filters:
    - type: RequestRedirect
      requestRedirect:
        scheme: https
        statusCode: 301`

charts/kube-prometheus-stack/values.yaml
```diff
@@ -4734,6 +4740,9 @@ thanosRuler:
       parentRefs: []
       # - name: acme-gw
 
+      # -- create http route for redirect (https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#http-to-https-redirects)
+      httpsRedirect: false
```
jkroepke, line 4744: Please add the info here, that if set to true, matches and filters (lines below) are ignored.

charts/kube-prometheus-stack/templates/prometheus/route.yaml
```diff
@@ -45,4 +52,5 @@ spec:
       {{- end }}
     {{- end }}
   {{- end }}
+  {{- end }}
```
jkroepke, line 55: please check, if the indent of the new end is correct.
I'm confused, because the new end is indent 2, while the {{- if $route.httpsRedirect }} is indent 4

charts/kube-prometheus-stack/values.yaml
RenePinnow: @jkroepke  
I added an additional comment.

charts/kube-prometheus-stack/templates/prometheus/route.yaml
RenePinnow: @jkroepke 
I fixed the indents.

RenePinnow:
@jkroepke 
Can you please have a look on my changes?

RenePinnow:
@andrewgkew @gianrubio @gkarthiks @GMartinez-Sisti @jkroepke @scottrigby @Xtigyro @QuentinBisson
Can someone please make the code review.

charts/kube-prometheus-stack/values.yaml
QuentinBisson: Can you add it to the other 2 as well?

charts/kube-prometheus-stack/values.yaml
```diff
@@ -4742,6 +4748,10 @@ thanosRuler:
       parentRefs: []
       # - name: acme-gw
 
+      ## create http route for https redirect (https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#http-to-https-redirects).
+      ## matches, filters and additionalRules will be ignored if this is set to true
+      httpsRedirect: false
```
jkroepke, line 4753: Sorry for doing one more review. GatewayAPI is still new to me and I'm doing review carefully.

Based on https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#http-to-https-redirects the code is correct.

However, just set `httpsRedirect=true` MAY leads to unexpected issues. (Ref: https://github.com/cilium/cilium/issues/28796)

I feel important that if `httpsRedirect=true` is set, the user must be take care that the HTTPRoute is enabled only on the http listener. Otherwise, infinite redirect may occurs. 

This should be a more bigger, visible warning.

RenePinnow: @jkroepke But for me this goes in the direction that the user must know what he is doing. If this problem occurs, it is not a critical error. You can no longer call up the prometheus UI at this moment and must then fix your routing accordingly. But it will not brake any core features of your monitoring. I will add an additional comment to warn the user of wrong usage.

jkroepke: I know what you mean. But please also have some understanding for the maintainer's point of view. People who don’t know what they’re doing create issues, and that disrupts us.

jkroepke (approved):

PR #5312, opened 2025-02-11
[snmp-exporter] Adding ability to add custom volumeName to config-reloader

boxstep:
Adding ability to set custom volumeName for config-reloader. since now it is hardcoded to "config" and does not watch other volumeMounts
With this approach you can create extraVolumes for example projected volumes which contain secrets for auth or snmp mib data and mount those configs to config reloader to watch and reload snmp-exporter.

Commit tested on working k8s cluster.

### DIFF
--- a/charts/prometheus-snmp-exporter/templates/deployment.yaml
+++ b/charts/prometheus-snmp-exporter/templates/deployment.yaml
@@ -104,7 +104,7 @@ spec:
 {{ toYaml .Values.configmapReload.resources | indent 12 }}
           volumeMounts:
             - mountPath: /etc/config
-              name: config
+              name: "{{ .Values.configmapReload.volumeName }}"
               readOnly: true
       {{- end }}
       {{- if .Values.imagePullSecrets }}
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 6.0.0
+version: 6.1.0
 appVersion: v0.27.0
 home: https://github.com/prometheus/snmp_exporter
 sources:
--- a/charts/prometheus-snmp-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-snmp-exporter/templates/daemonset.yaml
@@ -97,7 +97,7 @@ spec:
 {{ toYaml .Values.configmapReload.resources | indent 12 }}
           volumeMounts:
             - mountPath: /etc/config
-              name: config
+              name: "{{ .Values.configmapReload.volumeName }}"
               readOnly: true
       {{- end }}
       {{- if .Values.imagePullSecrets }}
--- a/charts/prometheus-snmp-exporter/values.yaml
+++ b/charts/prometheus-snmp-exporter/values.yaml
@@ -147,6 +147,10 @@ configmapReload:
   ## configmap-reload container name
   ##
   name: configmap-reload
+  ## volume name to mount to prometheus-config-reloader. Default "config"
+  ## Can use projected volume name in extraVolumes and mount multiple secrets, configmaps in single volume to watch for changes.
+  ##
+  volumeName: config
 
   ## configmap-reload container image
   ##


boxstep:
@xiu @walker-tom

walker-tom (changes requested):
Thanks for the PR, I think some further changes might be needed to make the configmap-reloader better. With this change you make it possible for different volumes other than `config` to be watched but if `.Values.config` isn't set then the configmap-reloader container doesn't get included in the Pod so you still need to populate `.Values.config` which could get confusing if it is then not used.

Instead I think the following changes could be combined to achieve what you want in a more scaleable and future proofed manner:
- Add an extra value `.Values.configmapReload.enabled` which controls whether or not the configmap-reload container gets added. Then we can enable it independently of setting `.Values.config`. 
- Add a value like `.Vales.configmapReload.extraVolumeMounts` which can then add the extra volumes that you've added to the Pod as desired.
- Add another value `.Values.configmapReload.extraArgs` which is an array and can be used to pass in extra arguments, like more `--watched-dir=...` flags to point at these extra volume mounts. 
- Wrap the existing configmap-reloader `volumeMounts` configuration in ` {{ if .Values.config }}` like is done down at the Pod `volumes` so that this is only mounted in if `.Values.config` is used and the corresponding `volume` `config` gets added.

If you start making some changes like these then I will be able to progress this change further and get it merged. Thanks!

boxstep:
@walker-tom 

Hello and thank you for reviewing my MR. I have made the following changes:

- Added .Values.configmapReload.enabled to control configmapReloader
- Added .Vales.configmapReload.extraVolumeMounts so configmapReloader could mount same volumes as snmp-exporter(Can use projected volumes to mount many configs in same directory)
- Added .Values.configmapReload.extraArgs to specify custom directories to watch
- Retained original if .Values.config for default directories previously used.

I had some problems with merge conflicts but seems OK now. Let's see if this can be merged or something more can be done:)

walker-tom:
@boxstep new changes look good and I've tested them out successfully as well. Please can you update `version` in `Chart.yaml` to be `9.0.0` and then add a new section to `README.md` like we did [for 8.0.0](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-snmp-exporter/README.md#to-800) that describes what has changed in this new major version so that users know how to upgrade, then we should be able to get this merged.

boxstep:
> @boxstep new changes look good and I've tested them out successfully as well. Please can you update `version` in `Chart.yaml` to be `9.0.0` and then add a new section to `README.md` like we did [for 8.0.0](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-snmp-exporter/README.md#to-800) that describes what has changed in this new major version so that users know how to upgrade, then we should be able to get this merged.

Changes made as requested. Chart version bumped to 9.0.0 and readme updated.

walker-tom:
> 
> @walker-tom Changes made as requested. Chart version bumped to 9.0.0 and readme updated.

Looks like there are some linting errors, if you check the files changed diff it shows which lines are erroring or you can view the github action logs. Can you take a look?

boxstep:
> > @walker-tom Changes made as requested. Chart version bumped to 9.0.0 and readme updated.
> 
> Looks like there are some linting errors, if you check the files changed diff it shows which lines are erroring or you can view the github action logs. Can you take a look?

@walker-tom fixed it, thanks!

charts/prometheus-snmp-exporter/templates/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+
```
walker-tom, line 1: Can you remove this extra line please.

walker-tom (changes requested):
One final change then everything looks good to merge.

charts/prometheus-snmp-exporter/templates/deployment.yaml
boxstep: removed

walker-tom (approved):
LGTM! Thanks for making all the additional requested changes, I think this puts the Helm Chart in a much more flexible and usable position.

boxstep:
> LGTM! Thanks for making all the additional requested changes, I think this puts the Helm Chart in a much more flexible and usable position.

Thank you aswell for your guidance, next time i will be able to contribute more effectively:)

PR #5313, opened 2025-02-12
[kube-prometheus-stack] Allow to configure kube-api-server service monitor path

lucaspouzac:
For EKS cluster, it's necessary to update path of serviceMonitor to scrape metrics (need to add /metrics).

I add this optional configuration on this serviceMonitor.


jkroepke:
I'm a bit confused here, since `/metrics` is the default value, if not defined.

lucaspouzac:
Maybe because I use alloy to scrape service monitor ?

jkroepke:
if so, it feels like an bug in alloy.

PR #5318, opened 2025-02-14
[kube-prometheus-stack] Add NetworkPolicy support for Alertmanager

### LINKED ISSUE #5247: [kube-prometheus-stack] Alertmanager needs networkPolicy

 Alertmanager needs networkPolicy as when networkPolicies are enabled, the ingress for alertmanager is forbidden with a 502.

Adding a networkPolicy for alertmanager with according parameters

Alternatively a key like extraDeploy can be implemented which deploys a manifest put into the key.

### Additional context.

_No response_

samuelarogbonlo:
#### Changes
- Added NetworkPolicy template for Alertmanager
- Added NetworkPolicy configuration options in values.yaml
- Added comprehensive unit tests
- Updated documentation

The NetworkPolicy configuration allows:

- Ingress from configured ingress controller
- Ingress from Prometheus pods
- Inter-pod communication for clustering
- Egress for DNS resolution
- Egress for SMTP notifications (ports 25, 587, 465)
- Egress for webhook notifications (ports 80, 443)

#### Testing
- Added unit tests for NetworkPolicy configuration
- Verified template rendering
- Tested with both default and custom configurations

#### Checklist
- [x] Chart Version bumped
- [x] DCO signed

Fixes #5247

### DIFF
--- a/charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
+++ b/charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
@@ -12,22 +12,22 @@ spec:
     matchLabels:
       app.kubernetes.io/name: alertmanager
   policyTypes:
-    - Ingress
-    - Egress
+    {{- toYaml .Values.alertmanager.networkPolicy.policyTypes | nindent 4 }}
   ingress:
-    # Allow ingress from the ingress controller
+    # Allow ingress from gateway
     - from:
         - namespaceSelector:
             matchLabels:
-              kubernetes.io/metadata.name: {{ .Values.alertmanager.networkPolicy.ingressNS }}
+              kubernetes.io/metadata.name: {{ .Values.alertmanager.networkPolicy.gateway.namespace }}
           podSelector:
             matchLabels:
-              {{- toYaml .Values.alertmanager.networkPolicy.ingressPodLabels | nindent 14 }}
+              {{- toYaml .Values.alertmanager.networkPolicy.gateway.podLabels | nindent 14 }}
       ports:
         - port: {{ .Values.alertmanager.service.port }}
           protocol: TCP
         - port: {{ .Values.alertmanager.service.clusterPort }}
           protocol: TCP
+    {{- if .Values.alertmanager.networkPolicy.monitoringRules.prometheus }}
     # Allow ingress from Prometheus
     - from:
         - podSelector:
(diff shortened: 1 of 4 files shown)


charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
```diff
@@ -0,0 +1,79 @@
...
+          protocol: TCP
+    # Allow ingress from Prometheus
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/name: prometheus
+      ports:
+        - port: {{ .Values.alertmanager.service.port }}
+          protocol: TCP
+  egress:
```
jkroepke, line 39: By default, egress should be not restricted. It can be an empty block from values and optionally filled via values.

For example, we have running a webhook pod inside the same namespace. Otherwise, end-users are using SMTP relay from a internal network. excluding internal network by default is not a good idea.

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
```diff
@@ -0,0 +1,79 @@
...
+    app: {{ template "kube-prometheus-stack.name" . }}-alertmanager
+    {{- include "kube-prometheus-stack.labels" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: alertmanager
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
```
jkroepke, line 17: loki can also sent alerts to alertmanager.

jkroepke (changes requested):

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
samuelarogbonlo: Fixed by making egress optional and configurable through egress.enabled: false by default

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
samuelarogbonlo: Added support through monitoringRules.loki: true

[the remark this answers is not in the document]
samuelarogbonlo: Already supported and made configurable via enableClusterRules: true

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
```diff
@@ -0,0 +1,79 @@
...
+    # Allow ingress from the ingress controller
+    - from:
```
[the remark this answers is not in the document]
samuelarogbonlo: Added additionalIngress for custom rules

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
```diff
@@ -0,0 +1,79 @@
...
+  name: {{ template "kube-prometheus-stack.fullname" . }}-alertmanager
+  namespace: {{ template "kube-prometheus-stack.namespace" . }}
+  labels:
+    app: {{ template "kube-prometheus-stack.name" . }}-alertmanager
+    {{- include "kube-prometheus-stack.labels" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: alertmanager
+  policyTypes:
```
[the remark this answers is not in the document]
samuelarogbonlo: Made policyTypes configurable with default as "Ingress" only

charts/kube-prometheus-stack/templates/alertmanager/networkpolicy.yaml
```diff
@@ -0,0 +1,80 @@
...
+    {{- toYaml .Values.alertmanager.networkPolicy.policyTypes | nindent 4 }}
+  ingress:
+    # Allow ingress from gateway
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Values.alertmanager.networkPolicy.gateway.namespace }}
+          podSelector:
+            matchLabels:
+              {{- toYaml .Values.alertmanager.networkPolicy.gateway.podLabels | nindent 14 }}
```
jkroepke, line 24: Question: what happens, if `.Values.alertmanager.networkPolicy.gateway.podLabels` is empty? Do we need to cover this scenario and omit the podSelector, if podLabels are empty?

samuelarogbonlo: Modified the template to omit podSelector when podLabels is empty or null. Added a test case that verifies this behavior by setting podLabels: null. The template now correctly handles cases where:

- podLabels is null
- podLabels is empty ({})
- podLabels is not set

jkroepke (approved):
LGTM

samuelarogbonlo:
> LGTM

@jkroepke Can I get a re-review for the workflow to run?

hornjo:
Thanks guys!

sebastiangaiser:
@jkroepke @samuelarogbonlo thanks for implementing this. 
May I ask why `monitoringRules.loki` is enabled by default as Loki is not a subchart of kube-prometheus-stack, so my recommendation would be to set it to `false`.
In addition why is `gateway.namespace` set to a static namespace (`ingress-nginx`)? Would it make sense to make this configurable (e.g. enable/disable) and set the default to an empty string like for the Ingress host?
Of course I can adjust them on my own but I think the default should be "more secure". Please let me know about your thoughts.

jkroepke:
@sebastiangaiser I see and accept your concern. I overlook the values part a bit. 

Do you want to create a follow up PR that solves the issues you mention?

I agree with loki and ingress should not enabled by default

samuelarogbonlo:
@sebastiangaiser I understand your concerns regarding the default values. I initially set `monitoringRules.loki: true` and `gateway.namespace: "ingress-nginx"` thinking they would provide a smoother out-of-box experience for common setups, but you're right that these defaults make too many assumptions about the deployment environment. You can submit a follow-up PR or I do if you want `monitoringRules.loki` to `false` and make `gateway.namespace` more configurable with an empty default value, which will provide better flexibility while maintaining security.

sebastiangaiser:
Feel free to go ahead. I probably won't make it in time.

PR #5319, opened 2025-02-14
[kube-prometheus-stack] Fix: Add default values to subchart to allow rendering

### LINKED ISSUE #5307: [kube-prometheus-stack crds] Manifest generation error : Failed to unmarshal "clusterrole.yaml": <nil> while upgrading to v69.2.2

Via Argocd, while upgrading crds from 67.11.0 to 69.2.2 I get the following error : 

```Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): rpc error: code = FailedPrecondition desc = Failed to unmarshal "clusterrole.yaml": <nil>```

### What's your helm version?

not relevant, using argocd version: v2.12.9

Client Version: v1.31.3 Kustomize Version: v5.4.2 Server Version: v1.31.4-eks-2d5f260

### Which chart?

crds

### What's the chart version?

69.2.2

### What happened?

_No response_

_No response_

### How to reproduce it?

Upgrade crds from 67.11.0 to 69.2.2 with argocd

NONE

argocd sync with default values

Before that last upgrade, I upgraded all the way from 57 to 67 without any issue.

I suspect there might be an issue in crds/templates/upgrade/clusterrole.yaml, specifically the namespace metadata which should not be needed. However I'm not sure if this would cause an unmarshalling error

### PR DESCRIPTION
**Pull request by jcrsilva**

The kube-prometheus-stack CRD subchart fails to render on the current version with

```
Error: template: crds/templates/upgrade/serviceaccount.yaml:1:18: executing "crds/templates/upgrade/serviceaccount.yaml" at <.Values.upgradeJob.enabled>: nil pointer evaluating interface {}.enabled
```

This is because of the preview upgradejob feature that was added:

> This feature is in preview, off by default and may change in the future.

The CRD subchart has no default `values.yaml` and sets no defaults itself for statements like

```
{{- if .Values.upgradeJob.enabled }}
```

Which produces the error above.

This PR adds a default values.yaml ensuring rendering and a unittest to prevent regression.

#### Note

I understand the intent might not be for people to install this CRD chart directly, but instead render it through the main `kube-prometheus-stack` chart, but people have taken to using it considering the linked issue. It's easier to point tools like Argo at this subchart to render the CRDs before installing the main chart.

When the feature leaves preview, it will be a moot point, but until then it would be nice if the chart maintained its original behavior and rendered without issues.

I bumped the main `kube-prometheus-stack` version since this chart's seems to be locked to `0.0.0`.

- fixes #5307

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/charts/crds/values.yaml
+++ b/charts/kube-prometheus-stack/charts/crds/values.yaml
@@ -1,0 +1,4 @@
+## Check out kube-prometheus-stack/values.yaml for more information
+## on this parameter
+upgradeJob:
+  enabled: false
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 69.2.4
+version: 69.2.5
 appVersion: v0.80.0
 kubeVersion: ">=1.19.0-0"
 home: https://github.com/prometheus-operator/kube-prometheus
--- a/charts/kube-prometheus-stack/charts/crds/unittests/rendering_test.yaml
+++ b/charts/kube-prometheus-stack/charts/crds/unittests/rendering_test.yaml
@@ -1,0 +1,6 @@
+suite: test rendering
+tests:
+  # 5307
+  - it: should render when no values are set
+    asserts:
+      - notFailedTemplate: {}

### CODE REVIEW

File: charts/kube-prometheus-stack/charts/crds/values.yaml [hunk 1]
```diff
@@ -0,0 +1,7 @@
+enabled: true
```
jkroepke, line 1:
```suggestion
```
Shoud

charts/kube-prometheus-stack/charts/crds/values.yaml
```diff
@@ -0,0 +1,7 @@
...
+## The CRD upgrade job mitigates the limitation of helm not being able to upgrade CRDs.
```
jkroepke, line 2: I would like to avoid the documentation twice. Can we say something like checkout kube-prometheus-stack values.yaml for more infomation?

jcrsilva: Done

jkroepke (approved):

jkroepke (approved):

PR #206, opened 2020-10-13
[WIP][kube-prometheus-stack] Support native TLS in prometheus-operator

mrueg:
#### What this PR does / why we need it:
Allow native TLS supported by prometheus-operator v0.39.0 and later 

#### Special notes for your reviewer:
This is not tested yet and still WIP. 
To be discussed: Deprecate tlsproxy to simplify chart?

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #5333, opened 2025-02-18
[kube-prometheus-stack] Fix TLS condition scope

sam-mcbr:

Fixes the scope of the `.Values.prometheusOperator.tls.enabled` condition introduced in #206. The current scoping prevents the use of `extraArgs` and `lifecycle` when not enabling TLS.

While this is a bug fix/patch, this will change the behavior of the deployment for any users who may have been setting the `extraArgs` and `lifecycle` values without enabling TLS. Let me know if I should bump the minor version instead of the patch version.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 69.3.1
+version: 69.3.2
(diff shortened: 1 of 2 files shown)


sam-mcbr:
@scottrigby @gkarthiks Would really appreciate a quick review on this as not being able to use `extraArgs` is a blocker for a feature our team is rolling out (we do not enable `tls`). Thank you so much for any feedback!

jkroepke (approved):
Keep in mind, if TLS is disable, Validating Webhook may not work (since they require TLS to my best knowledge)

PR #5338, opened 2025-02-20
[prometheus-mongodb-exporter] Fix unnecessary required auth when doing multi-target

### LINKED ISSUE #5337: [prometheus-mongodb-exporter] Password should not be required when doing multi-target

When using `serviceMonitor.multiTarget.enabled: true` then `serviceMonitor.multiTarget.username` is required.

Using multi-target should not require auth.

### What's your helm version?

version.BuildInfo{Version:"v3.17.1", GitCommit:"v3.17.1", GitTreeState:"", GoVersion:"go1.23.5"}

Client Version: v1.32.2

### Which chart?

prometheus-mongodb-exporter

### What's the chart version?

3.11.0

### What happened?

1. Have multiple MongoDB nodes in a MongoDB cluster that does not use authentication from which I want to export metrics
2. Set `serviceMonitor.multiTarget.enabled: true`
3. Set multiple targets in `serviceMonitor.multiTarget.targets`
4. Get error `The MongoDB username must be set` when trying to do `helm template` or `helm install`

It should work without auth

### How to reproduce it?

Try to template the chart, and it fails.

```yaml
serviceMonitor:
  enabled: true
  multiTarget:
    enabled: true
    targets:
    - uri: "mongodb://mongodb-0.staging.internal:27017/admin"
      name: "mongodb-0"
    - uri: "mongodb://mongodb-1.staging.internaltld:27017/admin"
      name: "mongodb-1"
```

```bash
helm template ./charts/prometheus-mongodb-exporter
```

To get around this I've disabled the chart's ServiceMonitor via `serviceMonitor.enabled: false` and instead deployed my own ServiceMonitor, and instead used the following values in my `values.yaml`:

```yaml
mongodb:
  uri: mongodb://mongodb-0.staging.internal:27017/admin,mongodb://mongodb-1.staging.internaltld:27017/admin

serviceMonitor:
  enabled: false
  multiTarget:
    enabled: false
```

This is tedious as I have to create that `mongodb.uri` manually, and in our actual use case I need to target 6 hosts and so the `mongodb.uri` becomes a really long string, which is super unpleasant to manage. (but works as a workaround)

I tried deploying this workaround and it works just fine when doing multi-target without auth.

applejag:

Removes arbitrary restriction of requiring auth in mongodb_exporter when doing multi-target.

- fixes #5337

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mongodb-exporter/.helmignore
+++ b/charts/prometheus-mongodb-exporter/.helmignore
@@ -20,3 +20,5 @@
 .idea/
 *.tmproj
 .vscode/
+
+ci/
--- a/charts/prometheus-mongodb-exporter/Chart.yaml
+++ b/charts/prometheus-mongodb-exporter/Chart.yaml
@@ -17,4 +17,4 @@ maintainers:
 name: prometheus-mongodb-exporter
 sources:
   - https://github.com/percona/mongodb_exporter
-version: 3.11.0
+version: 3.11.1
--- a/charts/prometheus-mongodb-exporter/ci/default-values.yaml
+++ b/charts/prometheus-mongodb-exporter/ci/default-values.yaml
@@ -1,0 +1,1 @@
+# default values test case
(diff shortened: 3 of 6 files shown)


zeritti (approved):
Thank you, @applejag, for your PR, LGTM.

PR #5342, opened 2025-02-21
[prometheus-smartctl-exporter] Extend PrometheusRule templating

hedgieinsocks:

- Improves rule declaration flexibility

Currently, the end user can either enable the default rule set or disable it. This PR declares rules in values file (like in other charts) instead of sourcing an external file. Because rules are now a dictionary, one could easily tweak each rule or add new ones.

- Normalizes labels

Some alert rules have severity labels set to `error` which I believe is not very common. This PR changes `error` severity labels to `critical`.  I consider it as a fix, and not a breaking change.

- Normalizes annotations

I believe it is common to use `summary` and `description` annotations. This PR adds both of them. 

I'm aware that other charts use `prometheusRule`  (without `s` at the end) key. But I left it as it is to avoid breaking changes.

#### Checklist

- [x] Chart Version bumped


hedgieinsocks:
@kfox1111 Hello, the workflow tagged only your colleague, but perhaps you could attend to it as well

kfox1111:
Hmm... Interesting.

Originally, I put it in a File rather then the templates, as the double golang engine passes (helm vs prometheus) made it kind of confusing to write rules. Have you hit a use case where it is useful to have the helm templateing applied to the values?

The other thing to discuss is the values api...
Its pretty straight forward to write the chart with arrays, but as a user of the chart, helms merging behavior of arrays (it doesnt) is unfortunate. You can't easy configure any one thing, but have to basically disable everything and copy/paste the rules back in.

We faced a similar thing in the spire chart, and used a dictionary instead. Then you can put 'enabled' on them so they can be individually turned on / off, and properties of them can be overriden by the users own values file.

See this for how it looks over there:
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/values.yaml#L604

What about tweaking the rules to have that kind of structure? It may even be possible then to have a 'template: true" property on a rule (or something like that), that enables/disables helm templating of that particular rule, so we get both options. Easier to understand if you dont need templating, and full templating support if you doo.

hedgieinsocks:
Thank you for your reply! Honestly, what I needed is to add common summary + description annotations and fix severity labels. The move to values I added only after taking a look at other charts here, and presuming it was some kind of standard. If you are ok with suggested annotation/label changes, we could keep the existing files approach, no problem. If  you would like me to attempt your dictionary idea, I'll certainly try ;)

kfox1111:
Yeah, it may make sense to break it up into two pr's then. One for the annotations/labels and one for the rest. That way we can merge your annotation/label changes quickly. It was kind of lost in the pr with so many other changes, I didn't really notice that yet.

As for the second part, if you have time to make an attempt at that, great. :)  if not, I'll probably take a stab at it eventually. Thanks! :)

hedgieinsocks:
I appreciate your cooperation. I'm not in a hurry, so let's try to implement that dictionary approach. If I fail, then I'll open another pr to update labels/annotations. The latest commit is the first stub at blind dictionary logic, which should allow precise tuning. And I kinda quite like it :)

kfox1111:
Yeah. Liking that more already. so, say, you have an issue with something flapping a little bit, you can:
```
prometheusRules:
  rules:
    smartCTLDInterfaceSlow:
      for: 10m
```

without having to rewrite everything.

charts/prometheus-smartctl-exporter/templates/prometheusrule.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
 {{- end }}
 spec:
   groups:
-  - name: smartctl-exporter.rules
-{{ .Files.Get "rules/rules.txt" | indent 4 }}
+    - name: {{ template "prometheus-smartctl-exporter.name" $ }}
+      rules:
+      {{- range .Values.prometheusRules.rules }}
+        - {{ toYaml . | nindent 10 | trim }}
```
kfox1111, line 19: how about something like (untested):
```suggestion
        {{- if .enabled }}
        - {{ toYaml (unset (deepCopy .) "enabled" ) | nindent 10 | trim }}
        {{- end }}
```

hedgieinsocks: huh, what a nice trick! it seems to template as expected

charts/prometheus-smartctl-exporter/templates/prometheusrule.yaml
```diff
@@ -13,6 +13,11 @@ metadata:
 {{- end }}
 spec:
   groups:
-  - name: smartctl-exporter.rules
-{{ .Files.Get "rules/rules.txt" | indent 4 }}
+    - name: {{ template "prometheus-smartctl-exporter.name" $ }}
+      rules:
+      {{- range .Values.prometheusRules.rules }}
+        {{- if .enabled }}
+        - {{ toYaml (unset (deepCopy .) "enabled" ) | nindent 10 | trim }}
```
kfox1111, line 20: Ok... thought about this a bit more.... what if we do something like this (untested). Top level has properties:
enabled, template, params, and/or raw. If you just want to do the easy thing, you put the bits under raw and it works without needing to escape gotl using gotl. If you want to use the template engine, you can do that easily too by just putting stuff under template 
```suggestion
        {{- if hasKey . "template" }}
        - {{ tpl (toYaml .template) (dict "Params" .params) | nindent 10 | trim }}
        {{- else }}
        - {{ toYaml .raw | nindent 10 | trim }}
        {{- end }}
```

Then we could have a rule like:
```
    smartCTLDDeviceTemperature:
      enabled: true
      template:
        alert: SmartCTLDDeviceTemperature
        expr: smartctl_device_temperature{temperature_type="current"} > {{ .Params.maxTemp }}
        for: 1m
        annotations:
          summary: Device has temperature higher than {{ .Params.maxTemp }}°C
          description: |
            Device {{ "{{ $labels.device }}" }} on instance {{ "{{ $labels.instance }}" }} has temperature higher than {{ .Params.maxTemp }}°C
        labels:
          severity: warning
      params:
        maxTemp: 60
```

hedgieinsocks: I'm afraid I don't like this approach. It's subjective, but the new layer or abstraction feels like unnecessary complication, both confusing and limiting. I understand that you don't like the very idea of escaping console templates, and I totally agree that this is unfortunate. But looking at the neighbors, this is an accepted drawback, and people will have to learn it anyway. Perhaps we could append something like this after the last rule to give the consumer clear instructions how they can do their customization without wasting time to google/chatgpt how to do proper escaping ;D

```yaml
    ##### You can declare additional custom rules:
    # smartCTLDDeviceMyCustomAlert:
    #   enabled: true
    #   alert: SmartCTLDDeviceMyCustomAlert
    #   expr: this_is_my_custom_prometheus_expression > 0
    #   for: 1m
    #   annotations:
    #     summary: This is my custom alert summary
    #     description: |
    #       Device {{ "{{ $labels.device }}" }} on instance {{ "{{ $labels.instance }}" }} is not ok
    #   labels:
    #     severity: warning
    ##### Note that you must properly escape console templates, so the below statement:
    ### summary: Device {{ $labels.device }} is not ok
    ##### Must be instead expressed in one of the following ways:
    ### summary: Device {{ "{{ $labels.device }}" }} is not ok
    ### summary: Device {{`{{`}} $labels.device {{`{{`}} is not ok
```

kfox1111: What is the value of running values through the helm template engine?

I see value in it if you can provide parameters to the template. If so, we would need a designated place where those parameters could live or else the chart can become very hard for us to maintain. Users would otherwise inject arbitrary values that would break if we needed to put our own value in place and we would have no way to know what values they would be using. By designating a place to put them, we guarantee we wont stomp on those values during upgrade.

hedgieinsocks: Hmm, we could template our rules like this:

```yaml
      rules:
      {{- range .Values.prometheusRules.rules }}
        {{- if .enabled }}
        - alert: {{ .alert | required "A rule requires a valid `alert` clause" }}
          expr: {{ .expr | required "A rule requires a valid `expr` clause" }}
          {{- with .for }}
          for: {{ . }}
          {{- end }}
          {{- with .keep_firing_for }}
          keep_firing_for: {{ . }}
          {{- end }}
          {{- with .annotations }}
          annotations:
            {{- toYaml . | nindent 12 }}
          {{- end }}
          {{- with .labels }}
          labels:
            {{- toYaml . | nindent 12 }}
          {{- end }}
        {{- end }}
      {{- end }}
```

kfox1111: Hmm.. That could help with some of it, yeah.

I'm talking more about the `tpl` function though. Do you still want to run everything through it?

if so, I think there needs to be a place for the values that are passed to it.

If not:
```
Device {{ "{{ $labels.device }}" }} on instance {{ "{{ $labels.instance }}" }} is not ok
```

should just be:
```
Device {{ $labels.device }} on instance {{ $labels.instance }} is not ok
```
as without going through `tpl`, it wont do the substitutions?

hedgieinsocks: Oh, I guess I don't need and don't want to use `tpl` at all. It's just was something I blindly picked up from the neighbor charts over here. So does the latest commit look closer to the desired state?

kfox1111: Looks good to me then.

Will need to fix the DCO and then we can merge this I think.

charts/prometheus-smartctl-exporter/Chart.yaml
```diff
@@ -21,7 +21,7 @@ version: 0.14.0
 # incremented each time you make changes to the application. Versions are not expected to
```
kfox1111, line 21: Need to bump:
```
version: 0.14.0
```
to
```
version: 0.15.0
```
above

kfox1111 (approved):
LGTM. Thanks! :)

PR #5379, opened 2025-02-27
kube-prometheus-stack/prometheus-adapter/prometheus-operator-admission-webhook-feat(chart): add support for revisionHistoryLimit on the cert

Aransh:
#### Which issue this PR fixes

Seeing lots of rotations on the certificate for the webhook. So want to mitigate the problem first by providing a revisionHistoryLimit on the cert according to https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #5394, opened 2025-03-02
[prometheus-operator-admission-webhook] feat(chart): add support for revisionHistoryLimit on the cert

Aransh:

Seeing lots of rotations on the certificate for the webhook. So want to mitigate the problem first by providing a revisionHistoryLimit on the cert according to https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec

- replaces https://github.com/prometheus-community/helm-charts/pull/5379

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-operator-admission-webhook/Chart.yaml
+++ b/charts/prometheus-operator-admission-webhook/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.19.2
+version: 0.20.0
 appVersion: 0.80.1
 home: https://github.com/prometheus-operator/prometheus-operator
 icon: https://github.com/prometheus-operator/prometheus-operator/raw/main/Documentation/logos/prometheus-operator-logo.png
--- a/charts/prometheus-operator-admission-webhook/Chart.yaml
+++ b/charts/prometheus-operator-admission-webhook/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.19.1
+version: 0.19.2
 appVersion: 0.80.1
 home: https://github.com/prometheus-operator/prometheus-operator
 icon: https://github.com/prometheus-operator/prometheus-operator/raw/main/Documentation/logos/prometheus-operator-logo.png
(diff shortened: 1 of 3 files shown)


charts/prometheus-operator-admission-webhook/values.yaml
```diff
@@ -110,9 +110,17 @@ certManager:
   ## rootCert allows setting certificate's lifetime when creating an issuer, defaults to 5y
   rootCert: {}
     # duration: "43800h0m0s"
+    # -- Set the revisionHistoryLimit on the Certificate. See
+    # https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec
+    # Defaults to nil.
+    # revisionHistoryLimit: ""
```
zeritti, line 116: The field's type is _int32_, it should not be given as _string_.

charts/prometheus-operator-admission-webhook/values.yaml
```diff
@@ -110,9 +110,17 @@ certManager:
...
   ## webhookCert allows setting webhook certificate's lifetime, defaults to 1y
   webhookCert: {}
     # duration: "8760h0m0s"
+    # -- Set the revisionHistoryLimit on the Certificate. See
+    # https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec
+    # Defaults to nil.
+    # revisionHistoryLimit: ""
```
zeritti, line 123: As above

charts/prometheus-operator-admission-webhook/Chart.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.19.1
+version: 0.19.2
```
zeritti, line 5: The change should be considered a new feature:

```suggestion
version: 0.20.0
```

zeritti (changes requested):
Thank you, @Aransh, for your PR. Please, see my comments below.

charts/prometheus-operator-admission-webhook/values.yaml
Aransh: It was given an empty string as "nil", I can change it to just a blank value though (from docs - "If set, revisionHistoryLimit must be a value of 1 or greater. If unset (nil), revisions will not be garbage collected. Default value is nil.")

zeritti (approved):
Thanks, @Aransh, LGTM.

PR #5397, opened 2025-03-02
[kube-prometheus-stack] Add nameValidationScheme field to Prometheus CR

sviatlo:
This PR adds support for configuring `nameValidationScheme` in Prometheus CR.
Setting this parameter to `Legacy` should result in prometheus-operator rendering `metric_name_validation_scheme: legacy` in config, thus configuring prometheus to request metrics/labels in the legacy way. 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -4503,6 +4503,10 @@ prometheus:
     ## in Prometheus so it may change in any upcoming release.
     allowOverlappingBlocks: false
 
+    ## Specifies the validation scheme for metric and label names.
+    ## Supported values are: Legacy, UTF8
+    nameValidationScheme: ""
+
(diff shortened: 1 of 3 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -4503,6 +4503,10 @@ prometheus:
     ## in Prometheus so it may change in any upcoming release.
     allowOverlappingBlocks: false
 
+    ## Specifies the validation scheme for metric and label names.
+    ## Supported values are: Legacy, UTF8
+    nameValidationScheme: Legacy
```
QuentinBisson, line 4508: I would probably make this value an empty string by default though so that when the default behavior changes upstream of it does we follow up with the intent

QuentinBisson:
Thank you :)

QuentinBisson (dismissed):

QuentinBisson (approved):

PR #5405, opened 2025-03-06
[kube-prometheus-stack] Add interval field for default dashboards

ecd0011:

This PR adds support for configuring the interval value used in Grafana dashboards through a new Helm value `grafana.defaultDashboardsInterval`.

Many Grafana dashboards in the chart have hardcoded `"interval": "1m"` values, which limits the resolution of metrics displayed regardless of your Prometheus scrape interval. If you're using a lower scrape interval (like 10s or 30s), you're missing out on higher resolution data that's already being collected.

This allows you to configure by making the interval configurable via Helm values, similar to how dashboard editability is already handled using the `sync_grafana_dashboards.py` script.

#### Checklist

- [x] Chart Version bumped


QuentinBisson:
Looks quite good to me

QuentinBisson (approved):

QuentinBisson (approved):

QuentinBisson:
Can you please fix the linting?

QuentinBisson (approved):

QuentinBisson:
@ecd0011 would you be able to do the same for the `$__rate_interval`?

ecd0011:
@jkroepke It seems like they're serving different purposes. 

`interval` in terms of this PR is adjusting how often the queries themselves are ran. `$__rate_interval` is only supported for Prometheus data sources and is meant to be used in the rate function for queries: https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#__rate_interval

In other words, `interval` is determining when the query is executed, but not what the query retrieves. `$__rate_interval` helps determine what the query retrieves, but not how often it's executed.

@QuentinBisson given the documentation @jkroepke recommends to use the `$__rate_interval` for certain query functions, I'd suggest leaving it using that variable and letting Prometheus determine what the value of that should be.

PR #5406, opened 2025-03-06
[prometheus-yet-another-cloudwatch-exporter] Allow to override image registry

szibis:
For those using some private image repositories, the current configuration always adds `/` even if the repository is empty.

This PR add helper and global configuration for the repository but allows to work with both private repositories or public quai.io.

#### Checklist

- [x] Chart Version bumped


charts/prometheus-yet-another-cloudwatch-exporter/values.yaml
```diff
@@ -2,10 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+global:
+  imageRegistry: ""
```
szibis, line 6:
```suggestion
  imageRegistry: "quay.io"
```

asherf:
FWIW, this introduces a breaking change. So I think it probably better to just fix the issue and not introduce a breaking change .
Since this chart only consumes a single image from a single repo, it doesn’t make a ton of sense to me to have a globa setting for the registry. We can leave the registry setting at its current location

cristiangreco:
> FWIW, this introduces a breaking change. So I think it probably better to just fix the issue and not introduce a breaking change . Since this chart only consumes a single image from a single repo, it doesn’t make a ton of sense to me to have a globa setting for the registry. We can leave the registry setting at its current location

I tend to agree with that. @szibis do you mind reworking the implementation?

szibis:
> > FWIW, this introduces a breaking change. So I think it probably better to just fix the issue and not introduce a breaking change . Since this chart only consumes a single image from a single repo, it doesn’t make a ton of sense to me to have a globa setting for the registry. We can leave the registry setting at its current location
> 
> I tend to agree with that. @szibis do you mind reworking the implementation?

Yes thank for the feedback. Additional changes added.

charts/prometheus-yet-another-cloudwatch-exporter/templates/_helpers.tpl
```diff
@@ -59,3 +59,15 @@ Create the name of the service account to use
...
+Return the correct image registry.
+*/}}
+{{- define "yet-another-cloudwatch-exporter.image.repository" -}}
+  {{- $registry := .Values.image.registry -}}
+  {{- if $registry -}}
+    {{- printf "%s/%s" $registry .Values.image.repository -}}
+  {{- else -}}
+    {{- printf "%s" .Values.image.repository -}}
+  {{- end }}
+{{- end -}}
```
szibis, line 73:
```suggestion
{{- end -}}

```

szibis:
Anything else I should do here @cristiangreco @asherf

asherf:
> Anything else I should do here @cristiangreco @asherf 

The DCO thing need to be resolved before this can be merged

szibis:
Closing this PR because have some issues with local cert will back once again with new PR

Issue #2181, opened 2022-06-21
[kube-state-metrics] some metrics appear to be duplicated


tomsucho:
### Describe the bug a clear and concise description of what the bug is.

Hello,

getting duplicate like metrics, e.g. some labels are a bit different but it is one and the same thing:
- kube_daemonset_status_number_ready{app_kubernetes_io_component="metrics", app_kubernetes_io_instance="prometheus", app_kubernetes_io_managed_by="Helm", app_kubernetes_io_name="kube-state-metrics", app_kubernetes_io_part_of="kube-state-metrics", app_kubernetes_io_version="2.3.0", daemonset="fluent-bit", deployment_level="cluster_overlays", helm_sh_chart="kube-state-metrics-4.5.0", instance="10.0.67.200:8080", job="kubernetes-service-endpoints", kubernetes_name="prometheus-kube-state-metrics", kubernetes_namespace="forge", kubernetes_node="ip-10-0-67-143.eu-west-1.compute.internal", namespace="fluent-bit", release="prometheus", team_owner="forge"}
- kube_daemonset_status_number_ready{container="kube-state-metrics", daemonset="fluent-bit", endpoint="http", instance="10.0.67.200:8080", job="kube-state-metrics", namespace="fluent-bit", pod="prometheus-kube-state-metrics-795b7f97bf-fqtrd", service="prometheus-kube-state-metrics"}

the latter one has endpoint="http" and pod="prometheus-kube-state-metrics-795b7f97bf-fqtrd" and I think is result of a serviceMonitor that we got configured for it, but wonder why this job="kubernetes-service-endpoints" gets to scrape the endpoints in addition to the job="kube-state-metrics" since in Prometheus cfg we got the below, and the endpoints are not having prometheus.io/scrape: "true" annotation on them.
```- job_name: kubernetes-service-endpoints
  kubernetes_sd_configs:
  - role: endpoints
  relabel_configs:
  - action: keep
    regex: true
    source_labels:
    - __meta_kubernetes_service_annotation_prometheus_io_scrape
```

### What's your helm version?

v3.7.1

### What's your kubectl version?

v1.21.12

### Which chart?

kube-state-metrics

### What's the chart version?

kube-state-metrics-4.5.0

### What happened?

Two endpoints are defined in the servicemonitor for kube-state-metrics. Based on my observations it results in duplicating the metrics.

### What you expected to happen?

kube-state-metrics not being duplicated.

### How to reproduce it?

[...]

PR #5470, opened 2025-03-22
[kube-state-metrics] Use scrapeConfig to have HA

jenciso:

This PR enables scrapeConfig to be used as alternative to serviceMonitor for kube-state-metrics.

We want to have a way to configure the kube-state-metrics in High Availiability. The simple way to do it is increasing the number of `replicas` to 2. However, it duplicates the kube-state-metrics metrics. 

With the use of scrapeConfig resource instead of serviceMonitor, you can scrape the kube-state-metrics as a traditional kubernetes services where each replica will expose its endpoint port giving high availability 

- https://github.com/kubernetes/kube-state-metrics/issues/611

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 5.31.0
+version: 5.32.0
 appVersion: 2.15.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -8,6 +8,7 @@ keywords:
   - kubernetes
 type: application
 version: 5.32.0
+# renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.15.0
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -239,6 +239,46 @@ prometheus:
         # name: secret-name
         # key:  key-name
       tlsConfig: {}
+  ## Create a scrapeConfig resource for scraping the kube-state-metrics service. Use this instead of serviceMonitor
+  ## to have more instances of kube-state-metrics safety.
+  scrapeconfig:
+    ## To avoid duplicate metrics, first disable the serviveMonitor creation via prometheus.monitor.enabled=false
+    enabled: false
+    annotations: {}
+    additionalLabels: {}
+    jobName: kube-state-metrics
+    ## SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
+    ##
+    sampleLimit: 0
+
+    ## TargetLimit defines a limit on the number of scraped targets that will be accepted.
+    ##
+    targetLimit: 0
+
+    ## Per-scrape limit on number of labels that will be accepted for a sample. Only valid in Prometheus versions 2.27.0 and newer.
+    ##
+    labelLimit: 0
+
+    ## Per-scrape limit on length of labels name that will be accepted for a sample. Only valid in Prometheus versions 2.27.0 and newer.
+    ##
+    labelNameLengthLimit: 0
+
+    ## Per-scrape limit on length of labels value that will be accepted for a sample. Only valid in Prometheus versions 2.27.0 and newer.
+    ##
+    labelValueLengthLimit: 0
+
+    ## StaticConfigLabels defines the labels to be used in the Prometheus static configuration for scraping.
+    staticConfigLabels: {}
+    scrapeInterval: ""
+    scrapeTimeout: ""
+    proxyUrl: ""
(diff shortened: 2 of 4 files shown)


jkroepke:
Whats the issue with double metrics? If could drop instance pod label to equalize them.

jenciso:
> Whats the issue with double metrics? If could drop instance pod label to equalize them.

Hi @jkroepke, The issue with double metrics is cost (storage) and memory efficiency (cardinality), why need I double metrics when it is possible to have only one? 
Another point is, why could we do extra work, when you could handle this service differently, as an alternative to serviceMonitor. The approach works, and it is being adopted (See [here](https://github.com/prometheus-community/helm-charts/issues/2181#issuecomment-2668644839) an example)

nicolastakashi:
I’m trying to think about possible issues that require the usage of KSM with multiple replicas.

Resource Constraints:

KSM is using more resources (especially memory) than the defined limits and it’s crash looping due to OOM. I might be wrong, but multiple replicas won’t improve reliability in that case, since both instances will be collecting the same data and using the same resource limits.

**Rollout of new versions:**

If you’re rolling out a new version and for some reason it has a startup problem, it’s true that multiple replicas might help here, because the second replica won’t be rolled out until the first one is ready and healthy.

But in my opinion, we can achieve the same behavior by properly configuring the deployment strategy, setting configs like minReadySeconds and others.

**Reschedules:**

Multiple replicas might also help with pod reschedules, but we can achieve stability using something like PDBs to avoid ending up with zero replicas.

If we really want to increase the availability of KSM, maybe we can consider the sharding feature, so you can have multiple KSM instances, each one collecting metrics from specific APIs.

jkroepke:
@nicolastakashi I see still a value for @jenciso requests. It's a low budget version of HA where the cluster is not big enough to need the sharing mode, but you will not lose metrics on rescheduling nodes. 

For scrape requests against blackbox_exporter, the same principal is used.

jenciso:
> I’m trying to think about possible issues that require the usage of KSM with multiple replicas.
> 
> Resource Constraints:
> 
> KSM is using more resources (especially memory) than the defined limits and it’s crash looping due to OOM. I might be wrong, but multiple replicas won’t improve reliability in that case, since both instances will be collecting the same data and using the same resource limits.
>
Maybe I could wrong. But, we could say the same for prometheus-HA mode stratey, where each prometheus server scrape and save the same content and share the same memory limits.
> 
> **Rollout of new versions:**
> 
> If you’re rolling out a new version and for some reason it has a startup problem, it’s true that multiple replicas might help here, because the second replica won’t be rolled out until the first one is ready and healthy.
> 
> But in my opinion, we can achieve the same behavior by properly configuring the deployment strategy, setting configs like minReadySeconds and others.
>
IMO, the main idea to having replicas of KSM is avoiding loss metrics during a node downtime (It is a common use-case).  We could recomend to use topology spreads constrains or a podAntiAffinity (by hostname) together to solve this problem.

> **Reschedules:**
> 
> Multiple replicas might also help with pod reschedules, but we can achieve stability using something like PDBs to avoid ending up with zero replicas.
> 
Yeah, but the problem is the duplicate of metrics basically. This PR dealing it offering an alternative to the use of serviceMonitor and propose the use of scrapeConfig to see the KSM as a simple kubernetes service.

> If we really want to increase the availability of KSM, maybe we can consider the sharding feature, so you can have multiple KSM instances, each one collecting metrics from specific APIs.
> 
The sharding feature is great, I use it in big cluster when I want to distribuite the load accross multiples KSM instances. But, if you lose a node running a specific shard, you will lose metrics and normally the schedule in statefulset is a headache if the node still in a Terminating state. Also, It already happens to Prometheus shards strategy too.

For small/medium clusters this PR solve a HA problem in a simple way. The KSM is an important service for alerts and SLO's metrics and we could have a minimal of availability if we look it as classic kubernetes service.

charts/kube-state-metrics/values.yaml
```diff
@@ -239,6 +239,46 @@ prometheus:
         # name: secret-name
         # key:  key-name
       tlsConfig: {}
+  ## Create a scrapeConfig resource for scraping the kube-state-metrics service. Use this instead of serviceMonitor
+  ## to have more instances of kube-state-metrics safety.
+  scrapeconfig:
+    ## To avoid duplicate metrics, first disable the serviveMonitor creation via prometheus.monitor.enabled=false
```
dotdc, line 245:
```suggestion
    ## To avoid duplicate metrics, first disable the serviceMonitor creation via prometheus.monitor.enabled=false
```

dotdc (approved):

charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,8 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 5.31.2
-# renovate: github-releases=kubernetes/kube-state-metrics
+version: 5.32.0
```
dotdc, line 10:
```suggestion
version: 5.32.0
# renovate: github-releases=kubernetes/kube-state-metrics
```

dotdc (approved):

dotdc:
Thank you for your contribution @jenciso !

PR #5473, opened 2025-03-24
[prometheus-systemd-exporter]  Allow set unit include and bump to v0.7.0

### LINKED ISSUE #4879: [prometheus-systemd-exporter] Customize DaemonSet args

The `args` in the DaemonSet template are hardcoded and there is no way to change them

https://github.com/prometheus-community/helm-charts/blob/24979b1d1aee425312e039882de3d27c8fa9b607/charts/prometheus-systemd-exporter/templates/daemonset.yaml#L52-L54

Those values are very restrictive, since some kubernetes engines use other service names, e.g. `k3s.service`, or users may want to monitor additional service units as well. It would be much better if users could override args completely.

Being able to override `args` using `values.yaml`

Adding my customization in `extraArgs`, but everything fails because `--systemd.collector.unit-include` can only be specified once.

### Additional context.

_No response_

kir4h:

Allows setting `systemd.collector.unit-include` argument, which was hardcoded.
This allows the user to select the systemd units he is interested into.

- fixes #4879 

~I have leveraged existing `extraArgs` for this. The exporter itself has a default value of `.+` for included units, but I have kept current behaviour not to do a breaking change.~

I have also:

~- Bumped exporter version to [v0.7.0](https://github.com/prometheus-community/systemd_exporter/releases/tag/v0.7.0) which has been recently released~ -> will do under separate PR
- Bumped the Chart version to ~`1.0.0`~ `0.4.0`
  ~- This is not a breaking change, so might be confusing. But according to https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md?plain=1#L51 charts should start at `1.0.0`~

I can undo any of those change, or do them in a separate PR if preferred.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-systemd-exporter/Chart.yaml
+++ b/charts/prometheus-systemd-exporter/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: prometheus-systemd-exporter
 description: A Helm chart for prometheus systemd-exporter
 type: application
-version: 1.0.0
-appVersion: "0.7.0"
+version: 0.4.0
+appVersion: "0.6.0"
 home: https://github.com/prometheus-community/systemd_exporter
 sources:
 - https://github.com/prometheus-community/systemd_exporter
--- a/charts/prometheus-systemd-exporter/Chart.yaml
+++ b/charts/prometheus-systemd-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus-systemd-exporter
 description: A Helm chart for prometheus systemd-exporter
 type: application
-version: 0.3.0
+version: 1.0.0
 appVersion: "0.6.0"
 home: https://github.com/prometheus-community/systemd_exporter
 sources:
--- a/charts/prometheus-systemd-exporter/Chart.yaml
+++ b/charts/prometheus-systemd-exporter/Chart.yaml
@@ -3,7 +3,7 @@ name: prometheus-systemd-exporter
 description: A Helm chart for prometheus systemd-exporter
 type: application
 version: 1.0.0
-appVersion: "0.6.0"
+appVersion: "0.7.0"
 home: https://github.com/prometheus-community/systemd_exporter
 sources:
 - https://github.com/prometheus-community/systemd_exporter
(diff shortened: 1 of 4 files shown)


charts/prometheus-systemd-exporter/values.yaml
```diff
@@ -51,7 +51,8 @@ env: {}
 ##    VARIABLE: value
 
 # extraArgs allows to pass command line arguments, as described on https://github.com/prometheus-community/systemd_exporter?tab=readme-ov-file#configuration
-extraArgs: []
+extraArgs:
+  - --systemd.collector.unit-include=kubelet.service|docker.service
```
maxime1907, line 55: if people have already set extraArgs it will be breaking since it will disable it, prefer to use a new field:
```yaml
config:
  log:
    level: info
  systemd:
    collector:
      unit-include:
        - kubelet.service
        - docker.service
```

charts/prometheus-systemd-exporter/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: prometheus-systemd-exporter
 description: A Helm chart for prometheus systemd-exporter
 type: application
-version: 0.3.0
-appVersion: "0.6.0"
+version: 1.0.0
```
maxime1907, line 5: this is a feature so 0.4.0

charts/prometheus-systemd-exporter/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
...
+appVersion: "0.7.0"
```
maxime1907, line 6: could you do a separate PR for this so that we can revert the change if its a bugged version

charts/prometheus-systemd-exporter/values.yaml
kir4h: My bad! That was my original approach but then thought `extraArgs` would be simpler to implement (didn't realize it was breaking for those already using it).

Will refactor it

maxime1907 (approved):

PR #5475, opened 2025-03-24
[kube-prometheus-stack] allow extraManifests as map/list

TheRealNoob:

Allow `extraManifests` to be of type map or list.  Useful for splitting the values across multiple files, such as one encrypted and another not.

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -1,5 +1,9 @@
 # this file tests some non default values to increase the test coverage
 
+commonLabels:
+  foo: one
+  bar: two
+
 defaultRules:
   additionalRuleLabels:
     key: value
@@ -93,3 +97,25 @@ nodeExporter:
 prometheus-node-exporter:
   kubeRBACProxy:
     enabled: true
+
+extraObjects:
+  - apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: test-cm
+    data:
+      plaintext: foobar
+      templated: '{{ print "foobar" | upper }}'
+  - |
+    apiVersion: v1
+    kind: Secret
+    type: Opaque
+    metadata:
+      name: test-secret
+      labels:
+        {{- range $key, $value := .Values.commonLabels }}
+        {{ $key: $value }}
+        {{- end }}
+    data:
+      plaintext: Zm9vYmFy
+      templated: '{{ print "foobar" | upper | b64enc }}'
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -114,7 +114,7 @@ extraObjects:
       name: test-secret
       labels:
         {{- range $key, $value := .Values.commonLabels }}
-        {{ $key: $value }}
+        {{ $key }}: {{ $value }}
         {{- end }}
     data:
       plaintext: Zm9vYmFy
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -1,9 +1,5 @@
 # this file tests some non default values to increase the test coverage
 
-commonLabels:
-  foo: one
-  bar: two
-
 defaultRules:
   additionalRuleLabels:
     key: value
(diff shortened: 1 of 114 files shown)


charts/kube-prometheus-stack/templates/extra-objects.yaml
```diff
@@ -1,4 +1,15 @@
...
+{{- end -}}
+
+{{- range $extraObjects }}
 ---
-{{ tpl (toYaml .) $ }}
-{{ end }}
+  {{- if kindIs "map" . }}
+    {{- tpl (toYaml .) $ | nindent 0 }}
+  {{- else if kindIs "string" . }}
+    {{- tpl . $ | nindent 0 }}
```
jkroepke, line 13: what is the use case for `nindent 0`

TheRealNoob: it just allows me to indent the if block, nothing else
```
  {{- if kindIs "map" . }}
    {{- tpl (toYaml .) $ | nindent 0 }}
  {{- else if kindIs "string" . }}
```

charts/kube-prometheus-stack/values.yaml
```diff
@@ -5300,8 +5300,10 @@ thanosRuler:
 ##
 cleanPrometheusOperatorObjectNames: false
 
-## Extra manifests to deploy as an array
-extraManifests: []
+## Extra manifests to deploy.  Can be of type dict or list.
+## If dict, keys are ignored and only values are used.
+## Items contained within extraObjects can be defined as dict or string and are passed through tpl.
+extraManifests: null
```
jkroepke, line 5306: Also please add a list entry below, which is an string and contains helm expressions, e.g. an if.

jkroepke (commented):
Could you please add a extraManifests example here (list) that contains object and string elements?

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml

and here a example with strings and objects, but extraManifests is an object:

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml

That should cover all possibilities through CI.

charts/kube-prometheus-stack/values.yaml
TheRealNoob: I think I covered what you asked for.  Happy to make further changes if you spot something.

GMartinez-Sisti:
> Will do. In the sake of testing I will share the helm-unittests that I wrote for one of our internal charts. All of these tests pass. Hopefully that gives you some confidence. The only difference is this chart doesn't call it `extraManifests` but `manifests`.
> 
> https://gist.github.com/TheRealNoob/ca7dee0e862a4e5526a115d54e863ec9

We have some tests on https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/unittests, if you have them already you can add them.

charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
```diff
@@ -93,3 +97,25 @@ nodeExporter:
...
+    type: Opaque
+    metadata:
+      name: test-secret
+      labels:
+        {{- range $key, $value := .Values.commonLabels }}
+        {{ $key }}: {{ $value }}
+        {{- end }}
+    data:
+      plaintext: Zm9vYmFy
+      templated: '{{ print "foobar" | upper | b64enc }}'
```
jkroepke, line 121: Consider to use a ConfigMap here, too. This should avoid trigger GitLeaks

charts/kube-prometheus-stack/ci/05-ingress-and-gateway-routes-values.yaml
```diff
@@ -88,3 +88,22 @@ thanosRuler:
...
+      templated: '{{ print "foobar" | upper }}'
+  secret: |
+    apiVersion: v1
+    kind: Secret
+    type: Opaque
+    metadata:
+      name: test-secret
+    data:
+      plaintext: Zm9vYmFy
+      templated: '{{ print "foobar" | upper | b64enc }}'
```
jkroepke, line 109: Consider to use a ConfigMap here, too. This should avoid trigger GitLeaks

jkroepke (commented):
LGTM. Optionally, add the unit tests, if you wish. @GMartinez-Sisti not sure, if they part of CI

GMartinez-Sisti:
> LGTM. But Gitleaks has some issues.
> 
> Optionally, add the unit tests, if you wish. @GMartinez-Sisti not sure, if they part of CI

They are disabled 😢 : https://github.com/prometheus-community/helm-charts/blob/main/.github/linters/ct.yaml#L4

Will send a PR to enable as I think there is great value on this.

TheRealNoob:
I removed the linting tests and added the unittest since it's more comprehensive.  This should also fix the gitleaks error.

@GMartinez-Sisti unfortunately not all of the tests pass right now, but I at least made sure that extraObjects does.  [Here](https://gist.github.com/TheRealNoob/da99af71788cb7915879e0a85d460ac7) is the output so you can have a look.  Also I changed the command slightly because it ran my test twice - find it at the top of the gist.  The reason I placed the test at `unittests/extra-objects_test.yaml` is because I figure we're supposed to follow the format of the `templates/` dir, which I did.  It just happens to be the singular exception to the structure.

TheRealNoob:
@jkroepke ready for merge?

jkroepke (approved):
LGTM

PR #5503, opened 2025-04-02
[prometheus-adapter] fix metrics api in README

kingdonb:
When the sum over average values is reported by Metrics API, the result for query consumers is a number of bytes with a decimal point and some precision after it. That's really not good for a number in bytes.

Headlamp UI somehow receives that number re-formatted as millibytes, which its parseRam function does not correctly balk and crash at, so the UI occasionally reports 1000x the actual memory usage, and sometimes flaps back to displaying correctly.

When I add these round wrappers the metric gets reported as a whole number of bytes and that doesn't happen anymore!

No need for a chart rev, I don't know where this query definition comes from and I can find no references to it in any prometheus-adapter docs other than this one README, so I'm fixing it right at the source. (If anyone knows where it comes from, maybe we can cajole them into fixing it in their upstream as well.)

Ref:

* kubernetes-sigs/headlamp#3067

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 4.14.0
+version: 4.14.1
 appVersion: v0.12.0
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/README.md
+++ b/charts/prometheus-adapter/README.md
@@ -136,15 +136,15 @@ rules:
       containerLabel: container
     memory:
       containerQuery: |
-        sum by (<<.GroupBy>>) (
+        round(sum by (<<.GroupBy>>) (
           avg_over_time(container_memory_working_set_bytes{container!="",<<.LabelMatchers>>}[3m])
-        )
+        ))
       nodeQuery: |
-        sum by (<<.GroupBy>>) (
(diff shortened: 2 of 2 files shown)


kingdonb:
alright, that version is newer than the one in the README, I bet it doesn't return bytes with decimal precision... testing.

I'll use that one instead of the one I just cobbled together, if it fixes my issue! Thanks very much.

kingdonb:
https://github.com/prometheus-community/helm-charts/commit/d666aec1166414ef527949b660e6e1e7a6287fcd

@SuperQ might have an opinion on this. I'm going to update my PR to make the README match that config-map.yml example, but I see it's not necessarily as straightforward as "one version is newer" - there might be some parts of the query in README here that are better.

It just should not return a number of bytes as a decimal, I think VPA also was having issues with that. Nobody should pay any attention to those decimal values but they do.

kingdonb:
I'm actually gonna revert mine to the version I had before... just realized my clusters don't report node_memory_working_set_bytes or node_cpu_seconds_total with the labels I'm using for matching, and I'm a dog on the internet with Prometheus and don't know what I'm doing

Glad to participate in the discussion but if there's someone more knowledgeable than me about this, I would happily defer.

kingdonb:
The linter seems to be saying I still need a version bump, even though this change only updates documentation. Maybe some happy soul who is updating prometheus-adapter can incorporate this PR into their next release. 🎉

steven-sheehy:
You can just bump the patch version. Otherwise this will sit here forever.

steven-sheehy (approved):

PR #5533, opened 2025-04-11
[prometheus-operator-crds] add opt-out for CRDs

### LINKED ISSUE #3860: [prometheus-operator-crd] Enable or disable some CRDs

Some other collectors, such as Grafana Agent or OpenTelemetry Collector commonly are able to use ServiceMonitor, PodMonitor and PrometheusRule CRs, but not the rest of them

It would be great if we could select which to deploy or not, like this:

```yaml
  crds:
    podMonitor:
      enable: false
    serviceMonitor:
      enable: false
    prometheusRule:
      enable: false
```

The other alternative is to create a meta-chart with some of the CRDs included. The problem of this solution is more or less the same, as if you have several environments, you probably want to disable PrometheusRule CR in other environments that is not the main one having the metrics DB. For example, when you use Mimir, in order to upload Prometheus rules without exposing the Ruler (to send rules over internet through the Mimir Ruler API, which is increasing the attack surface)

### Additional context.

_No response_

sberz:
This change adds an opt-out mechanism for CRDs so only the CRDs needed for the cluster are installed (e.g. only pod- and servicemonitors for setups using Mimir).

- fixes #3860

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-operator-crds/values.yaml
+++ b/charts/prometheus-operator-crds/values.yaml
@@ -1,25 +1,55 @@
-## Annotations for CRDs
+## Settings for CRDs
 ##
 crds:
+  ## annotations add additional annotations to all CRDs
   annotations: {}
 
+  ## alertmanagerconfigs configures the AlertManagerConfig CRD
   alertmanagerconfigs:
+    ## enabled defines if the CRD should be installed
     enabled: true
+
+  ## alertmanagers configures the AlertManager CRD
   alertmanagers:
+    ## enabled defines if the CRD should be installed
     enabled: true
+
+  ## podmonitors configures the PodMonitor CRD
   podmonitors:
+    ## enabled defines if the CRD should be installed
     enabled: true
+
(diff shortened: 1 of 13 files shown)


charts/prometheus-operator-crds/values.yaml
```diff
@@ -2,3 +2,24 @@
...
+  prometheuses:
+    enabled: true
+  prometheusrules:
+    enabled: true
+  scrapeconfigs:
+    enabled: true
+  servicemonitors:
+    enabled: true
+  thanosrulers:
+    enabled: true
```
QuentinBisson, line 25:
```suggestion
    enabled: true

```

sberz: I added comments for all properties following the Helm conventions in 4448089f936904c7dcaaf98d41328a42113744e0 @QuentinBisson

QuentinBisson (approved):
LGTM Thank you :)

PR #5558, opened 2025-04-24
[kube-prometheus-stack] Fix: Return comma-separated list from `dnsNames` helper to prevent malformed YAML

johnswarbrick-napier:
The `kube-prometheus-stack.operator.admission-webhook.dnsNames` helper currently emits a newline-separated list of hostnames. When this was consumed in template contexts like:

  - --host={{- include "..." . | replace "\n" "," }}

…the result could break YAML parsing due to unquoted, comma-separated values. For example:

`Error: YAML parse error on kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml: error converting YAML to JSON: yaml: line 46: did not find expected key`

This commit modifies the helper to construct a list using `list` and `print`, then returns a comma-separated string using `join`. This ensures downstream templates can safely consume the output without needing to sanitize or quote it.

Consumers can now use:

  - "--host={{ include "kube-prometheus-stack.operator.admission-webhook.dnsNames" . }}"

Which is considered safe and avoids parsing issues


charts/kube-prometheus-stack/templates/_helpers.tpl
```diff
@@ -319,15 +319,15 @@ global:
...
-{{- end }}
+{{- define "kube-prometheus-stack.operator.admission-webhook.dnsNames" -}}
+{{- $fullname := include "kube-prometheus-stack.operator.fullname" . -}}
+{{- $namespace := include "kube-prometheus-stack.namespace" . -}}
+{{- $names := list $fullname (print $fullname "." $namespace ".svc") -}}
+{{- if .Values.prometheusOperator.admissionWebhooks.deployment.enabled -}}
+{{- $names = append $names (print $fullname "-webhook") -}}
+{{- $names = append $names (print $fullname "-webhook." $namespace ".svc") -}}
+{{- end -}}
+{{- join "," $names -}}
```
jkroepke, line 330: We should not join here, since the it used on other manifests as well, but differently.

https://github.com/prometheus-community/helm-charts/blob/b0e74dc44bfba0d288098e0905589701c7322b23/charts/kube-prometheus-stack/templates/prometheus-operator/certmanager.yaml#L60

johnswarbrick-napier: Good spot. In that case I'll revert the changes to `_helpers.tpl` and instead just double quote in `job-createSecret.yaml`. That should encapsulate the comma-separated value as a single YAML string to avoid parsing errors.

johnswarbrick-napier: @jkroepke - please review when you have the chance

charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml
```diff
@@ -44,7 +44,7 @@ spec:
           imagePullPolicy: {{ .Values.prometheusOperator.admissionWebhooks.patch.image.pullPolicy }}
           args:
             - create
-            - --host={{- include "kube-prometheus-stack.operator.admission-webhook.dnsNames" . | replace "\n" "," }}
+            - "--host={{ include "kube-prometheus-stack.operator.admission-webhook.dnsNames" . | replace "\n" "," }}"
```
jkroepke, line 47: There is a reason for omit the hypen before the `include`?

PR #5581, opened 2025-04-29
fix(kube-prometheus-stack): faulty logic when choosing imageRegistry

dminca:
Choose which imageRegistry to use based on whichever is defined.

Currently if `.Values.global.imageRegistry` is set to quay.io for example, this image will fail to be pulled because it's located in a different registry.

Therefore, the `.Values.prometheusOperator.admissionWebhooks.patch.image.registry` must take precedence over the former since this image is located in a different registry.

- fixes #

#### Checklist

- [x] Chart Version bumped


jkroepke (changes requested):
The current logic works as indented. The `Values.global.imageRegistry` is primary designed for registry mirrors. If you set `Values.global.imageRegistry` to an mirror which does not contain the patch image, you have to manually override `.Values.prometheusOperator.admissionWebhooks.patch.image.registry`.

It would be a faulty logic if every image follow the same pattern except one.

dminca:
@jkroepke I agree, that's how it should work. However the situation is different.

Let's say you are mirroring `quay.io` since that's where majority of the images are so you set `.Values.global.imageRegistry` to it.

But in reality the image `ingress-nginx/kube-webhook-certgen` is pulled from `registry.k8s.io` so then you decide to overwrite `.Values.prometheusOperator.admissionWebhooks.patch.image.registry` and point to the k8s-io-mirror.

The problem's `.Values.prometheusOperator.admissionWebhooks.patch.image.registry` is not overwritten, instead `.Values.global.imageRegistry` is used.

Hope I managed to explain it properly so that you understand the issue here 😅

jkroepke:
Okay, I understand that `Values.global.imageRegistry` always wins and I agree this should not be the case.

But its wrong everywhere, right?

https://github.com/search?q=repo%3Aprometheus-community%2Fhelm-charts+Values.global.imageRegistry+path%3A%2F%5Echarts%5C%2Fkube-prometheus-stack%5C%2F%2F&type=code

dminca:
@jkroepke it's not wrong everywhere. Only in the places where the image is located in a different container registry from `quay.io`

jkroepke:
In general, I for an constant behavior. An specific registry definition should always prefer over global.imageRegistry and as I understand, this is not the case, nowhere. Right?

dminca:
hmm, I guess so. Shall I apply this everywhere?

jkroepke:
Yes please

W48B1T:
This is still an issue. I think this was forgotten about?

jkroepke:
Pick up the PR, if you want.

Poil:
I think it should be more logic to have 3 levels

* global.imageRegistry
* prometheus.global.imageRegistry to configure the registry of Prometheus community images
* componentLevel.imageRegistry

In my case we use ECR pull cache, so every registry have to be cache on a subpath
We have
* {{ .aws.accountId }}.dkr.ecr.{{ .aws.region }}.amazonaws.com/quay for quay.io
* {{ .aws.accountId }}.dkr.ecr.{{ .aws.region }}.amazonaws.com/dockerhub for docker.io
* (...)

On my side I'm going to configure  global.imageRegistry with my ECR and configure repository on each component to include the "registry prefix"

regards

jkroepke:
Feel free to create a PR

PR #5582, opened 2025-04-29
[alertmanager] Allow customizing the name of the container port

faevourite:
It can be useful to customize the name of Alertmanager's container port in situations where another K8s object is observing new deployments (e.g. Prometheus collector that watches for ports named "metrics"). This PR enables this use case with a new top-level `containerPortName` value, which defaults to the current, hard-coded "http" value.

I left the name of the Service as-is just to keep the PR small.

#### Checklist

- [x] Chart Version bumped


faevourite:
@monotek @naseemkullah Hi folks 👋🏻  Is there anything I need to do to move this forward? :)

monotek (dismissed):

PR #5584, opened 2025-04-30
[prometheus-smartctl-exporter] Allow PodMonitor and PriorityClass configuration

briend:

- Allows use of PodMonitor instead of the traditional ServiceMonitor.  This can be more efficient than creating a Service (unnecessary iptables rules overhead) and avoid ServiceMonitor Endpoint issues with >1000 Pods: 
  - https://github.com/prometheus-operator/prometheus-operator/pull/6672
  - https://github.com/prometheus-operator/prometheus-operator/issues/3119
- Allows setting a priorityClassName
- If ServiceMonitor is not enabled, the Service is not created either

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-smartctl-exporter/templates/service.yaml
+++ b/charts/prometheus-smartctl-exporter/templates/service.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.serviceMonitor.enabled }}
+{{- if .Values.service.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
--- a/charts/prometheus-smartctl-exporter/templates/service.yaml
+++ b/charts/prometheus-smartctl-exporter/templates/service.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.serviceMonitor.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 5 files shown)


charts/prometheus-smartctl-exporter/templates/service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.serviceMonitor.enabled }}
```
kfox1111, line 1: Maybe make this service.enabled... There could be folks using the service along with a manually installed service monitor or prometheus config.

The rest looks good though. Thanks! :)

briend: Thanks @kfox1111 , I agree.  I've added that and leave Service enabled by default just in case.

charts/prometheus-smartctl-exporter/values.yaml
```diff
@@ -31,6 +31,24 @@ serviceMonitor:
   relabelings: []
   metricRelabelings: []
 
+
+
```
kfox1111, line 35:
```suggestion
```

PR #5597, opened 2025-05-02
[prometheus-pushgateway] allow for global.image.registry and global.imagePullSecrets overrides

Stevenpc3:

allow for global.imageRegistry and global.imagePullSecrets overrides with the local chart settings taking precedence over the global settings.

This is needed to allow users that must host their own registry easily override where the chart pulls from.

#### Checklist

- [x] Chart Version bumped

@gianrubio @cstaud @zeritti

### DIFF
--- a/charts/prometheus-pushgateway/templates/_helpers.tpl
+++ b/charts/prometheus-pushgateway/templates/_helpers.tpl
@@ -143,6 +143,20 @@ password: {{ $password | b64enc | quote }}
 {{- end }}
 
 {{/*
+Set the image with or without the registry
+*/}}
+{{- define "prometheus-pushgateway.image" -}}
+{{- $registry := default .Values.image.registry (.Values.global).imageRegistry }}
+{{- $repository := .Values.image.repository }}
+{{- $tag := default .Chart.AppVersion .Values.image.tag }}
+{{- if $registry }}
+    {{- printf "%s/%s:%s" $registry $repository $tag -}}
+{{- else -}}
+    {{- printf "%s:%s"  $repository $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Returns pod spec
 */}}
 {{- define "prometheus-pushgateway.podSpec" -}}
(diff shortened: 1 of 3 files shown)


charts/prometheus-pushgateway/templates/_helpers.tpl
```diff
@@ -142,6 +142,20 @@ user: {{ $user | b64enc | quote }}
 password: {{ $password | b64enc | quote }}
 {{- end }}
 
+{{/*
+Set the image with or without the registry
+*/}}
+{{- define "prometheus-pushgateway.image" -}}
+{{- $registry := default (.Values.global).imageRegistry .Values.image.registry }}
```
zeritti, line 149: In most charts in this repository, it is `global.imageRegistry` that overrides `image.registry`. I think we should keep this.

```suggestion
{{- $registry := default .Values.image.registry (.Values.global).imageRegistry }}
```

charts/prometheus-pushgateway/templates/_helpers.tpl
```diff
@@ -155,7 +169,7 @@ priorityClassName: {{ . | quote }}
 hostAliases:
 {{- toYaml . | nindent 2 }}
 {{- end }}
-{{- with .Values.imagePullSecrets }}
+{{- with .Values.imagePullSecrets | default (.Values.global).imagePullSecrets }}
```
zeritti, line 172: As above.

```suggestion
{{- with (.Values.global).imagePullSecrets | default .Values.imagePullSecrets }}
```

zeritti (changes requested):
Thank you, @Stevenpc3, for your PR.

Please, enter the new fields in the values file making them known to the users:

```yaml
global:
  imageRegistry: ""
  imagePullSecrets: []

image:
  registry: ""
```

charts/prometheus-pushgateway/templates/_helpers.tpl
Stevenpc3: ok that is fine.  We don't do that for ours, but I will keep the way it's done for these charts

Stevenpc3:
updated

zeritti (approved):
Thanks, @Stevenpc3, LGTM.

PR #5601, opened 2025-05-02
[kube-prometheus-stack] Use named port from coredns for metrics

jcpunk:

Use the named `metrics` port on the coredns deployment for associating to metrics.

- fixes #

This was added back in the 1.14 timeline when the port was introduced: https://github.com/kubernetes/kubernetes/commit/e9220dd408ec10b787e214eb7f8b9a765f5a8614

#### Checklist

- [x] Chart Version bumped


jkroepke:
I'm guess using the real port here is a better approach. 

There is no well defined standard approach the deploy CoreDNS server. Not sure, if every distribution gives the port a name.

jcpunk:
Looks like RKE2 labels the metrics port as `tcp-9153` rather than `metrics`.  I'll close this out since at least one k8s distribution has a different name here.

PR #5608, opened 2025-05-05
[prometheus-operator-admission-webhook] reorder job-patchwebhook

### LINKED ISSUE #5607: [prometheus-operator-admission-webhook] job-patchWebhook runs too late

When using the default `jobs.enabled=true` to create the TLS key-pairs, the [job-patchWebhook](https://github.com/prometheus-community/helm-charts/blob/4d3448baee00e2a5336a84ee23e92a5ea8fef9ca/charts/prometheus-operator-admission-webhook/templates/hooks/job-patchWebhook.yaml) runs too late in the release life-cycle.  If you deploy any `prometheusRules` as part of the release, they will fail with the error `unknown Certificate Authority` causing the helm release to error out.  At present the patch job is applied with ["helm.sh/hook": post-install,post-upgrade](https://github.com/prometheus-community/helm-charts/blob/4d3448baee00e2a5336a84ee23e92a5ea8fef9ca/charts/prometheus-operator-admission-webhook/templates/hooks/job-patchWebhook.yaml#L13) which means **after** all other resources are applied.

What are the scenarios where a prometheusRule could be deployed with the release?

* This chart has native support for it in `prometheusRule.enabled`, but it is disabled by default
* If this chart is used as part of another chart (IE subchart), and one of the other charts deploys prometheusRule objects

Proposed solution

I think it makes the most sense to run `job-patchWebhook` using `"helm.sh/hook": pre-install`.  This is what the `job-createSecret` uses.  The only thing we need to ensure is that the patch runs **after** the create and we can do that with `helm.sh/hook-weight: <int>`.

### What's your helm version?

version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.3"}

mike@EliteBook-840:~$ k version Client Version: v1.30.11 Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3 Server Version: v1.31.7-eks-bcf3d70

### Which chart?

prometheus-operator-admission-webhook

### What's the chart version?

0.24.1

### What happened?

helm release fails

`caBundle` needs to be defined before `prometheusRule` objects are applied

### How to reproduce it?

NA

prometheusRule.enabled=true

helm install prometheus-operator-admission-webhook prometheus-community/prometheus-operator-admission-webhook --set prometheusRule.enabled=true

Nope!

TheRealNoob:

Modifies release order of Helm hook jobs.  See below issue for deeper description.

- fixes #5607

#### Checklist

- [x] Chart Version bumped


TheRealNoob:
tomorrow i will test this out and figure out the error

charts/prometheus-operator-admission-webhook/Chart.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.24.2
+version: 0.24.3
```
zeritti, line 5: Should really be a minor change.
```suggestion
version: 0.25.0
```

zeritti (changes requested):
Thanks, @TheRealNoob, for the change.

I would still prefer making the webhook configurations in to hooks optional only, i.e. the default installation would not install them as hooks. A field like `webhooks.configurationsAsHooks` (bool), for example, would then be used as a toggle. Enabling the toggle (and jobs) would set the appropriate annotations in both the configurations and the jobs.

TheRealNoob:
@zeritti Appologies for leave this stale for so long.  Other stuff came up at work.  I haven't forgotten about this though.

After much thought I came to the conclusion that helm-hooks are the only solution I see for this.  At least without a complete refactor.  And I have no problem going with what you said, making this opt-in.  To that end, would you prefer a `webhooks.configurationsAsHooks` type option as you proposed, or leave it up to the user to add all the annotations themselves?  We could leave them in as comments.  I will admit though it's a complicated problem to follow without really giving it your full focus, especially if you're not extremely comfortable using helm.  If going the manual annotations route we will need to add support for customizing the annotations on each job independently.  At present the values file allows an annotation to be injected into both via one field.

zeritti:
> @zeritti Appologies for leave this stale for so long. Other stuff came up at work. I haven't forgotten about this though.
> 
> After much thought I came to the conclusion that helm-hooks are the only solution I see for this. At least without a complete refactor. And I have no problem going with what you said, making this opt-in. To that end, would you prefer a `webhooks.configurationsAsHooks` type option as you proposed, or leave it up to the user to add all the annotations themselves? We could leave them in as comments. I will admit though it's a complicated problem to follow without really giving it your full focus, especially if you're not extremely comfortable using helm. If going the manual annotations route we will need to add support for customizing the annotations on each job independently. At present the values file allows an annotation to be injected into both via one field.

Letting users set annotations on the jobs and giving them full control in this respect is fine with me. Comments would be helpful, indeed.

charts/prometheus-operator-admission-webhook/templates/hooks/job-createSecret.yaml
```diff
@@ -10,15 +10,12 @@ metadata:
   name:  {{ template "prometheus-operator-admission-webhook.fullname" . }}-create
   namespace: {{ template "prometheus-operator-admission-webhook.namespace" . }}
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
-    {{- with .Values.jobs.annotations }}
-    {{- toYaml . | nindent 4 }}
-    {{- end }}
+    {{- $defaultAnnotations := dict "helm.sh/hook" "pre-install,pre-upgrade" "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded" }}
+    {{- mustMerge .Values.jobs.createSecret.annotations .Values.jobs.annotations $defaultAnnotations | toYaml | nindent 4 }}
```
TheRealNoob, line 14: @zeritti what do you think of this approach?  Before I finish fleshing it out

zeritti: That looks good (apart from the colon in the second key). I would suggest using named (helper) templates where reasonable to avoid repeating the same code in the files.

TheRealNoob: I considered that at first, infact I even started implementing it, but the default `helm.sh/hook` annotation is different for each of them.  So it kind of defeats the purpose pretty quickly.

TheRealNoob:
signing commits can be so painful sometimes :(.  one of these days i'll figure out how to sign from vscode.

charts/prometheus-operator-admission-webhook/templates/hooks/job-createSecret.yaml
TheRealNoob: Okay fixed everything.  Just haven't added descriptions to the comments I added.  Really struggling to describe it in any way that's even remotely brief.

charts/prometheus-operator-admission-webhook/templates/hooks/job-patchWebhook.yaml
```diff
@@ -10,15 +10,12 @@ metadata:
   name:  {{ template "prometheus-operator-admission-webhook.fullname" . }}-patch
   namespace: {{ template "prometheus-operator-admission-webhook.namespace" . }}
   annotations:
-    "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
-    {{- with .Values.jobs.annotations }}
-    {{- toYaml . | nindent 4 }}
-    {{- end }}
+    {{- $defaultAnnotations := dict "helm.sh/hook" "post-install,post-upgrade" "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded" }}
```
zeritti:
```suggestion
    {{- $defaultAnnotations := dict "helm.sh/hook" "post-install,post-upgrade" "helm.sh/hook-delete-policy" "before-hook-creation,hook-succeeded" }}
```

zeritti (changes requested):
Thanks, @TheRealNoob. Setting the hook weight on the configurations via `webhook.annotations` then should support the required ordering with respect to the jobs. 

What about the resources that the jobs require, e.g. service account, role, rolebinding etc.? Do they get created in time before the create secret job starts once its hook-weight has been adjusted? I am not sure about how a hook resource with an unspecified hook-weight relates to that with a specified one.

charts/prometheus-operator-admission-webhook/templates/hooks/job-patchWebhook.yaml
TheRealNoob: hmmmm very good thought.  I started working on testing this shortly after my last comment and separately implemented your above change.  My testing didn't reveal any issues; all looks good.  I installed and then upgraded the chart with the same values files and it worked nominally both times.

I also used audit logs to list out all of the kube-apiserver calls that helm made just so I knew with certainty it was working as expected.  I actually learned just a little about helm; it makes a lot more API calls than you would think.  I truncated out the info that wasn't helpful; top of the file is the beginning of the deployment.  The output is limited to just the helm install of the operator-admission-webhook, it does not include the install of the operator-crds.
https://gist.github.com/TheRealNoob/4f428e98ccad4ce5386873ed4337decf

And you can find Helm's install order [here](https://github.com/helm/helm/blob/a73c51ca08297fda17f40b3b11ff602e22893334/pkg/release/util/kind_sorter.go#L31).

zeritti (approved):

PR #5636, opened 2025-05-13
[alertmanager] Variable to add extra k8s Objects

rakesh-von:
This PR lets users add additional k8s objects to fulfill their custom needs like adding additional service or targetgroupbindings.

- fixes #

#### Checklist

- [x] Chart Version bumped


monotek:
I'm not a fan of such additions.
Users might come up with weird error reports which are not part of the chart but some totally unrelated problems.
Why not use gitops to add stuff which is not part of the chart?

rakesh-von:
> I'm not a fan of such additions. Users might come up with weird error reports which are not part of the chart but some totally unrelated problems. Why not use gitops to add stuff which is not part of the chart?

HI @monotek 

We are currently using gitops to add additional stuff but our aim is to not have multiple places from where a deployment can be changed or affected, rather have everything have from 1 source for the ease of maintenance and troubleshooting. Can you please consider this change? i am sure it will be helpful for others as well and people coming with weird error reports would anyhow come as they might not have enough helm experience?

monotek:
I mean error reports about stuff which is added in the additional manifests.
They would not report such things now.

rakesh-von:
> I mean error reports about stuff which is added in the additional manifests. They would not report such things now.

ah ok, now i understand what you meant. but this feature is disabled by default and people enabling this would know the consequences of it i guess and i feel this should be fine as it is disabled by default.

rakesh-von:
moreover this feature is available in other public helm repos as well and is widely used. for example https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/extra-objects.yaml

monotek:
@naseemkullah thoughts?

rakesh-von:
@naseemkullah and @monotek any updates please?

rakesh-von:
HI @naseemkullah did you get some time to review the PR please?

monotek:
This gets annoying. Closing and locking conversation...

Issue #5671, opened 2025-05-22
[All] Drop suuport for non current Kuberentes versions


DrFaust92:
### Is your feature request related to a problem ?

Many charts have a predicate for support very old/non current K8s versions (for example https://github.com/prometheus-community/helm-charts/blob/bbae3ecb238b94f17eec0d3be8b07c1147b121d0/charts/alertmanager/Chart.yaml#L12)

This introduces a burden of maintaining support for old API versions and actually not being tested as CI runs on a current K8s version.

### Describe the solution you'd like.

I propose bumping the K8s version to something current (latest - 3? it aligns with most major cloud providers) and removing the support for old APIs and simplifying templates

### Describe alternatives you've considered.

leave as is, making it dead code

DrFaust92:
Not stale

PR #5684, opened 2025-05-27
[prometheus] drop support for deprecated ingress version

DrFaust92:

- related #5671

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/templates/ingress.yaml
+++ b/charts/prometheus/templates/ingress.yaml
@@ -20,7 +20,9 @@ metadata:
   name: {{ template "prometheus.server.fullname" . }}
   namespace: {{ include "prometheus.namespace" . }}
 spec:
+  {{- if .Values.server.ingress.ingressClassName }}
   ingressClassName: {{ .Values.server.ingress.ingressClassName }}
+  {{- end }}
   rules:
   {{- range .Values.server.ingress.hosts }}
     {{- $url := splitList "/" . }}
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

zeritti (changes requested):
Thank you, @DrFaust92, for your PR, Please, see my comment below.

charts/prometheus/templates/ingress.yaml
```diff
@@ -23,9 +20,7 @@ metadata:
   name: {{ template "prometheus.server.fullname" . }}
   namespace: {{ include "prometheus.namespace" . }}
 spec:
-  {{- if and $ingressSupportsIngressClassName .Values.server.ingress.ingressClassName }}
   ingressClassName: {{ .Values.server.ingress.ingressClassName }}
```
zeritti, line 23: As we are not setting a default value, I'd suggest keeping the conditional:
```suggestion
  {{- if .Values.server.ingress.ingressClassName }}
  ingressClassName: {{ .Values.server.ingress.ingressClassName }}
  {{- end }}
```

I would also uncomment the field `.server.ingress.ingressClassName` in `values.yaml` and set it to an empty string. We should also remove the comments mentioning k8s version 1.18 in `server.ingress`.

zeritti (changes requested):

charts/prometheus/templates/ingress.yaml
DrFaust92: done

zeritti (approved):
Thanks, @DrFaust92, LGTM

PR #5693, opened 2025-05-29
[prometheus-blackbox-exporter] Added supporting of setting any namespace monitoring name in netwrokpolicy object

sanya2022:
Adds support for setting a custom namespace name in the NetworkPolicy. Currently, the name is hardcoded as "monitoring", which doesn't always match real-world usage. This PR does not change the default behavior, but simply adds the ability to specify an arbitrary name.

#### Checklist
- [x] Chart Version bumped


charts/prometheus-blackbox-exporter/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus Blackbox Exporter
 name: prometheus-blackbox-exporter
-version: 9.7.0
+version: 9.7.1
```
monotek:
```suggestion
version: 10.1.0
```

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -380,9 +380,11 @@ networkPolicy:
   # Enable network policy and allow access from anywhere
   enabled: false
   # Limit access only from monitoring namespace
-  # Before setting this value to true, you must add the name=monitoring label to the monitoring namespace
+  # Before setting this value to true, you must add the name=monitoring label to the monitoring namespace.  Name can be rewritten by monitoringNamespaceName
   # Network Policy uses label filtering
   allowMonitoringNamespace: false
+  # Rewrite monitoring namespace in network policy (default value monitoring)
+  monitoringNamespaceName: ""
```
monotek:
```suggestion
  monitoringNamespaceName: "monitoring"
```

charts/prometheus-blackbox-exporter/templates/networkpolicy.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   - from:
     - namespaceSelector:
         matchLabels:
-          name: monitoring
+          name: {{ default "monitoring" .Values.networkPolicy.monitoringNamespaceName }}
```
monotek: Any reason not to use: `Release.Namespace`?

```suggestion
          name: {{ .Values.networkPolicy.monitoringNamespaceName }}
```

sanya2022: `Release.Namespace` is not quite suitable in this situation, as it reflects the namespace where the exporter is deployed, rather than the namespace used by the monitoring systems.

PR #5722, opened 2025-06-02
[prometheus-blackbox-exporter] adjust extraEnv handling

Jakob3xD:
Change the `extraEnv` handling to cover all use cases.
The current `exntraEnv` and `extraEnvFromSecret` are limited in their use, as `extraEnv` always uses `value` and it is therefore not possible to use `valueFrom` to reference fields from the pod.
The `extraEnvFromSecret` is limited to Secrets but configMaps are also a valid source, so replacing it with a generic `extraEnvFrom` where both can be referenced.

As this is a braking change, the chart version gets a major bump and an entry to the README gets added.

#### Checklist

- [x] Chart Version bumped


charts/prometheus-blackbox-exporter/README.md
```diff
@@ -48,6 +48,11 @@ helm upgrade [RELEASE_NAME] [CHART] --install
 
 _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documentation._
 
+### To 10.0.0
+
+- `extraEnvFromSecret` got replaces with `extraEnvFrom`
```
monotek:
```suggestion
- `extraEnvFromSecret` got replaced with `extraEnvFrom`
```

charts/prometheus-blackbox-exporter/README.md
```diff
@@ -48,6 +48,11 @@ helm upgrade [RELEASE_NAME] [CHART] --install
...
+- `extraEnv` handling got changed to use valid k8s env definitions
```
monotek:
```suggestion
- `extraEnv` handling got changed to use k8s env definitions as list of maps
```

monotek (changes requested):

PR #5771, opened 2025-06-13
[prometheus] Fix small NodePort bugs on services

EoinFarrell:

The nodeport field on a service is only valid on Services of type NodePort.
Other charts in this repo handle this correctly, so adding the same if statement in kube-state-metrics chart.

Additionally, on the headless-svc, as type is not defined, it will default to a ClusterIP, thus the nodeport field is not valid.

Thanks in advance :)

#### Checklist

- [x] Chart Version bumped


charts/prometheus/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus
 # renovate: github=prometheus/prometheus
 appVersion: v3.4.1
-version: 27.20.0
+version: 27.21.0
```
zeritti: I'd suggest to make it a patch.
```suggestion
version: 27.20.1
```

zeritti (changes requested):
Thank you, @EoinFarrell, for your PR.

charts/prometheus/Chart.yaml
EoinFarrell: @zeritti  updated to patch 👍

zeritti:
@EoinFarrell Please, sign your commit for DCO to pass.

EoinFarrell:
My bad @zeritti sorry, commit is fixed now.

PR #5795, opened 2025-06-18
[kube-prometheus-stack] add ability to run insecure admission-webhook server

ushacow:

Currently admission-webhook can be started only in secure mode. When tls is disabled, it fails to start, since the default mode inside the server is tls enabled, so it's needed to specify `--web.enable-tls=false` and `--web.listen-address=:8080` to follow the insecure configuration in the template.

#### Checklist

- [ ] Chart Version bumped


jkroepke:
Could you explain the use-case?

ushacow:
> Could you explain the use-case?

I'm using Istio for mTLS, and it makes sense to me to disable tls in all the components of kube-prometheus-stack

jkroepke:
could you add the logic here as well?

https://github.com/prometheus-community/helm-charts/blob/05d9a64213e0c8650d19d2db41481ae4ad3f2790/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L140

ushacow:
> could you add the logic here as well?
> 
> https://github.com/prometheus-community/helm-charts/blob/05d9a64213e0c8650d19d2db41481ae4ad3f2790/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L140

Thanks @jkroepke for you help!

Actually I don't have any issue in the prometheus-operator deployment. It starts with `tls: enabled: false` perfectly well and listens on 8080 port. So I'm not sure if it makes sense to add explicit arguments there. Wdyt?

jkroepke:
Some have both deployments constient. Not more.

PR #5797, opened 2025-06-18
[kube-prometheus-stack] Allow null value in alertmanagerConfigNamespaceSelector

### LINKED ISSUE #4704: [kube-prometheus-stack] Default values prevent to override operator configuration

At differente places we can see these config

https://github.com/prometheus-community/helm-charts/blob/b320323b9852a169145b3aa64e98af7eae3a0178/charts/kube-prometheus-stack/values.yaml#L706C1-L708C44

```
    ## Namespaces to be selected for AlertmanagerConfig discovery. If nil, only check own namespace.
    ##
    alertmanagerConfigNamespaceSelector: {}
```

Thanks to this comment https://github.com/prometheus-operator/prometheus-operator/issues/5327#issuecomment-1424036851

> I see, then if I'm not mistaken you also have to set alertmanagerConfigNamespaceSelector to {} to select all the namespaces, if it's nil (which is the default if you don't configure it) the operator will look for AlertmanagerConfig in the other namespaces.

The helm chart forces the operator to look across all namespaces. One could ask if the helm chart should follow the default values of the operator or if having opinionated values is better, but that's another question.

The main issue here is that if you want to override the value by nil/null, helm will raise an error because types can't be merged

```
coalesce.go:286: warning: cannot overwrite table with non table for prometheus-stack.kube-prometheus-stack.alertmanager.alertmanagerSpec.alertmanagerConfigNamespaceSelector (map[])
```

### What's your helm version?

version.BuildInfo{Version:"v3.15.2", GitCommit:"1a500d5625419a524fdae4b33de351cc4f58ec35", GitTreeState:"clean", GoVersion:"go1.22.4"}

Client Version: v1.29.6 Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3 Server Version: v1.30.1

### Which chart?

kube-prometheus-stack

### What's the chart version?

61.2.0

### What happened?

nil value can't be merged with a map

I need a solution which allow me to set null or map like the operator allows

### How to reproduce it?

run helm template of the chart with this override

```
  alertmanager:
    alertmanagerSpec:
      alertmanagerConfigNamespaceSelector:
```      

_No response_

helm template -f prometheus-stack/values.yaml charts/prometheus-stack | more

I was working on the alertmanagerConfigNamespaceSelector value but there is different other places where this bug happens.

I think the best way to fix it would be to have a default value to nil. It can be override by a map and it also align with the default values of the operator

The bug is maybe only related to helm https://github.com/helm/helm/pull/12879

yaringol:
This PR fixes an issue where the `alertmanagerConfigNamespaceSelector` field was intended to be `null` when unspecified, but the chart rendered it as an empty map instead. This caused `Alertmanager` to select resources from all namespaces, rather than restricting discovery to its own namespace.

- fixes #4704 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 75.3.0
+version: 75.3.1
 
 # Please do not add a renovate hint here, since appVersion updates involves manual tasks
 appVersion: v0.83.0
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 75.3.1
+version: 75.3.2
 
 # Please do not add a renovate hint here, since appVersion updates involves manual tasks
 appVersion: v0.83.0
--- a/charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
+++ b/charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
@@ -75,12 +75,8 @@ spec:
 {{ else }}
   alertmanagerConfigSelector: {}
 {{- end }}
-{{- if .Values.alertmanager.alertmanagerSpec.alertmanagerConfigNamespaceSelector }}
   alertmanagerConfigNamespaceSelector:
-{{ tpl (toYaml .Values.alertmanager.alertmanagerSpec.alertmanagerConfigNamespaceSelector | indent 4) . }}
-{{ else }}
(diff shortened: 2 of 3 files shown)


jkroepke:
Feel like, if you set an empty in values, null will be rendered. In any case, this PR must include a unit test to ensure both is working

yaringol:
I've added unit tests to cover both cases — when the value is explicitly set to null, and when it's an empty map ({}). These tests ensure the chart renders the correct output in each scenario and avoids the original issue. 
All tests have passed successfully.

Let me know if there's anything else you'd like me to update.

jkroepke:
Thanks, that helps me a lot.

jkroepke (dismissed):
LGTM

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 75.3.0
+version: 75.3.1
```
jkroepke:
```suggestion
version: 75.3.2
```

jkroepke (approved):

Issue #5852, opened 2025-06-30
[kube-prometheus-stack] Migrate away from the python bindings for C/C++ jsonnet


mallardduck:
### Is your feature request related to a problem ?

Our team uses `kube-prometheus-stack` and I've recently started looking into how the tooling here is setup to generate dashboards/rules/etc. My investigation was done out of curiosity rather than with a specific purpose.

In my research I quickly hit issues running the Python code for the jsonnet related functions. This seems to be an incompatibility with mac and the pyhon bindings; Jsonnet for Python wraps the C API which wraps the C++ implementation - so its a few layers of abstraction. I'm fuzzy on the root issue with macs (which may be specific to arm macs?). However, in my investigation of that I learned that Python uses an older implementation of jsonnet and fixing the python dependency issues could be hard because of that.

Long story short, I learned that the Jsonnet library [desires to migrate users to the Go implementation](https://jsonnet.org/learning/getting_started.html#cpp-or-go). However the python version is still using C/C++ APIs until concerns about go availability is resolved (I can't find my source to discussions on this but I did see it).

The scripts used for this project already [depend on golang](https://github.com/prometheus-community/helm-charts/blob/bb2b3a7ad26d97ada0c8aa9916977887565dc8f8/charts/kube-prometheus-stack/hack/renovate-post-upgrade-hook.sh#L8). This provides an opportunity to migrate to the Go version directly.

### Describe the solution you'd like.

In a CLI tool my team uses I implemented the dashboard/rules/etc building scripts in golang. I would like to pull this out and submit it as a PR here which would replace the usage of Python in `charts/kube-prometheus-stack/hack`.

The end result would be that this new tooling could be run directly via `go run` and that the helper shell scripts would be updated to invoke the new go based tools. In turn, this would unblock myself, and other hopeful contributors, experiencing issues with python's jsonnet tooling.

### Describe alternatives you've considered.

1. Wait for python package to be migrated - this could take a while.
2. Prepare my contributions from a supported workstation - don't want to switch around HW.
3. Execute the scripts in a container - oddly feels like more effort than it's worth compared to migrating to a solution that can (in-theory) run native on any platform.

PR #5853, opened 2025-07-01
[kube-prometheus-stack][tooling] Migrate the jsonnet tooling to use native golang scripts

mallardduck:
Replacing the Python based (which is really C<->C++) jsonnet tooling with Golang native tooling.

- Addresses: https://github.com/prometheus-community/helm-charts/issues/5852

#### Checklist

- [n/a] Chart Version bumped - Not sure if needs to be bumped as no effective changes were made?


jkroepke:
I have mixed feeling here. I agreed that the python code was spaghetti code, but not sure how many kube-prometheus-stack maintainers can maintain go code. 

If agreed, I will do a deeper review here.

GMartinez-Sisti:
> I have mixed feeling here. I agreed that the python code was spaghetti code, but not sure how many kube-prometheus-stack maintainers can maintain go code.
> 
> If agreed, I will do a deeper review here.

Same feeling, but in reality the entire ecosystem is in go (k8s, helm, prometheus, alertmanager, etc) so I think it makes sense.

In the end it's all about opinion. For the record I prefer golang, and I use it pretty much every day. I can also review this later.

QuentinBisson:
I also use go on a daily basis but I would tend to not change things that work but if more people want to do that, it's fine for me and we can review and maintain this

The main issue with this PR is that's it's a big bang and it's doing too much. 85 changed files and more than 3000 added lines is too much. 

If @jkroepke and @GMartinez-Sisti are fine with this (wait for their response) then I would prefer that you split this PR into multiple ones with at least:
- 1 for skaffolding
- 1 for rules and switching to use the new go one
- 1 for dashboards  and switching to use the new go one
- 1 to update the resources

GMartinez-Sisti:
> then I would prefer that you split this PR into multiple ones

This is always a good approach, so I support it.

Also I would only expect the checksum on the header of the generated files to change, the files have changes on the content and github doesn't make it easy to understand what actually changed.

mallardduck:
> the files have changes on the content and github doesn't make it easy to understand what actually changed.

The only changes to files are superficial ones related to how yaml is marshalled. Of note:
- k8s-coredns.yaml path is corrected - the python relative path was incorrect for source VS dest paths. (So this fixed a bug)
- All dashboards have a superficial change of field order change (golang vs python order of `allValue` field)
- Some `description` fields use different yaml string format; I think this is because python was splitting lines at 1000 characters.

So it will be very hard to get an exact 1:1 result out of golang and python due to subtle differences in yaml library implementations.

---

> > then I would prefer that you split this PR into multiple ones
> 
> This is always a good approach, so I support it.


TBH I wasn't sure how much interest there would be in this, so didn't know how to prepare it. I'm happy to split up this effort into multiple PRs if that's the preferred way to digest it.

As mentioned this all came from an idea to add this to internal tooling my team uses, so pulling it out of the original source where I created it is why it's quite a big bang PR. As I have time this week, I will split up this PR into the suggested PRs and close this one.

jkroepke:
Hi, could you please skip the changes on the template for now? I'm unable to perform a review. Somehow the Github UI is terrible slow on the file diffs.

charts/kube-prometheus-stack/hack/internal/config/repos.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func UpdateRepoRefsConfig(configRoot string) error {
+	// First populate the head SHA
+	for key, repoConfig := range Repos {
+		if repoConfig.HeadSha != "" {
+			continue
+		}
+		headSha := util.Must(git.FindRepoHeadSha(repoConfig.RepoURL))
+		repoConfig.SetHeadFromHash(headSha)
```
jkroepke, line 38: By default, the SHA Commit should be not updated. Meanwhile, renovate take care of updates.

charts/kube-prometheus-stack/hack/cmd/mixins/main.go
```diff
@@ -0,0 +1,74 @@
...
+	"time"
+
+	"main/internal/config"
+	"main/internal/git"
+	"main/internal/util"
+)
+
+var (
+	scriptsRoot string
+	hackRoot    string
```
jkroepke, line 20: seems not in use

charts/kube-prometheus-stack/hack/internal/cmd/dashboards/patch/consts.go
```diff
@@ -0,0 +1,63 @@
+package patch
+
+import (
+	"bytes"
+	"encoding/json"
+	"main/internal/types"
+	"reflect"
+)
+
+var ReplacementMap = []types.DashboardReplacementRule{
```
jkroepke, line 10: what about using strings.Replacer instead loop with strings.Replace?

charts/kube-prometheus-stack/hack/internal/log/log.go
```diff
@@ -0,0 +1,26 @@
+package log
+
+import (
+	"github.com/sirupsen/logrus"
```
jkroepke, line 4: would it be possible to use log/slog instead dedicated logging lib?

jkroepke (changes requested):

mallardduck:
Appreciate the reviews and feedback - just wanted to clarify my game plan here. I do want to keep this moving too; but actual work projects came up & I over looked the holiday, so both those delayed any progress on this so far.

🤞 I'm hoping this week will be better for finding time to prioritize this. I had already started splitting up the changes based on this feedback. So I wasn't planning on directly modifying or submitting this PR further.

> * 1 for skaffolding
> * 1 for rules and switching to use the new go one
> * 1 for dashboards  and switching to use the new go one
> * 1 to update the resources

---

Sorry all, I haven't been able to circle back to allocate time to this community work. Hopefully in the coming weeks 🤞 .

charts/kube-prometheus-stack/hack/internal/config/repos.go
mallardduck: I can exclude this behaviour if you prefer - however the [existing scripts](https://github.com/prometheus-community/helm-charts/blob/5c25ea7c49c20bc7ade1120e0bfbb6910e91a94b/charts/kube-prometheus-stack/hack/update_mixins.sh#L31-L32) that this code replaces actually does this.

PR #5855, opened 2025-07-02
fix: improve helper text for metricLabelsAllowlist

NotAwar:

My team and I looked at the values file trying to make sense of whether or not the example was correct or not. We found the explanation might need additional clarity so I thought adding a clear example of expected values might help. Though I do still see room for improvement. 

- fixes #

#### Checklist

- [ ] Chart Version bumped


dotdc:
Hi @NotAwar,
Looks like the format is resource=[label-key1,label-key2], it maps resources to label keys, not label keys to values.

PR #5961, opened 2025-07-21
[prometheus-windows-exporter] Allow setting resources for initContainer configure-firewall

vladimirpetr:

This PR allows setting resources for the init container `configure-firewall`. Due to active policies in some environments, defining resources for init containers is required.

None

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-windows-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-windows-exporter/templates/daemonset.yaml
@@ -39,6 +39,9 @@ spec:
           image: {{ include "prometheus-windows-exporter.image" . }}
           command: [ "powershell" ]
           args: [ "New-NetFirewallRule", "-DisplayName", "'windows-exporter'", "-Direction", "inbound", "-Profile", "Any", "-Action", "Allow", "-LocalPort", "{{ .Values.service.port }}", "-Protocol", "TCP" ]
+          {{- with .Values.configureFirewall.resources }}
+          resources: {{ toYaml . | nindent  12 }}
+          {{- end }}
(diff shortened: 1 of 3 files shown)


charts/prometheus-windows-exporter/templates/daemonset.yaml
```diff
@@ -39,6 +39,9 @@ spec:
           image: {{ include "prometheus-windows-exporter.image" . }}
           command: [ "powershell" ]
           args: [ "New-NetFirewallRule", "-DisplayName", "'windows-exporter'", "-Direction", "inbound", "-Profile", "Any", "-Action", "Allow", "-LocalPort", "{{ .Values.service.port }}", "-Protocol", "TCP" ]
+          {{- with .Values.configureFirewall.resources }}
+          resources: {{ toYaml . | nindent  12 }}
```
jkroepke:
```suggestion
          resources: {{- toYaml . | nindent 12 }}
```

jkroepke:
```
          {{- with .Values.resources }}
          resources:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

Can we not re-use the same value? Any downsides?

vladimirpetr: Yes, resources can be common, indeed

jkroepke: Can you adjust the PR? That would be great.

jkroepke: Nvm.

PR #5971, opened 2025-07-24
[prometheus-nats-exporter] Support customization of timeout, period for NATS probes

romil93:
In our deployment of this chart, we frequently hit the timeout on the liveness probe because the metrics endpoint gets fairly large. We would want to be able to customize the timeout and frequency of the check.

@okgolove @caarlos0 

#### Checklist

- [x] Chart Version bumped


okgolove:
Hi @romil93. Thank you for the improvements. I think probes values should be under root key, not `service`. Could you adjust that? Thanks!

romil93:
> Hi @romil93. Thank you for the improvements. I think probes values should be under root key, not `service`. Could you adjust that? Thanks!

Thank you for the review @okgolove. Made the respective changes. Please let me know if anything else is needed

okgolove:
@romil93 you have to bump chart's version, lint-test workflow is failed because of that

romil93:
> @romil93 you have to bump chart's version, lint-test workflow is failed because of that

Yeah seems like another PR that bumped up the version went in before this one. Updated it

okgolove (approved):

PR #5998, opened 2025-08-04
[alertmanager] add user namespaces for alertmanager

dongjiang1989:

Add the HostUsers field to support user namespace feature in Kubernetes in alertmanager.

https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/

- fixes #

#### Checklist

- [x] Chart Version bumped


charts/alertmanager/Chart.yaml
```diff
@@ -6,7 +6,7 @@ icon: https://raw.githubusercontent.com/prometheus/prometheus.github.io/master/a
 sources:
   - https://github.com/prometheus/alertmanager
 type: application
-version: 1.23.1
+version: 1.23.2
```
monotek: new features should raise minor version.
```suggestion
version: 1.24.0
```

dongjiang1989:
cc @monotek PTAL

charts/alertmanager/templates/statefulset.yaml
```diff
@@ -33,6 +33,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
+      hostUsers: {{ .Values.hostUsers }}
```
monotek: Have you checked what happens if that field is added an the k8s version does not support it yet?
Some cloud providers don't enable beta features and that option is IMHO GA only from k8s 1.33 and above...

charts/alertmanager/Chart.yaml
dongjiang1989: Thanks @monotek 
Updated

charts/alertmanager/templates/statefulset.yaml
dongjiang1989: Thanks @monotek 
The added doc description and it is set to `false` by default.

dongjiang1989:
@monotek Please re-check it

monotek (approved):

PR #5999, opened 2025-08-04
[kube-prometheus-stack] support encoded string for thanos sidecar secret

trouaux:

allow users to pass directly an encoded string instead of a plaintext obj store config to create Secret for thanos-sidecar

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 77.1.2
+version: 77.1.3
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.85.0
 kubeVersion: ">=1.25.0-0"


charts/kube-prometheus-stack/templates/prometheus/secret.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
     app.kubernetes.io/component: prometheus
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
 data:
-  object-storage-configs.yaml: {{ toYaml .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secret | b64enc | quote }}
+  object-storage-configs.yaml: {{- if .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secretIsBase64 }} {{ toYaml .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secret | quote }}
+{{- else }} {{ toYaml .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secret | b64enc | quote }}
+{{- end }}
```
jkroepke: I'm not really happy with that approach. What about

```suggestion
object-storage-configs.yaml: {{ tpl (toYaml .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secret) $ | b64enc | quote }}
```

Then, you could do `{{ printf "%s" "base64str" | b64dec }}` on your values.

trouaux: in the end i can work without this feature, but can fix CI if it still makes sense to add this

jkroepke: I think the `tpl (toYaml .Values.prometheus.prometheusSpec.thanos.objectStorageConfig.secret) $` is a valid and a more generic approach which may cover more use-cases.

jkroepke (approved):
LGTM

Issue #5980, opened 2025-07-29
[all] Allow enabling user namespaces


sebhoss:
### Is your feature request related to a problem ?

Starting with Kubernetes 1.33, [user namespaces](https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/) are now enabled by default. We want to enable this feature for as many pods as possible.

### Describe the solution you'd like.

It would be great if every chart in this repository would allow consumers to set `.spec.hostUsers` to `false` in all deployments where this is possible. 

### Describe alternatives you've considered.

We currently apply kustomize patches on top of the charts used by us.

jkroepke:
Are you interested into a PR? Keep in mind, each chart needs a distinct PR.

KyriosGN0:
i can take a look into this 👀

jkroepke:
@KyriosGN0 example PR https://github.com/grafana/loki/pull/18661

KyriosGN0:
submitted #6024 for ksm, if that approach is ok , i can continue to the rest of the charts

PR #6024, opened 2025-08-07
[kube-state-metrics]: add option to set user namespaces

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

This PR intorduces the option to use user namespaces in k8s version 1.33 and higher, by default it is unset. 

- relates #5980

#### Checklist

- [X] Chart Version bumped


charts/kube-state-metrics/values.yaml
```diff
@@ -55,6 +55,9 @@ extraArgs: []
 # If false then the user will opt out of automounting API credentials.
 automountServiceAccountToken: true
 
+# -- Use the host's user namespace available in kubernetes 1.33+.
+hostUsers: nil
```
mrueg: why nil and not false by default?

KyriosGN0: to allow users to opt-in into setting this value, since its a new field in beta not all k8s distribution support it by default

KyriosGN0:
sorry, rebased and had some git troubles

KyriosGN0:
hey @dotdc @tariq1890, would it be possible to get a review here? 
Thanks in advance!

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -42,6 +42,9 @@ spec:
       {{ toYaml .Values.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
+      {{- if and (semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version) (kindIs "bool" .Values.hostUsers) }}
+      hostUsers: {{ .Values.hostUsers }}
+      {{- end }}
```
dacamposol: I'd prefer to avoid no-op rendering. Current implementation would only apply the field if the value is a boolean, but Helm will render nothing if it's not, which may be difficult to debug.

I would recommend something more idiomatic, like:

```suggestion
      {{- if and (semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version) (not (empty .Values.hostUsers)) }}
      hostUsers: {{ ternary "true" "false" .Values.hostUsers }}
      {{- end }}
```

This would skip rendering when the key is missing or explicitly `nil`, it would not silently fail when the key has any value, but it would still enforce the value of the flag to hold a boolean value.

dacamposol (changes requested):

charts/kube-state-metrics/templates/deployment.yaml
KyriosGN0: @dacamposol, I see, this makes a lot of sense, i applied this recommendation

zanhsieh (changes requested):
@KyriosGN0 Can you bump the chart version plz?

KyriosGN0:
> @KyriosGN0 Can you bump the chart version plz?

@zanhsieh i did bump the chart's version to 6.5.0

zanhsieh (dismissed):

KyriosGN0:
hey @dacamposol, could a i get a review here? i have addressed your review

dacamposol (dismissed):

KyriosGN0:
hey @dotdc  @tariq1890 , gentle ping on this PR 
Thanks in advance!

zanhsieh (dismissed):

maxwo (dismissed):

KyriosGN0:
@dotdc @tariq1890 @mrueg, could i get a review here? 
Thanks!

jcpunk:
With this deployment not having any physical volumes by default, would it make sense to default to `hostUser: false` in 1.33+ systems?

Note: `hostUsers: false` doesn't work with `kind`.

jcpunk:
Can you rebase this off HEAD, another patch just got merged for kube-state-metrics?

KyriosGN0:
Hey, @dotdc, @tariq1890, @mrueg, i would appreciate it if some could take a look at my PR, thanks!

jcpunk:
I think you need to bump the chart version?

PR #6027, opened 2025-08-08
[prometheus-pgbouncer-exporter] enable pgsql dep only in CI + move refs to bitnami legacy

DrFaust92:

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-pgbouncer-exporter/ci/ci-values.yaml
+++ b/charts/prometheus-pgbouncer-exporter/ci/ci-values.yaml
@@ -10,7 +10,7 @@ rbac:
   pspEnabled: false
 extraContainers:
   - name: pgbouncer
-    image: bitnami/pgbouncer:latest
+    image: bitnamilegacy/pgbouncer:latest
     env:
       - name: POSTGRESQL_PASSWORD
         value: "password"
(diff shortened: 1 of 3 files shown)


charts/prometheus-pgbouncer-exporter/ci/ci-values.yaml
```diff
@@ -10,7 +10,7 @@ rbac:
   pspEnabled: false
 extraContainers:
   - name: pgbouncer
-    image: bitnami/pgbouncer:latest
+    image: bitnamilegacy/pgbouncer:latest
```
jkroepke: bitnamilegacy is gone soon. lets try https://hub.docker.com/r/edoburu/pgbouncer/

DrFaust92: bitnamilegacy replaces bitnami registry though.

I dont mind either way, whatever you prefer

zeritti (commented):
The other thing with respect to Bitnami's images in this chart is the postgresql subchart (dependency) and the corresponding image which this PR does not address.

My suggestion would be to remove the dependency and the extra container from the CI configuration entirely. This would cause no problems in CI as PgBouncer exporter does not require PgBouncer to start successfully. In fact, the CI jobs do not validate nor test whether the exporter functions as expected - launching the target service in CI seems in this sense and in this case unnecessary.

DrFaust92:
[zeritti](https://github.com/zeritti) let me try and if so ill take a look doing it more globally with other such deps (kafka, etc)

DrFaust92:
Thanks @zeritti should be even cleaner now

PR #6097, opened 2025-08-29
[prometheus] use named targetport

DrFaust92:

- fixes #

I didnt cover grpc port as its not present in the pod spec. can be added as part of different PR with other relevant flags

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus
 # renovate: github=prometheus/prometheus
 appVersion: v3.5.0
-version: 27.35.0
+version: 27.36.0
 kubeVersion: ">=1.19.0-0"
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/
--- a/charts/prometheus/templates/_helpers.tpl
+++ b/charts/prometheus/templates/_helpers.tpl
@@ -6,6 +6,9 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Return the port for prometheus.
+*/}}
 {{- define "prometheus.port" -}}
 9090
 {{- end -}}
--- a/charts/prometheus/templates/_helpers.tpl
+++ b/charts/prometheus/templates/_helpers.tpl
@@ -6,6 +6,10 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "prometheus.port" -}}
+9090
+{{- end -}}
+
 {{/*
 Create chart name and version as used by the chart label.
 */}}
--- a/charts/prometheus/templates/deploy.yaml
+++ b/charts/prometheus/templates/deploy.yaml
@@ -235,7 +235,11 @@ spec:
             {{- if not .Values.server.tcpSocketProbeEnabled }}
             httpGet:
               path: {{ .Values.server.prefixURL }}/-/ready
+              {{- if .Values.server.portName }}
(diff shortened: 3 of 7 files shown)


zanhsieh (dismissed):

charts/prometheus/templates/_helpers.tpl
```diff
@@ -6,6 +6,10 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "prometheus.port" -}}
```
zeritti, line 12: Please, add a comment to the helper template

charts/prometheus/templates/deploy.yaml
```diff
@@ -235,15 +235,23 @@ spec:
             {{- if not .Values.server.tcpSocketProbeEnabled }}
             httpGet:
               path: {{ .Values.server.prefixURL }}/-/ready
-              port: 9090
+              {{- if .Values.server.portName }}
+              port: {{ .Values.server.portName }}
+              {{- else }}
+              port: {{ include "prometheus.port" . }}
+              {{- end }}
```
zeritti: Each conditional could be simplified as there is a default value and an overriding (optional) value there:

```suggestion
              port: {{ default (include "prometheus.port" .) .Values.server.portName }}
```

charts/prometheus/templates/deploy.yaml
```diff
@@ -224,7 +224,7 @@ spec:
           {{- end }}
         {{- end }}
           ports:
-            - containerPort: 9090
+            - containerPort: {{ include "prometheus.port" . }}
```
zeritti, line 227: Please, consider adjusting the port at lines 115 and 117 as well.

charts/prometheus/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus
 # renovate: github=prometheus/prometheus
 appVersion: v3.5.0
-version: 27.32.0
+version: 27.32.1
```
zeritti: The change feels more like a new feature extending use of `server.portName` than a bugfix/patch.

zeritti (changes requested):
Thank you, @DrFaust92, for your PR. Please, see my comments below.

charts/prometheus/templates/headless-svc.yaml
```diff
@@ -19,7 +19,7 @@ spec:
     - name: http
       port: {{ .Values.server.statefulSet.headless.servicePort }}
       protocol: TCP
-      targetPort: 9090
+      targetPort: {{ default (include "prometheus.port" .) .Values.server.portName }}
```
DrFaust92, line 22: @zeritti I wonder if its worth to set this as another param in helper to not repeat it

charts/prometheus/templates/network-policy.yaml
```diff
@@ -12,5 +12,5 @@ spec:
       {{- include "prometheus.server.matchLabels" . | nindent 6 }}
   ingress:
     - ports:
-      - port: 9090
+      - port: {{ include "prometheus.port" . }}
```
DrFaust92: does this take port name? unsure so keeping as is for now

DrFaust92:
Thanks @zeritti should be cleaner now

zanhsieh (dismissed):

charts/prometheus/templates/headless-svc.yaml
zeritti: Thanks, @DrFaust92. I think we can leave it.

charts/prometheus/templates/network-policy.yaml
zeritti: Yes, it does.

```suggestion
      - port: {{ default (include "prometheus.port" .) .Values.server.portName }}
```

zeritti (changes requested):

PR #6101, opened 2025-09-01
[kube-prometheus-stack] set expected jobLabel for node-exporter PodMonitor

z0rc:

Set pre-defined jobLabel for node-exporter's PodMonitor when deployed via kube-prometheus-stack. This way node-exporter metrics label `job` remains stable and expected when switching from default ServiceMonitor to PodMonitor.

N/A

#### Checklist

- [x] Chart Version bumped


charts/kube-prometheus-stack/values.yaml
```diff
@@ -2608,6 +2608,11 @@ prometheus-node-exporter:
       # attachMetadata:
       #   node: false
 
+    podMonitor:
+      enabled: false
+
+      jobLabel: jobLabel
```
QuentinBisson, line 2614: I don't think you want to use jobLabel here and we probably do not want to maintain 2 différent fields for this. @jkroepke any thoughts other than a breaking  change?

jkroepke: Hm, I have the feeling that this change make sense, because all other serviceMonitors have `jobLabel: jobLabel` as well.

z0rc: https://github.com/prometheus-community/helm-charts/blob/109b69a45e89b28fa63f102442f9e99eb50bb2e2/charts/kube-prometheus-stack/values.yaml#L2551

ServiceMonitor for node-exporter has explicitly defined `jobLabel`. I just replicated the same configuration for PodMonitor.

When I switched from from ServiceMonitor to PodMonitor, the metric's `job` label changed from `node-exporter` (explicitly set by `kube-prometheus-stack` default values) to `prometheus-node-exporter` (fallback to `app.kubernetes.io/name` label value, when jobLabel is empty, which is node-exporter's chart default).

With incorrect `job` label on metrics, a lot of prom rules and grafana dashboards are broken. So I consider this as fix, rather than a breaking change.

QuentinBisson: I checked a bit more and you both seem to be right, just looking at the values I thought this change was about using an anchor. Thanks for the contribution @z0rc

QuentinBisson (approved):

PR #6066, opened 2025-08-19
[alertmanager] Add support for PVC annotations in `StatefulSet`

rb090:
#### What this PR does / why we need it

This PR adds support for passing `persistence.annotations` and `persistence.labels` to the PVC created by the alertmanager `StatefulSet`.

Currently, it is not possible to attach custom annotations to the `alertmanager.persistence` or custom labels through `persistence.labels`. This limits the ability to write configuration for persistence like set encryption settings over `annotations`.  And to have `labels` to fe. organize objects for the purpose to query them.

With this change, users can now set custom annotations to configure custom storage attributes or encryption keys.

**Example** use case:  
On Open Telekom Cloud (Everest CSI), PVC annotations are required to enable features like disk type selection, KMS encryption keys, or custom volume name prefixes:

```yaml
alertmanager:
  enabled: true
  persistence:
    enabled: true
    storageClass: csi-disk-topology
    size: 5Gi
    annotations:
      everest.io/disk-volume-type: "SAS"
      everest.io/crypt-key-id: "xxxx-xxxx-xxxx"
      everest.io/csi.volume-name-prefix: "whatever-here"
      .....
    labels:
      app.kubernetes.io/part-of: monitoring
      app.kubernetes.io/component: alertmanager
      .....
```

#### Special notes for your reviewer

@monotek @naseemkullah I really hope I did everything correct and followed the best practices. I did really my best with that.

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6116, opened 2025-09-04
[prometheus] Add new persistence attributes from `alertmanager` to prometheus values file

rb090:

This PR follows up on [PR #6066 – [alertmanager] Add support for PVC annotations in StatefulSet](https://github.com/prometheus-community/helm-charts/pull/6066), which introduced support for `annotations` and `labels` in the `alertmanager` persistence configuration.

The `values.yaml` and `values.schema.json` have been updated to include the new persistence attributes from the `alertmanager` chart. The goal is to ensure consistency on those values.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/values.schema.json
+++ b/charts/prometheus/values.schema.json
@@ -12,41 +12,12 @@
                     "type": "boolean"
                 },
                 "persistence": {
-                  "type": "object",
-                  "required": ["enabled", "size"],
-                  "properties": {
-                    "enabled": {
-                      "type": "boolean"
-                    },
-                    "storageClass": {
-                      "type": "string"
-                    },
-                    "accessModes": {
-                      "type": "array",
-                      "items": {
-                        "type": "string"
-                      }
-                    },
-                    "size": {
-                      "type": "string"
-                    },
-                    "annotations": {
-                      "type": "object",
-                      "description": "Custom annotations to apply to the PersistentVolumeClaim created by the Alertmanager StatefulSet.",
-                      "additionalProperties": {
-                        "type": "string"
-                      },
-                      "default": {}
-                    },
-                    "labels": {
-                      "type": "object",
-                      "description": "Custom labels to apply to the PersistentVolumeClaim created by the Alertmanager StatefulSet.",
-                      "additionalProperties": {
-                        "type": "string"
-                      },
-                      "default": {}
+                    "type": "object",
+                    "properties": {
+                        "size": {
+                            "type": "string"
+                        }
                     }
-                  }
                 },
                 "podSecurityContext": {
                     "type": "object",


rb090:
@zanhsieh Thank you very much for your review and approval. I’ve also updated this branch with the latest changes from the main branch, the CI checks are all passing. It looks like I don’t have the necessary permissions to merge this PR. Could you please merge it to main?

charts/prometheus/values.yaml
```diff
@@ -1311,7 +1311,33 @@ alertmanager:
...
+    ## Persistent Volume Storage Class
+    ## If defined, storageClassName: <storageClass>
+    ## If set to "-", storageClassName: "", which disables dynamic provisioning
+    ## If undefined (the default) or set to null, no storageClassName spec is
+    ## set, choosing the default provisioner.
+    ##
+    # storageClass: "-"
+    accessModes:
+      - ReadWriteOnce
+    size: 50Mi
```
zeritti: Changing a default value is a major change. Users would need to know that it has changed via an upgrade note.
An upgrade without specifying the field's custom (user's) value will fail as the corresponding field in the statefulset is immutable. Again, users would have to be told what needs to be done before upgrading.

Please, keep the previous value. With all respect, there is no reason for changing it.

charts/prometheus/values.schema.json
```diff
@@ -12,12 +12,41 @@
                     "type": "boolean"
                 },
                 "persistence": {
-                    "type": "object",
-                    "properties": {
-                        "size": {
-                            "type": "string"
-                        }
+                  "type": "object",
```
zeritti: This JSON schema file applies to the prometheus chart itself, it should not include subcharts' field definitions. The subchart, in this case alertmanager, has its own JSON schema that gets applied.

zeritti (changes requested):
Thank you, @rb090, for your PR. Please, see my comments below.

charts/prometheus/values.schema.json
rb090: @zeritti thank you for the clarification.

Before my change, the `persistence` object in `charts/prometheus/values.schema.json` was defined as:

```
"persistence": {
    "type": "object",
    "properties": {
        "size": {
            "type": "string"
        }
    }
},
```

My question now is:
Should I remove the entire `properties` object from the Prometheus schema too, or should I simply revert the `alertmanager` object to how it was?

```
"alertmanager": {
    "type": "object",
    "properties": {
        "enabled": {
            "type": "boolean"
        },
        "persistence": {
            "type": "object",
            "properties": {
                "size": {
                    "type": "string"
                }
            }
        },
        "podSecurityContext": {
            "type": "object",
            "properties": {
                "fsGroup": {
                    "type": "integer"
                },
                "runAsGroup": {
                    "type": "integer"
                },
                "runAsNonRoot": {
                    "type": "boolean"
                },
                "runAsUser": {
                    "type": "integer"
                }
            }
        }
    }
},
```

charts/prometheus/values.yaml
rb090: @zeritti thank you for the feedback on this, good point. I’ve reverted the default back to `2Gi` with https://github.com/prometheus-community/helm-charts/pull/6116/commits/940faca9fc43746d866facd9db549cf0f7e35b04.

I noticed that in the `alertmanager` chart itself the default for `persistence.size` is currently `50Mi`. To keep things consistent across charts, do you think it would make sense to align that value to `2Gi` as well? Of course, that would be a separate PR, but I am just thinking if it make sense.

charts/prometheus/values.schema.json
zeritti: Please, revert to the previous state of the schema file.

zeritti (changes requested):

charts/prometheus/values.yaml
zeritti: In my opinion, that would be unnecessary. The prometheus chart chose different defaults as charts often do for their subcharts.

charts/prometheus/values.schema.json
rb090: Okay, no problem, done with https://github.com/prometheus-community/helm-charts/pull/6116/commits/7b4e7c1b47c24754ff152263c084c3f32cead58c 🙂.

PR #6131, opened 2025-09-13
[kube-prometheus-stack] Pass Grafana ds and dashboard `label` and `labelValue` thru `tpl`

mahendrapaipuri:

According to [Grafana docs](https://github.com/grafana/helm-charts/blob/d73a322972ab1ef4220154199b1e28405b60d53a/charts/grafana/values.yaml#L1005-L1008) datasource sidecar's `label` and `labelValue` can be templated. This commit adds those missing `tpl` functions in default datasource template. Similarly dashboard sidecar `label` and `labelValue` also support [templating](https://github.com/grafana/helm-charts/blob/d73a322972ab1ef4220154199b1e28405b60d53a/charts/grafana/values.yaml#L1087-L1090) which is added in `sync_grafana_dashboards.py` file.

Currently downstream charts that have `kube-prometheus-stack` as dependency do not allow templating of `label` and `labelValues` of the Grafana datasource and dashboard configmaps/secrets due to missing `tpl` functions in templates of Grafana default datasources and dashboards. This PR fixes!

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
+++ b/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
@@ -132,7 +132,7 @@ metadata:
 {{ toYaml .Values.grafana.sidecar.dashboards.annotations | indent 4 }}
   labels:
     {{- if $.Values.grafana.sidecar.dashboards.label }}
-    {{ $.Values.grafana.sidecar.dashboards.label }}: {{ ternary $.Values.grafana.sidecar.dashboards.labelValue "1" (not (empty $.Values.grafana.sidecar.dashboards.labelValue)) | quote }}
+    {{ tpl $.Values.grafana.sidecar.dashboards.label $ }}: {{ ternary (tpl $.Values.grafana.sidecar.dashboards.labelValue $) "1" (not (empty $.Values.grafana.sidecar.dashboards.labelValue)) | quote }}
     {{- end }}
     app: {{ template "kube-prometheus-stack.name" $ }}-grafana
 {{ include "kube-prometheus-stack.labels" $ | indent 4 }}
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 77.6.2
+version: 77.6.3
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.85.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
+++ b/charts/kube-prometheus-stack/templates/grafana/configmaps-datasources.yaml
@@ -9,7 +9,7 @@ metadata:
     {{- toYaml .Values.grafana.sidecar.datasources.annotations | nindent 4 }}
 {{- end }}
   labels:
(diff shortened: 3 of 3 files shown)


charts/kube-prometheus-stack/templates/grafana/dashboards-1.14/apiserver.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
 {{ toYaml .Values.grafana.sidecar.dashboards.annotations | indent 4 }}
   labels:
     {{- if $.Values.grafana.sidecar.dashboards.label }}
-    {{ $.Values.grafana.sidecar.dashboards.label }}: {{ ternary $.Values.grafana.sidecar.dashboards.labelValue "1" (not (empty $.Values.grafana.sidecar.dashboards.labelValue)) | quote }}
+    {{ tpl $.Values.grafana.sidecar.dashboards.label $ }}: {{ ternary (tpl $.Values.grafana.sidecar.dashboards.labelValue $) "1" (not (empty $.Values.grafana.sidecar.dashboards.labelValue)) | quote }}
```
jkroepke: How `(not (empty $.Values.grafana.sidecar.dashboards.labelValue)) ` should be handeled? Because the value can be non empty, but empty after tpl

mahendrapaipuri: Ah, my bad! I missed it. We should pass it through the `tpl` function and then check for `empty` condition: `(not (empty (tpl $.Values.grafana.sidecar.dashboards.labelValue $)))`. What do you think?

jkroepke: what did you think about `{{ (tpl $.Values.grafana.sidecar.dashboards.labelValue $) | default "1" }}`

mahendrapaipuri: Yeah, much simpler and cleaner!! I will push the changes!

charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
```diff
@@ -132,7 +132,7 @@ def new_representer(dumper, data):
 {{ toYaml .Values.grafana.sidecar.dashboards.annotations | indent 4 }}
   labels:
     {{- if $.Values.grafana.sidecar.dashboards.label }}
-    {{ $.Values.grafana.sidecar.dashboards.label }}: {{ ternary $.Values.grafana.sidecar.dashboards.labelValue "1" (not (empty $.Values.grafana.sidecar.dashboards.labelValue)) | quote }}
+    {{ tpl $.Values.grafana.sidecar.dashboards.label $ }}: {{ ((tpl $.Values.grafana.sidecar.dashboards.labelValue $) | quote) | default "1" }}
```
jkroepke: `{{ ((tpl $.Values.grafana.sidecar.dashboards.labelValue $) | quote) | default "1" }}`

I quite the quote function should be outside/after default functions. An empty values would result into `""` (including quotes) and the default function never hits

mahendrapaipuri: You are right! Cheers for the pointer. I thought Go templates will complain passing a `int` default to a `string` value. Seems like that is not the [case](https://helm-playground.com/#t=LQhQ4bwgCAKWAuAHANtAdANQIYoK4CmAzugOYBO2AZtgHbbpECWAJgQMbbnovZEAWAIwD2XFiRTZBBFDnwEMASmgAfaGxp4UCaAEZlagI55hCBQF9zQA&v=LQhQAJPUHMCcCGAzBA7BAuC4DOBLAEwFMBjBOLKcAhHACwCMB7cgnSq8AGwQaK4BqCLgFciGcACJJoIA). Will push the changes!

PR #6135, opened 2025-09-16
[prometheus-kafka-exporter] Add new values to set TLS file names

michaelbeutler:

This PR updates the **prometheus-kafka-exporter** Helm chart to allow users to specify custom TLS file names via new values:  
```yaml
tls:
  ...
  # secretName: <name of an existing secret>
  caFile: ca.crt
  certFile: tls.crt
  keyFile: tls.key

server:
  tls:
    ...
    # secretName: <name of an existing secret>
    caFile: ca.crt
    certFile: tls.crt
    keyFile: tls.key
```

The deployment template has been updated to use these values for both the main and server TLS configuration, replacing previously hardcoded file names.  

Default values are set to:
- `ca.crt`
- `tls.crt`
- `tls.key`

This change makes the chart more flexible when working with secrets created by tools like the [Strimzi Kafka Operator](https://strimzi.io/), which names secret keys differently.

- The change is backward compatible, since defaults align with existing behavior.  
- No changes are required for existing users unless they want to override the file names.  

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-kafka-exporter]`)

### DIFF
--- a/charts/prometheus-kafka-exporter/Chart.yaml
+++ b/charts/prometheus-kafka-exporter/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v1.9.0"
 description: A Helm chart to export the metrics from Kafka in Prometheus format using the kafka-exporter from https://github.com/danielqsj/kafka_exporter
 name: prometheus-kafka-exporter
 home: https://github.com/danielqsj/kafka_exporter
-version: 2.17.0
+version: 2.18.0
 kubeVersion: ">=1.19.0-0"
 sources:
   - https://gkarthiks.github.io/helm-charts/charts/prometheus-kafka-exporter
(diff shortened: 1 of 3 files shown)


zeritti (commented):
Thank you, @michaelbeutler, for your PR, LGTM.

It appears that we have a bug with respect to `.server.tls`. If configured as required, the needed flags do not get set at all. The reason seems to be in `deployment.yaml` in that the required flags get set only if `sasl.enabled` is set which is not correct - the exporter's TLS settings do not depend on SASL configuration. The conditional at line 68 should end on line 91 and not on line 97. Would you be willing to correct it?

zeritti (approved):
Thank you, @michaelbeutler, for your changes.

PR #5815, opened 2025-06-21
[prometheus-kafka-exporter]: remove support for psp

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

#### What this PR does / why we need it

remove support for psp 

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- relates #5671

#### Checklist

- [X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6157, opened 2025-09-23
[prometheus-kafka-exporter] remove psp support, require K8s 1.25

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

Removes support for PSP resources, requires Kubernetes 1.25

- relates #5671

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 77.10.0
+version: 77.11.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.85.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
+++ b/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
@@ -29,11 +29,11 @@ def change_style(style, representer):
 
 refs = {
     # renovate: git-refs=https://github.com/prometheus-operator/kube-prometheus branch=main
-    'ref.kube-prometheus': '760b786b29edb4aa917e87746cb9e2cab9cf3066',
+    'ref.kube-prometheus': '68619d9afe02555c25b784ed2e73e907380f0876',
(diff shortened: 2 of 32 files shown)


KyriosGN0:
i had a really bad rebase in #5815, so i re-branched to have a clear commit history and opened this PR

charts/prometheus-kafka-exporter/README.md
```diff
@@ -55,3 +55,11 @@ See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_h
...
 helm show values oci://ghcr.io/prometheus-community/charts/prometheus-kafka-exporter
 ```
+
+### Upgrading an existing Release to a new major version
+
+A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.
+
+#### 2.x to 3.x
+
+The Helm parameter `rbac.pspEnabled` was removed (was set to to false before)
```
zeritti: Please, be more specific as to what the removal means. E.g. "From release 3.0, this helm chart does not support pod security policies any longer".

The conditional in template file `role.yaml` (L11-17) has to be removed as it refers to PSP.

Since PSP won't be supported by the chart any longer, I'd suggest to bump `kubeVersion` in Chart.yaml to 1.25. That would be consistent as PSP were removed in that release and I believe we have to move a bit in this respect. A short note in README will be needed, though.

KyriosGN0: @zeritti Thanks! i have address your comments

charts/prometheus-kafka-exporter/README.md
```diff
@@ -55,3 +55,11 @@ See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_h
...
 helm show values oci://ghcr.io/prometheus-community/charts/prometheus-kafka-exporter
 ```
+
+### Upgrading an existing Release to a new major version
+
+A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.
+
+#### 2.x to 3.x
+
+From version 3.0.0, support for pod security policies was removed, this means this chart doesn't support kubernetes version earlier then 1.25.
```
zeritti: Let us slightly adjust the wording if it is fine with you.

 
```suggestion
From version 3.0.0, support for pod security policies was removed.

The minimum Kubernetes version required by the helm chart is 1.25.
```

PR #6178, opened 2025-10-01
[kube-prometheus-stack] explicitly allow a null Prometheus ruleSelector

ba-work:
This PR enables the user to explicitly set a `Prometheus` CRD's `spec.ruleSelector` value to `null`. This is a supported/documented value that prevents the operator from importing any `PrometheusRule`s. You can check this with `kubectl explain prometheus.spec.ruleSelector` or [see it documented here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api-reference/api.md?plain=1#L3673-L3675)

I encountered a situation where this is required (CRs created and validated, but not imported), and this chart doesn't allow explicitly setting `prometheus.prometheusSpec.ruleSelector=null`; currently it always coalesces to `{}`.

I made it also require `prometheus.prometheusSpec.ruleSelectorNilUsesHelmValues=false` since that seemed like the most intuitive and backwards compatible with current behavior.

Tested with:
```bash
helm install -g charts/kube-prometheus-stack \
  --set prometheus.prometheusSpec.ruleSelectorNilUsesHelmValues=false \
  --set prometheus.prometheusSpec.ruleSelector=null
```

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/unittests/prometheus/rule_selector_test.yaml
+++ b/charts/kube-prometheus-stack/unittests/prometheus/rule_selector_test.yaml
@@ -35,9 +35,8 @@ tests:
           ruleSelector: null
           ruleSelectorNilUsesHelmValues: false
     asserts:
-      - equal:
+      - notExists:
           path: spec.ruleSelector
-          value: null
   - it: should match the release name if set to null and ruleSelectorNilUsesHelmValues is true
     set:
       prometheus:


charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -274,6 +274,8 @@ spec:
   ruleSelector:
     matchLabels:
       release: {{ $.Release.Name | quote }}
+{{ else if eq .Values.prometheus.prometheusSpec.ruleSelector nil }}
```
jkroepke:
```suggestion
{{ else if not (kindIs "invalid" .Values.prometheus.prometheusSpec.ruleSelector) }}
```

Lets use the same pattern as used on other values.

jkroepke (changes requested):
Could you please extend the unit tests in that case, maybe start by copy this file:

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/unittests/prometheus/scrape_config_selector_test.yaml

charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
ba-work: This would bring it in line with the other fields, but note that this would be a breaking change. With the suggested change, it will now result in `null` when exclusively setting `prometheus.prometheusSpec.ruleSelectorNilUsesHelmValues=false`, where it currently results in `{}`

image

Is this acceptable?

ba-work: corrected this discrepancy

ba-work:
> Could you please extend the unit tests in that case, maybe start by copy this file:
> 
> [`main`/charts/kube-prometheus-stack/unittests/prometheus/scrape_config_selector_test.yaml](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/unittests/prometheus/scrape_config_selector_test.yaml)

done!

ba-work:
for anyone that happens to stumble on this before its merged or closed, turns out you can get the same behavior but it relies on the way the operator currently handles on config precedence: **if there are keys with the same name, the latter wins.**

so that means, as of this comment, the following will result `ruleSelector: null` when the CR is actually applied to the cluster:
```bash
helm install -g charts/kube-prometheus-stack \
  --set prometheus.prometheusSpec.ruleSelectorNilUsesHelmValues=false \
  --set prometheus.prometheusSpec.ruleSelector=null \
  --set prometheus.prometheusSpec.additionalConfig.ruleSelector=null
```

Really, this will make helm generate this:
```yaml
---
# Source: kube-prometheus-stack/templates/prometheus/prometheus.yaml
kind: Prometheus
...
spec:
  ...
  ruleSelector: {}
  ...
  ruleSelector: null # this one wins
```

ba-work:
@jkroepke renovate-bot necessitated another version bump. sorry!

charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
```diff
@@ -274,8 +274,10 @@ spec:
   ruleSelector:
     matchLabels:
       release: {{ $.Release.Name | quote }}
-{{ else }}
+{{ else if not (kindIs "invalid" .Values.prometheus.prometheusSpec.ruleSelector) }}
   ruleSelector: {}
+{{ else }}
+  ruleSelector: null
```
jkroepke:
```suggestion
```

charts/kube-prometheus-stack/unittests/prometheus/rule_selector_test.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - it: should be null if set to null and ruleSelectorNilUsesHelmValues is false
+    set:
+      prometheus:
+        prometheusSpec:
+          ruleSelector: null
+          ruleSelectorNilUsesHelmValues: false
+    asserts:
+      - equal:
+          path: spec.ruleSelector
+          value: null
```
jkroepke:
```suggestion
    asserts:
      - notExists:
          path: spec.ruleSelector
```

PR #5486, opened 2025-03-28
[prometheus-yet-another-cloudwatch-exporter] allow templating serviceaccount annotation values

fcrespofastly:
#### What this PR does / why we need it
This allows templating on `serviceaccount.annotations` values which facilitates the use of IRSA specifically. Platform maintainer usually have/want centralized configuration metadata that we pass to every single app: AWS account IDs, regions, cluster name, etc. If you don't have this functionality then you have to pass the IRSA annotation for every cluster (should the role be different per cluster, which it's our case) repeating the same metadata again and again.

I proposed this change in well recognized and maintained CNCF apps like External DNS, Cert Manager and it got accepted:

https://github.com/kubernetes-sigs/external-dns/pull/4958
https://github.com/cert-manager/cert-manager/pull/7501

This change does not break previous functionality and not templated values can still be set and they will render the same way it was rendering before.

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6224, opened 2025-10-10
[prometheus-stackdriver-exporter] allow templating sa annotations

fcrespofastly:

This allows templating on `serviceaccount.annotations` values which facilitates the use of Workload Identity specifically. Platform maintainer usually have/want centralized configuration metadata that we pass to every single app: GCP project IDs, regions, cluster name, etc. If you don't have this functionality then you have to pass the Workload Identity annotation for every cluster (should the role be different per cluster, which it's our case) repeating the same metadata again and again.

I proposed the very same change and was accepted in prometheus-yet-another-cloudwatch-exporter as well as in CNCF projects like External DNS, Cert Manager and Cluster Autoscaler:

https://github.com/prometheus-community/helm-charts/pull/5486
https://github.com/kubernetes-sigs/external-dns/pull/4958
https://github.com/cert-manager/cert-manager/pull/7501
https://github.com/kubernetes/autoscaler/pull/7936

This change does not break previous functionality and not templated values can still be set and they will render the same way it was rendering before.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-stackdriver-exporter/templates/serviceaccount.yaml
+++ b/charts/prometheus-stackdriver-exporter/templates/serviceaccount.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     {{- include "stackdriver-exporter.labels" . | indent 4 }}
   name: {{ template "stackdriver-exporter.serviceAccountName" . }}
-  {{-  if .Values.serviceAccount.annotations }}
-  annotations: {{- tpl (toYaml .Values.serviceAccount.annotations) $ | nindent 4 }}
+  {{- with .Values.serviceAccount.annotations }}
+  annotations: {{- tpl (toYaml .) $ | nindent 4 }}
   {{- end }}
 {{- end -}}


charts/prometheus-stackdriver-exporter/templates/serviceaccount.yaml
```diff
@@ -5,8 +5,10 @@ metadata:
   labels:
     {{- include "stackdriver-exporter.labels" . | indent 4 }}
   name: {{ template "stackdriver-exporter.serviceAccountName" . }}
-  {{- if .Values.serviceAccount.annotations }}
+  {{- with .Values.serviceAccount.annotations }}
   annotations:
-    {{- toYaml .Values.serviceAccount.annotations | nindent 4 }}
+    {{- range $k, $v := . }}
+      {{- printf "%s: %s" (tpl $k $) (tpl $v $) | nindent 4 }}
+    {{- end }}
```
jkroepke: Whats the benefit over

```suggestion
{{- tpl (toYaml .) $ | nindent 4 }}
```

fcrespofastly: @jkroepke the main advantage is that with the range you can also template keys if you want. Was one of the feedback comments I got when I first proposed this change I think in external-dns or cert-manager.

jkroepke: I don't get this. With toYaml, you can do this as well?

```
annotations:
  a: b
  "{{ .Release.Name }}": "value"
```

and `{{ .Release.Name }}` would be templated as well.

fcrespofastly: Oh my bad, you're right! It's just the pattern I followed everywhere else but I don't mind switching to `tpl (toYaml .) $`

fcrespofastly: pushed!

charts/prometheus-stackdriver-exporter/templates/serviceaccount.yaml
```diff
@@ -5,8 +5,7 @@ metadata:
   labels:
     {{- include "stackdriver-exporter.labels" . | indent 4 }}
   name: {{ template "stackdriver-exporter.serviceAccountName" . }}
-  {{- if .Values.serviceAccount.annotations }}
-  annotations:
-    {{- toYaml .Values.serviceAccount.annotations | nindent 4 }}
+  {{-  if .Values.serviceAccount.annotations }}
+  annotations: {{- tpl (toYaml .Values.serviceAccount.annotations) $ | nindent 4 }}
```
jkroepke:
```suggestion
  {{- with .Values.serviceAccount.annotations }}
  annotations: {{- tpl (toYaml .) $ | nindent 4 }}
```

PR #6228, opened 2025-10-11
[kube-prometheus-stack] Parameterize job names

abelfodil:
Parameterize `kubelet`, `kube-controller-manager`, `kube-scheduler` and `kube-proxy` job names.

This is needed for k3s clusters which use Rancher's `pushprox` to pull metrics. All these jobs would be overriden to `k3s-server` in a k3s setup with pushprox (example: https://github.com/arch-anes/self-hosted-services/commit/d744c58e44bfba68f2c5bd3289a571ad7a6094b4)

None in this repo.

1. https://github.com/arch-anes/self-hosted-services/issues/522#issuecomment-3393473859
2. I diffed the before and after resulting `helm template`. There was no difference.

#### Checklist

- [x] Chart Version bumped


charts/kube-prometheus-stack/hack/sync_prometheus_rules.py
```diff
@@ -205,10 +205,22 @@ def new_representer(dumper, data):
         'replacement': 'job="{{ $kubeStateMetricsJob }}", namespace{{ $namespaceOperator }}"{{ $targetNamespace }}"',
         'limitGroup': ['kubernetes-apps'],
         'init': '{{- $targetNamespace := .Values.defaultRules.appNamespacesTarget }}{{- $namespaceOperator := .Values.defaultRules.appNamespacesOperator | default "=~" }}'},
-    'job="kubelet"': {
-        'replacement': 'job="kubelet", namespace{{ $namespaceOperator }}"{{ $targetNamespace }}"',
+    'job="kubelet", metrics_path="/metrics': {
```
jkroepke, line 208:
```
    'job="kubelet", metrics_path="/metrics': {
```

Not sure about this?

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1598,6 +1598,8 @@ kubelet:
   enabled: true
   namespace: kube-system
 
+  jobNameOverride: kubelet
```
jkroepke: Please add a comment here.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -1838,6 +1840,8 @@ kubelet:
 kubeControllerManager:
   enabled: true
 
+  jobNameOverride: kube-controller-manager
```
jkroepke: Please add a comment here.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -2234,6 +2238,8 @@ kubeEtcd:
 kubeScheduler:
   enabled: true
 
+  jobNameOverride: kube-scheduler
```
jkroepke: Please add a comment here.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -2340,6 +2346,8 @@ kubeScheduler:
 kubeProxy:
   enabled: true
 
+  jobNameOverride: kube-proxy
```
jkroepke: Please add a comment here.

charts/kube-prometheus-stack/hack/sync_prometheus_rules.py
abelfodil: The output remains identical, but now I can add a second `job="kubelet` statement to target other files.

jkroepke:
What did you think about shenanigans like

https://github.com/portworx/helm/blob/9c057a2b28fb01e6be54349328c15bdd3a2b1503/charts/portworx-daemonSet/templates/portworx-ds.yaml#L3

to automaticly detect k3s and use the correct names? Maybe keep the `jobNameOverride` empty by default for auto-configure but let users allow to override the names?

abelfodil:
> to automaticly detect k3s and use the correct names

It's similar to what `rancher-monitoring` does with https://github.com/rancher/charts/blob/dev-v2.12/charts/rancher-monitoring/107.2.0%2Bup69.8.2-rancher.20/templates/_helpers.tpl#L64-L94.

If the project allows it, I could include a turn-key k3s configuration in a subsequent PR. I know people struggled a lot with this chart + k3s, so it would be great:
1. https://github.com/arch-anes/self-hosted-services/issues/16 (myself)
2. https://github.com/SovereignCloudStack/k8s-observability/issues/48
3. https://github.com/k3s-io/k3s/issues/3619
4. https://github.com/grafana/k8s-monitoring-helm/issues/1145

The rancher chart (https://github.com/rancher/charts/blob/main/charts/rancher-monitoring) solves all the problems but also creates new ones (documented in https://github.com/arch-anes/self-hosted-services/issues/522).

A critical piece though is the pushprox proxy (https://ranchermanager.docs.rancher.com/integrations-in-rancher/monitoring-and-alerting/how-monitoring-works). Can we also include (as a dependency?) it in this project?

abelfodil:
> Maybe keep the jobNameOverride empty by default for auto-configure but let users allow to override the names?

Done

jkroepke:
> pushprox proxy

No not really. We would like to avoid distribution specific dependencies, since we cant test them.

abelfodil:
Understood, then we could simply document the setup in some readme.

abelfodil:
@jkroepke anything left blocking the merge?

PR #6256, opened 2025-10-21
[prometheus-postgres-exporter] introduce podmonitoring

joawin:
- Allows use of PodMonitor instead of the traditional ServiceMonitor.

#### Checklist

- [X] Chart Version bumped


zanhsieh (dismissed):

zeritti (commented):
Thank you, @joawin, for your PR. I am of the opinion that we should not include custom resources from outside of _monitoring.coreos.com_ in the chart as is the case of _PodMonitoring_. Please, consider using the field `extraManifests` for any additional resources you'd like to deploy.

zeritti (commented):
@joawin By using the values' field `extraManifests` I meant using this field for your custom resources in your own values at a release install/upgrade. In other words, there is no need to introduce _PodMonitoring_ in the chart in any form.

joawin:
@zeritti  why not. We want to deploy it on GCP environment where we need it as Servicemonitor is not supported.
For Blackbox-Exporter it is also implemented

charts/prometheus-postgres-exporter/templates/extra-manifests.yaml
```diff
@@ -1,4 +1,14 @@
...
+  name: {{ template "prometheus-postgres-exporter.fullname" $ }}
+  namespace: {{ template "prometheus-postgres-exporter.namespace" $ }}
+  labels:
+    {{- include "prometheus-postgres-exporter.labels" $ | nindent 4 }}
+    {{- if or $.Values.podMonitoring.defaults.labels .labels }}
+      {{- toYaml (.labels | default $.Values.podMonitoring.defaults.labels) | nindent 4 }}
+    {{- end }}
 {{ tpl (toYaml .) $ }}
 {{ end }}
+{{- end }}
```
QuentinBisson:
```suggestion
{{- end }}

```

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -291,14 +306,18 @@ extraVolumeMounts: []
...
 ## Extra manifests to deploy as an array
-extraManifests: []
-  # - apiVersion: v1
-  #   kind: ConfigMap
-  #   metadata:
-  #   labels:
-  #     name: postgres-exporter-extra
-  #   data:
-  #     extra-data: "value"
+extraManifests:
```
QuentinBisson, line 309: Is there a reason you are not providing it as an extra object directly?

charts/prometheus-postgres-exporter/templates/extra-manifests.yaml
joawin: newline was added

charts/prometheus-postgres-exporter/values.yaml
joawin: the idea was to enable it via podMonitoring.enabled in the values yaml. Ideas on how I could solve it better

QuentinBisson: Sure but the initial idea from @zeritti would be that you configure the podmonitoring via your values and not in the chart template

joawin:
Seems like podmonitor was introduced with https://github.com/prometheus-community/helm-charts/pull/7083

PR #6281, opened 2025-10-31
[kube-prometheus-stack] Pass through Thanos configuration

abelfodil:

Pass through Thanos configuration instead of recreating every field manually. This would stabilize the feature as it would follow https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api-reference/api.md#thanosspec 1:1.

None

#### Checklist

- [x] Chart Version bumped


jkroepke:
Hm, it feels like just breaking things. (No Secret anymore)

abelfodil:
I don't see the need of defining the secret through the chart. It's not an object from CRD, so people can define theirs separately and reference it. If people really want to define a secret through the chart, they can use `extraSecret` and reference it in Thanos sidecar spec.

While it does break things indeed, it also opens more opportunities for customization.

jkroepke:
Hm. I say that more opportunities are possible, without breaking existing things.

abelfodil:
Ack

PR #6284, opened 2025-10-31
[crds] fix missing quotes

sim500:
It fixes missing quotes in certain values of the `alertmanagerconfig` CRDs. The lack of quotes causes issues with certain YAML parsers.


QuentinBisson:
I think the CRDS are coming from the prometheus operator project. You probably want to have it fixed there instead

jkroepke:
Yep, they need to be fixed upstream. Otherwise, renovate will reset the changes.

Issue #6164, opened 2025-09-25
[kube-state-metrics] Probes fail with name instead of port numbers


lucasmlp:
### Describe the bug a clear and concise description of what the bug is.

Installing the kube-state-metrics Helm chart using version `6.3.0` or newer while setting `kubeRBACProxy.enabled=true` doesn't work. The `kube-state-metrics` container never gets ready.

### What's your helm version?

version.BuildInfo{Version:"v3.18.6", GitCommit:"b76a950f6835474e0906b96c9ec68a2eff3a6430", GitTreeState:"clean", GoVersion:"go1.24.6"}

### What's your kubectl version?

Client Version: v1.33.2 Kustomize Version: v5.6.0 Server Version: v1.33.4+k3s1

### Which chart?

kube-state-metrics

### What's the chart version?

6.3.0 and 6.4.0

### What happened?

I installed the chart setting `kubeRBACProxy.enabled=true`. The `kube-state-metrics` container never gets ready. I believe this problem was [pointed out](https://github.com/prometheus-community/helm-charts/pull/6098#issuecomment-3335529700) in the very [PR](https://github.com/prometheus-community/helm-charts/pull/6098) that caused it.

### What you expected to happen?

Both containers would get ready, and there'd be no issues with the probes.

### How to reproduce it?

The command used to install the chart was:

```sh
helm install ksm prometheus-community/kube-state-metrics --version 6.3.0 --set kubeRBACProxy.enabled=true
```

The issue can be checked like this:

```sh
$ kubectl get pods
NAME                                      READY   STATUS    RESTARTS   AGE
ksm-kube-state-metrics-7d98c5d4c9-fx4n7   1/2     Running   0          2m1s 
```

```sh
kubectl describe pod ksm-kube-state-metrics-7d98c5d4c9-fx4n7
Name:             ksm-kube-state-metrics-7d98c5d4c9-fx4n7
Namespace:        default
Priority:         0
Service Account:  ksm-kube-state-metrics
Node:             lima-rancher-desktop/192.168.5.15
Start Time:       Thu, 25 Sep 2025 19:22:31 -0300
Labels:           app.kubernetes.io/component=metrics
                  app.kubernetes.io/instance=ksm
                  app.kubernetes.io/managed-by=Helm
                  app.kubernetes.io/name=kube-state-metrics
                  app.kubernetes.io/part-of=kube-state-metrics
                  app.kubernetes.io/version=2.17.0
[...]

PR #6110, opened 2025-09-03
[kube-state-metrics] If kubeRBACProxy is enabled do not enable Probes

yellowhat:
#### What this PR does / why we need it

The [MR](https://github.com/prometheus-community/helm-charts/pull/6098) broke the deployment with `kubeRBACProxy.enabled=true`, as `*Probes`:

```yaml
livenessProbe:
  httpGet:
    ...
    port: http
...
readinessProbe:
  httpGet:
    port: metrics
....
```

reference a `port` name that it is not defined for that container.

As alternative we could define:

```diff
        - name: kube-state-metrics
          args:
            - --host=127.0.0.1
            - --port=9090
            - --resources=certificatesigningrequests,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,leases,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments
            - --metric-labels-allowlist=pods=[app,component,environment],deployments=[app,component,environment],daemonsets=[app,component,environment],jobs=[app,component,environment],services=[app,component,environment],statefulsets=[app,component,environment]
            - --metric-annotations-allowlist=pods=[alfaview.com/projectPath],deployments=[alfaview.com/projectPath],daemonsets=[alfaview.com/projectPath],jobs=[alfaview.com/projectPath],services=[alfaview.com/projectPath],statefulsets=[alfaview.com/projectPath]
            - --telemetry-host=127.0.0.1
            - --telemetry-port=9091
          imagePullPolicy: IfNotPresent
          image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.17.0
+         ports:
+           - containerPort: 8080
+             name: http
+           - containerPort: 8081
+             name: metrics
```

even if `kubeRBACProxy.enabled=true` but we would point to the ports of another container.

I would not do that:

* Due to warning of duplicated port:
    
    ```
    Warning: spec.template.spec.containers[1].ports[0]: duplicate port definition with spec.template.spec.containers[0].ports[0]
    Warning: spec.template.spec.containers[2].ports[0]: duplicate port definition with spec.template.spec.containers[0].ports[1]
    ```

* We are checking the same port 2 times (generating more request [I know they are small]) without getting anything
[...]

PR #5858, opened 2025-07-03
[kube-state-metrics] Fix probes when kubeRBACProxy is enabled

HaveFun83:
#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes # https://github.com/prometheus-community/helm-charts/issues/5529#issuecomment-2915267574

This change add .Values.kubeRBACProxy.ignoreProbePaths to allow the kube-state-metrics readinessProbe and livenessProbe work properly when kubeRBACProxy is enabled

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6302, opened 2025-11-06
[kube-state-metrics] Fix telemetryPort and Probes for KubeRBACProxy

Alja9:
- Allow users to custom `telemetryPort` for KubeRBACProxy when it is enabled.

Instead of using _hardcode 9091_ :
https://github.com/prometheus-community/helm-charts/blob/1a04bc9e7e988a210d0f1d1882243ff1e3bd1c03/charts/kube-state-metrics/templates/deployment.yaml#L64

- Fixing probes when `KubeRBACProxy` is `enabled`.

- When the kube-state-metrics using `hostNetwork: true`, it may experience a port conflict if port 9091 is already in use.
- Fixing probes when enabling `KubeRBACProxy` because it uses the port name while the container port is removed if `KubeRBACProxy` is `enabled` : https://github.com/prometheus-community/helm-charts/blob/df60c023ac067d7406e12e558640a6ebd481aca3/charts/kube-state-metrics/templates/deployment.yaml#L162-L170
 And use host `127.0.0.1` only for the `hostNetwork: true` condition, example like this is used in probes:
https://github.com/prometheus-community/helm-charts/blob/df60c023ac067d7406e12e558640a6ebd481aca3/charts/kube-state-metrics/templates/deployment.yaml#L175-L177

- fixes #

#### Checklist

- [x] Chart Version bumped


HaveFun83:
When i try release 6.4.1 with kube-rbac-proxy enabled i got the following errors
```
  Warning  Unhealthy  7s (x5 over 47s)  kubelet            Liveness probe errored and resulted in unknown state: strconv.Atoi: parsing "http": invalid syntax
  Warning  Unhealthy  2s (x7 over 43s)  kubelet            Readiness probe errored and resulted in unknown state: strconv.Atoi: parsing "metrics": invalid syntax
  ```
  
  @Alja9  Can you share your helm values?

**Comment by Alja9** [comment]
Hi @HaveFun83 
Yes, same with you, helm chart 6.4.1.
Since the `telemetryPort` is _hardcoded_, I use `telemetryPort` variable on `selfMonitor` section even though `selfMonitor` is disabled to avoid port conflicts.

But I also encountered like that so the problems on the probes, and I ignored the probes.
```
Warning  Unhealthy  4m7s (x13 over 5m49s)  kubelet            Readiness probe errored and resulted in unknown state: strconv.Atoi: parsing "metrics": invalid syntax
Warning  Unhealthy  52s (x31 over 5m52s)   kubelet            Liveness probe errored and resulted in unknown state: strconv.Atoi: parsing "http": invalid syntax
```
Okay, I'll try to fix the probes too.

And after I tried again with `selfMonitor` is enabled, it turned out that telemetry `selfMonitor` was affected by a conflict between container `kube-state-metrics` and `kube-rbac-proxy-telemetry`. 
```
E1107 13:51:43.051361       1 run.go:72] "command failed" err="failed to run groups: failed to listen on secure address: listen tcp :9123: bind: address already in use"
stream closed EOF for monitoring/ksm-2-kube-state-metrics-5bbd6c6dd7-zcl55 (kube-rbac-proxy-telemetry)
```
I'll try to fix that too.

HaveFun83:
Thanks a lot for looking into this. 
See also https://github.com/prometheus-community/helm-charts/pull/6110
IMHO. For long-term fix we should focus on the `auth-filter` option https://github.com/prometheus-community/helm-charts/pull/5858#issuecomment-3106318320

I did some basic testing with this option but need more time.

Alja9:
Yes, I'm looking forward to that feature, while I'm using `kube-rbac-proxy` .

I've added the newest commit, you can try it again and let me know what you think. 
Thank you.

And regarding probes in `kube-state-metrics` container, it should not use https because the only probes that need to use https are those in the `kube-rbac-proxy` container. That's what I think and you can see in `node-exporter` too.

sebastiangaiser (changes requested):
Thanks for tackling this, @Alja9. The core problem is the approach: probing KSM directly requires binding it to all interfaces (the `--host`/`--telemetry-host` changes), which exposes the metrics endpoints unauthenticated and defeats the kube-rbac-proxy, and readiness still fails in the default config (without `hostNetwork`, `--telemetry-port` isn't emitted, so telemetry listens on `8081` while the probe targets `9091`). Keeping KSM on localhost and routing the probes through the proxy avoids both; #6615 takes that route and I've reviewed it as the base for #6164 (and closed the duplicate #6600 in its favor). I'll close this once #6615 lands. You're very welcome to take it forward yourself, too, a PR that supersedes #6615 with the corrected approach would be more than welcome.

Alja9:
Thanks @sebastiangaiser for the detailed feedback. I understand the concern now, 
keeping KSM on localhost and routing probes through the proxy is the right approach.

Since the PR mentioned already takes the correct approach and is further along, I'll close this PR.

Thanks for your review, appreciate it!

PR #6318, opened 2025-11-08
[kube-prometheus-stack] Exclude containerd shm tmpfs mounts

dragoangel:
This PR will exclude collecting of `tmpfs on /run/containerd/io.containerd.grpc.v1.cri/sandboxes/<sandbox-id>/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)` mounted partitions in node-exporter from K8s nodes, which commonly seen when containerd runtime is used, f.e. in kubespray

N/A

N/A

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -31,7 +31,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 79.4.0
+version: 79.4.1
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.86.2
 kubeVersion: ">=1.25.0-0"


dragoangel:
> Is there a reason why this would need to be excluded by default? If not, you can override the extra args directly

Hi @QuentinBisson , is there any reason why someone will get profit from this shm tmpfs metrics? Maybe I missed something but I don't see use in them. Containerd will be common runtime over time and it will be more and more common, forcing every user do define from scratch own excludes without sane defaults not look like good practice to me.

dragoangel:
Before creating this PR I analyzed if this can be useful by myself and asked GPT, here is more clear answer:

That path is the shared memory (SHM) mount created per container sandbox by containerd (through the CRI interface).
* It’s typically mounted at /dev/shm inside the container, but from the host, it shows up under /run/containerd/.../shm.
* Size is fixed (commonly 64 MiB — size=65536k), ephemeral, and removed with the container.
* It’s purely in-memory storage for IPC, not persistent disk I/O.

Why it’s not useful
1. Per-container noise: Every container sandbox adds a new shm mount, so metric cardinality explodes (thousands of mountpoints).
2. Short-lived: The sandbox disappears when the pod/container dies, causing constant churn in Prometheus time series.
3. Static capacity: Size is always the same (64 MiB unless manually overridden).
4. Low operational value: SHM rarely becomes a bottleneck in typical workloads. If it does, container-level alerting on memory or process limits is more appropriate.

If you’re debugging specific IPC-heavy workloads that exhaust /dev/shm (e.g., Chrome, PostgreSQL, or ML frameworks using shared memory).
In such cases, targeted collection (inside that container or pod) — not cluster-wide scraping — makes sense.

Exclude /run/containerd/io.containerd.grpc.v1.cri/sandboxes/.../shm mounts from Prometheus scraping (e.g., use --collector.filesystem.ignored-mount-points).

Focus instead on:
1. node_memory_* metrics for overall memory pressure.
2. Container-level memory limits (container_memory_usage_bytes).

QuentinBisson:
Thank you for the detailled explanation. TBH, I am not sure why anyone would want to collect this but i'm sure a lot of people do it anyway. Having the reasons explained in the PR is better than nothing. Ideally it would bé in the comment but this is already good enough :)

Thanks

PR #6391, opened 2025-12-09
[prometheus-elasticsearch-exporter] Make liveness and readiness probes configurable

setycz:

When elasticsearch-es-elastic-user secret changes (created and managed by eck-operator) either because of password rotation or changes in deployment, prometheus-elasticsearch-exporter keeps using the old credentials and does not update values. One of the solutions is to update the liveness probe, e.g. to:

```
  livenessProbe:
    exec:
      command:
        - /bin/sh
        - -c
        - |
          #!/bin/sh
          # 1. Check if the exporter itself is alive
          if ! wget -qO- http://localhost:9108/healthz > /dev/null; then
            exit 1
          fi

          # 2. Check if the credentials are still valid by hitting ES directly
          AUTH=$(printf "%s:%s" "$ES_USERNAME" "$ES_PASSWORD" | base64 | tr -d '\n')
          ES_URI="http://elasticsearch-es-http.elastic-stack.svc:9200"
          if ! wget --spider --quiet --header "Authorization: Basic $AUTH" "$ES_URI"; then
             exit 1
          fi
```

But this is not possible, since the liveness probe is hardcoded. Thus, I would like to make liveness (and also for consistency rediness) probes configurable.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
+++ b/charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
@@ -145,26 +145,10 @@ spec:
           {{- if .Values.livenessProbe }}
           livenessProbe:
             {{- toYaml .Values.livenessProbe | nindent 12 }}
-          {{- else }}
-          livenessProbe:
-            httpGet:
-              path: /healthz
-              port: {{ .Values.service.metricsPort.name }}
-            initialDelaySeconds: 5
-            timeoutSeconds: 5
-            periodSeconds: 5
           {{- end }}
           {{- if .Values.readinessProbe }}
           readinessProbe:
             {{- toYaml .Values.readinessProbe | nindent 12 }}
-          {{- else }}
-          readinessProbe:
-            httpGet:
-              path: /healthz
-              port: {{ .Values.service.metricsPort.name }}
-            initialDelaySeconds: 1
-            timeoutSeconds: 5
-            periodSeconds: 5
           {{- end }}
           lifecycle:
             preStop:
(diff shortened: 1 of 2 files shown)


charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
```diff
@@ -142,20 +142,30 @@ spec:
           ports:
             - containerPort: {{ .Values.service.httpPort }}
               name: {{ .Values.service.metricsPort.name }}
+          {{- if .Values.livenessProbe }}
```
zeritti, line 145: If a user wants to customize a probe's field, they must define (uncomment and modify as per your change) all the required probe's fields, otherwise the deployment will fail. Such a customization should be more practical and transparent. The default probes' settings should be defined in `values.yaml` where the default values are really expected. Customizing the probes will then be much easier.

charts/prometheus-elasticsearch-exporter/values.yaml
```diff
@@ -79,6 +79,46 @@ resources: {}
   #   cpu: 100m
   #   memory: 128Mi
```
zeritti, line 81: As above, the default probes should be defined in this file.

zeritti (changes requested):
Thank you, @setycz, for your PR. Please, see my comments below.

charts/prometheus-elasticsearch-exporter/templates/deployment.yaml
setycz: Good point, I have included your feedback. Have a look please.

PR #6402, opened 2025-12-11
kube-state-metrics: fix probes when RBAC proxy is enabled

LaurentGoderre:

When installing the kube-state-metrics with RBAC proxy enabled, the pod never reaches a healthy state.

```
> helm install rbac oci://ghcr.io/prometheus-community/charts/kube-state-metrics --set kubeRBACProxy.enabled=true
> kubectl describe pod/rbac-kube-state-metrics-668c776c-7r4vg

[...]
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
[...]
  Warning  Unhealthy  9s (x2 over 9s)   kubelet            Readiness probe errored and resulted in unknown state: strconv.Atoi: parsing "metrics": invalid syntax
  Warning  Unhealthy  3s (x2 over 13s)  kubelet            Liveness probe errored and resulted in unknown state: strconv.Atoi: parsing "http": invalid syntax
```

This is because the named ports are not created when rbac proxy is enabled. Additionally, having the host set to `127.0.0.1` when rbac proxy is enabled breaks the health check because the check originates from Kubernetes (outside the container) and therefore the connection is refused.

With this fix, the probes work as expected in all scenarios.

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/templates/deployment.yaml
+++ b/charts/kube-state-metrics/templates/deployment.yaml
@@ -86,9 +86,6 @@ spec:
         {{-  if .Values.extraArgs  }}
         {{- .Values.extraArgs | toYaml | nindent 8 }}
         {{-  end  }}
-        {{- if .Values.kubeRBACProxy.enabled }}
-        - --host=127.0.0.1
-        {{- end }}
         - --port={{ $servicePort }}
         {{-  if .Values.collectors  }}
         - --resources={{ .Values.collectors | join "," }}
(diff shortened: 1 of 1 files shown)


sebastiangaiser (changes requested):
Thanks @LaurentGoderre, the repro (`strconv.Atoi: parsing "metrics"`) nails the named-port bug, and your port mapping is right: main for `/healthz`+`/livez`, telemetry for `/readyz`, keeping `--telemetry-port`.
The blocker is how it gets there: removing `--host`/`--telemetry-host=127.0.0.1` binds KSM to `::`, so `podIP:9090` and `podIP:9091` serve metrics unauthenticated and the kube-rbac-proxy is bypassed. (Also no version bump → CI is red, and there's no unittest.)

The fix should keep that mapping but route the probes through the proxies instead of exposing KSM directly. I've set up #6615 as the base for #6164 with a detailed review; I'll close this once #6615 merges. You're very welcome to take the fix forward yourself, though, a PR that supersedes #6615 with the corrected routing would be more than welcome.

Issue #48, opened 2020-09-06
[prometheus-blackbox-exporter] Allow configurable dnsPolicy and dnsConfig


jstewart612:
Subject is self-explanatory.

jstewart612:
Should be resolved once https://github.com/prometheus-community/helm-charts/pull/42 passes reviews and is merged.

monotek:
#42 merged. closing....

PR #42, opened 2020-09-05
[prometheus-blackbox-exporter] allow configurable DNS settings

jstewart612:
Signed-off-by: John Stewart <jstewart@rentpath.com>

PR #6439, opened 2025-12-28
[prometheus-node-exporter] Add dnsPolicy configuration option

sashanicolas:
## Description

Adds a `dnsPolicy` configuration option to the prometheus-node-exporter chart.

## Problem

When `hostNetwork: true` (the default), Kubernetes defaults to `dnsPolicy: ClusterFirst`, which prepends cluster DNS to the host's DNS servers. If the host has 3 nameservers configured, this results in 4 total nameservers, exceeding the limit and triggering:

```
Nameserver limits were exceeded, some nameservers have been omitted
```

## Solution

Add a `dnsPolicy` value (similar to the existing `dnsConfig`) allowing users to set:
- `Default` - use only host DNS (appropriate when pod doesn't need cluster DNS)
- `ClusterFirstWithHostNet` - prefer cluster DNS with host DNS fallback

## Changes

- `Chart.yaml`: version bump 4.49.2 → 4.50.0
- `values.yaml`: Added `dnsPolicy: ""` with documentation
- `templates/daemonset.yaml`: Added template logic to render `dnsPolicy` when set

## Usage

```yaml
prometheus-node-exporter:
  dnsPolicy: "Default"
```

## Related

- Similar to the fix in prometheus-blackbox-exporter (#48, #42)
- Ref: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

Signed-off-by: sashanicolas <sasha.nicolas@gmail.com>

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.49.2
+version: 4.50.0
 # renovate: github=prometheus/node_exporter
 appVersion: 1.10.2
 home: https://github.com/prometheus/node_exporter/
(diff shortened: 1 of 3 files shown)


zeritti:
@sashanicolas Please, sign your commits correctly so that the DCO check can pass

zanhsieh:
@zeritti Please review this PR. Thanks.

zeritti (approved):

PR #6442, opened 2025-12-29
[kube-state-metrics] Probes use ports by name rather than number

jcpunk:

The linter I'm running (popeyecli.io) reported that:
* the probes are using port numbers rather than names
* the probes are using a port that is not defined by the container

when installed not using the RBAC mode, but not self-monitor. This shifts to name based mappings.

https://popeyecli.io/

I view this as really a bugfix since the port isn't listed currently.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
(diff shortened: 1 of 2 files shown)


DrFaust92:
LGTM @prometheus-community/helm-charts-maintainers

jkroepke (approved):
LGTM

PR #6454, opened 2026-01-06
[kube-prometheus-stack] allow multiple networkpolicies

trouaux:

allow to configure multiple networkpolicies

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -483,6 +483,31 @@ alertmanager:
       ##
       configReloader: true
 
+  ## Configure multiple network policies for Prometheus (alternative to networkPolicy above)
+  ## When networkPolicies is set, the networkPolicy (singular) configuration above will be ignored
+  ## This allows creating multiple NetworkPolicy resources with different configurations
+  # networkPolicies:
+  #   - enabled: true
+  #     flavor: kubernetes
+  #     nameSuffix: egress-only  # Optional: custom suffix for policy name
+  #     namespace: monitoring
+  #     egress:
+  #       - to:
+  #         - namespaceSelector:
+  #             matchLabels:
+  #               name: kube-system
+  #     podSelector:
+  #       matchLabels:
+  #         app: prometheus
+  #   - enabled: true
+  #     flavor: kubernetes
+  #     nameSuffix: ingress-only  # Optional: custom suffix for policy name
+  #     ingress:
+  #       - from:
+  #         - namespaceSelector:
+  #             matchLabels:
+  #               name: monitoring
+
   ## Service account for Alertmanager to use.
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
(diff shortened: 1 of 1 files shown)


charts/kube-prometheus-stack/values.yaml
```diff
@@ -483,6 +483,31 @@ alertmanager:
       ##
       configReloader: true
 
+  ## Configure multiple network policies for prometheus (alternative to networkPolicy above)
```
viennaa:
```suggestion
  ## Configure multiple network policies for Prometheus (alternative to networkPolicy above)
```

GMartinez-Sisti:
Hi @trouaux, to get more context on this, any reason why you can't use the existing ingress and egress parameters and just add the requires entries there?

Also to make sure, you mention `Configure multiple network policies for Prometheus` but this entry is inside `.alertmanager`. Can you please clarify?

trouaux:
> Hi @trouaux, to get more context on this, any reason why you can't use the existing ingress and egress parameters and just add the requires entries there?
> 
> Also to make sure, you mention `Configure multiple network policies for Prometheus` but this entry is inside `.alertmanager`. Can you please clarify?

thanks for the review, indeed the example config is inside the wrong spec i'll fix this

regarding the use-case we are looking to create network policies in multiple namespaces for a federation setup

GMartinez-Sisti:
> > Hi @trouaux, to get more context on this, any reason why you can't use the existing ingress and egress parameters and just add the requires entries there?
> > Also to make sure, you mention `Configure multiple network policies for Prometheus` but this entry is inside `.alertmanager`. Can you please clarify?
> 
> thanks for the review, indeed the example config is inside the wrong spec i'll fix this
> 
> regarding the use-case we are looking to create network policies in multiple namespaces for a federation setup

Thanks for the clarification. Looks like this would cover your use case? https://github.com/prometheus-community/helm-charts/blob/71a97291d90cba0b126adef41744ff8805300425/charts/kube-prometheus-stack/values.yaml#L5401-L5404

trouaux:
> Thanks for the clarification. Looks like this would cover your use case?
> 
> https://github.com/prometheus-community/helm-charts/blob/71a97291d90cba0b126adef41744ff8805300425/charts/kube-prometheus-stack/values.yaml#L5401-L5404

true, we can go that route if adding this feature is not an option

the new spec feels like it could make things easier to maintain from user pov but i can understand not adding complexity to the template

GMartinez-Sisti:
> true, we can go that route if adding this feature is not an option
> 
> the new spec feels like it could make things easier to maintain from user pov but i can understand not adding complexity to the template

As this seems very specific to your use case, I would prefer not to add more to the already big values file 🙏 . We have `extraManifests` to allow for this type of customization and for what I've seen (my xp alone) most folks use a helm wrapper approach, so those extra manifests and logic would be part of the parent chart without the need to adjust this one.

Would like to hear from another maintainer if possible.

QuentinBisson:
@GMartinez-Sisti I agree as well. This chart is made to deploy 1 Prometheus and 1 Alertmanager. If user need to deploy multiple prometheus, I would advise to add a networkpolicy with the Prometheus CR

trouaux:
understood, thanks for reviewing @GMartinez-Sisti @QuentinBisson

PR #6461, opened 2026-01-07
[kube-state-metrics] Add exclude and extra collectors options

Alja9:
- Allows users to remove specific collectors from the default list.
- Allows users to append additional collectors.

With the above options, then users can avoids copy-paste of long default lists or without the need to redefine the entire list.

If we have different custom values, it is quite tiring to copy-paste the entire default list.

- fixes #

#### Checklist

- [x] Chart Version bumped


DrFaust92 (dismissed):

charts/kube-state-metrics/values.yaml
```diff
@@ -441,6 +441,14 @@ collectors:
   # - rolebindings
   # - serviceaccounts
 
+# Makes it easier to configure the default collector that is enabled
+# collectorExclude removes collectors from the default list
+collectorExclude: []
+  # - pods
+# collectorExtra always adds collectors after exclusion
+collectorExtra: []
+  # - rolebindings
```
```suggestion
# collectorExclude / collectorExtra tweak the default `collectors` list above
# without redefining it. Both operate on `.Values.collectors` (not on KSM's
# built-in defaults): with `collectors: []` there is nothing to layer onto and
# `collectorExtra` becomes the entire --resources list.
# collectorExclude removes entries from the list:
collectorExclude: []
  # - pods
# collectorExtra appends entries after exclusion (deduplicated):
collectorExtra: []
  # - rolebindings
```

sebastiangaiser (changes requested):
Nice, this is a clean way to avoid copy-pasting the full collector list. A few things before merge:
1. Could you add a helm-unittest suite covering the rendered `--resources` for the default, `collectorExclude`, `collectorExtra`, and the exclude+extra combination? That's exactly the behavior we want regression-protected.
2. Excluding the entire list (or excluding everything with no extra) renders `--resources=` empty. KSM skips empty entries when parsing, so an empty resource set falls back to all default resources (pkg/app/server.go: `len(opts.Resources) == 0` --> DefaultResources). So "exclude everything" silently re-enables everything, the opposite of the intent. Since KSM can't express an empty resource set via `--resources`, please guard the template to fail on an empty result instead of rendering an empty flag, e.g.:
```
{{- if not $final }}
{{- fail "collectorExclude removed all collectors; keep at least one" }}
{{- end }}
```
3. Naming: KSM itself has no resource exclude/denylist (only the `--resources` allowlist, which is this chart's collectors), so this is purely a chart-side list transform, that's fine. Could you rename `collectorExclude`/`collectorExtra` to `collectorsExclude`/`collectorsExtra` to match the existing collectors key? I'd avoid a Denylist name here on purpose, since we only use that where there's an actual upstream flag (`--metric-denylist`, `--namespaces-denylist`).

sebastiangaiser:
One more thing, and it's the important one I missed on the first pass: role.yaml derives its RBAC rules from `.Values.collectors` (each rule is gated on `{{ if has "<resource>" $.Values.collectors }})`, but this PR builds `--resources` from collectors - collectorExclude + collectorExtra. The two desync:
- `collectorExtra` adds a resource to `--resources` but grants no RBAC for it, so KSM can't watch it (e.g. `collectorExtra: [rolebindings]` when rolebindings isn't in collectors --> permission errors, no metrics).
- `collectorExclude` drops a resource from `--resources` but leaves its RBAC rule in place.

Alja9:
Hi @sebastiangaiser 
Thanks for catching those issues!

Will come back with updated commits. I'll let you know.

Alja9:
Hi @sebastiangaiser 

All points have been addressed and ready for review when you have time.
Thank you!

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -93,9 +93,24 @@ spec:
...
+        {{- $final = append $final . }}
+        {{- end }}
+        {{- end }}
+        {{- range $extra }}
+        {{- if not (has . $final) }}
+        {{- $final = append $final . }}
+        {{- end }}
+        {{- end }}
+        {{- if not $final }}
+        {{- fail "collectorExclude removed all collectors; keep at least one" }}
```
sebastiangaiser:
```suggestion
        {{- fail "collectorsExclude removed all collectors; keep at least one" }}
```

sebastiangaiser (changes requested):
Really nice work on this! The RBAC now tracks `--resources`, the empty-exclude guard, the rename, and the unittest are all in great shape.

One last thing and then it's good to go: the effective collector list is now computed identically in both `deployment.yaml` and `role.yaml`. Could you please extract that into a named template in `_helpers.tpl` and call it from both, so the two can't drift?

Alja9:
Ah Yes, it would be better if it was included in `_helpers.tpl`.

Alja9:
@sebastiangaiser All feedback addressed and refactored into a shared helper template. 
Ready for review again!

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -93,9 +93,11 @@ spec:
         - --host=127.0.0.1
         {{- end }}
         - --port={{ $servicePort }}
-        {{-  if .Values.collectors  }}
-        - --resources={{ .Values.collectors | join "," }}
-        {{-  end  }}
+        {{- $activeCollectors := include "kube-state-metrics.collectors" . | fromYamlArray }}
+        {{- if not $activeCollectors }}
+        {{- fail "collectorsExclude removed all collectors; keep at least one" }}
```
sebastiangaiser:
```suggestion
        {{- fail "collectors minus collectorsExclude plus collectorsExtra is empty; keep at least one collector" }}
```

sebastiangaiser (changes requested):

charts/kube-state-metrics/unittests/collectors_test.yaml
```diff
@@ -0,0 +1,211 @@
...
+          content: --resources=nodes
+
+  - it: should fail when collectors empty and no extra
+    template: deployment.yaml
+    set:
+      collectors: []
+      collectorsExclude: []
+    asserts:
+      - failedTemplate:
+          errorMessage: "collectorsExclude removed all collectors; keep at least one"
```
sebastiangaiser:
```suggestion
          errorMessage: "collectors minus collectorsExclude plus collectorsExtra is empty; keep at least one collector"
```

charts/kube-state-metrics/unittests/collectors_test.yaml
```diff
@@ -0,0 +1,211 @@
...
+        - resourcequotas
+        - secrets
+        - services
+        - statefulsets
+        - storageclasses
+        - validatingwebhookconfigurations
+        - volumeattachments
+    asserts:
+      - failedTemplate:
+          errorMessage: "collectorsExclude removed all collectors; keep at least one"
```
sebastiangaiser:
```suggestion
          errorMessage: "collectors minus collectorsExclude plus collectorsExtra is empty; keep at least one collector"
```

sebastiangaiser (changes requested):

sebastiangaiser (approved):
Thanks for the quick iterations on this! Everything's addressed now: the collector logic lives in a shared helper, the RBAC tracks `--resources`, and the tests cover the exclude/extra/empty cases. I applied a small wording tweak to the fail message and updated the two matching test assertions myself, hope that's ok. Approving and merging now, thanks for the contribution!

PR #6476, opened 2026-01-13
[kube-prometheus-stack] Make thanosService clusterIP optional

randrusiak:

It adds thanosService `clusterIP` only when the value is not empty. 
I'm using ArgoCD with **Server-Side-Apply**, and when I set `clusterIP: ""` Argo reports OutOfSync state because the live state differs from the desired state. 

The same pattern is used in other parts of this Helm chart.

#### Checklist

- [x] Chart Version bumped


jkroepke:
Thanks

jkroepke (approved):
LGTM

PR #6488, opened 2026-01-16
[kube-state-metrics] feat: add option to add aggregationRule

cwrau:

This PR adds the option to add an aggregation rule to the ClusterRole
used by kube-state-metrics.

That way the user can more easily plug in 3rd party configs, like
<https://github.com/kubernetes-sigs/cluster-api/blob/main/config/metrics/crd-clusterrole.yaml#L6>

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-state-metrics/values.yaml
+++ b/charts/kube-state-metrics/values.yaml
@@ -108,6 +108,17 @@ rbac:
   #   verbs: ["list", "watch"]
   extraRules: []
 
+  # Aggregate ClusterRoles using label selectors (only applies when useClusterRole: true).
+  # When set, this ClusterRole aggregates matching ClusterRoles instead of defining rules
+  # directly, so the built-in collector rules above are NOT included. Make sure the
+  # aggregated ClusterRoles grant the permissions kube-state-metrics needs.
+  # ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
+  # Example:
+  # clusterRoleSelectors:
+  #   - matchLabels:
+  #       rbac.example.com/aggregate-to-monitoring: "true"
+  clusterRoleSelectors: []
+
 # Enable kube-state-metrics native request authn/authz on the metrics endpoints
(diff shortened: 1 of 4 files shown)


charts/kube-state-metrics/values.yaml
```diff
@@ -99,6 +99,14 @@ rbac:
   #   verbs: ["list", "watch"]
   extraRules: []
 
+  # Aggregate ClusterRoles using label selectors (only applies when useClusterRole: true)
+  # ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
+  # Example:
+  # clusterRoleSelectors:
+  #   - matchLabels:
+  #       rbac.example.com/aggregate-to-monitoring: "true"
+  clusterRoleSelectors: []
```
```suggestion
  # Aggregate ClusterRoles using label selectors (only applies when useClusterRole: true).
  # When set, this ClusterRole aggregates matching ClusterRoles instead of defining rules
  # directly, so the built-in collector rules above are NOT included. Make sure the
  # aggregated ClusterRoles grant the permissions kube-state-metrics needs.
  # ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
  # Example:
  # - matchLabels:
  #     rbac.example.com/aggregate-to-monitoring: "true"
  clusterRoleSelectors: []
```

sebastiangaiser (changes requested):
Thanks for the PR! A few things:
- Worth making explicit for users: setting `clusterRoleSelectors` turns this into a purely aggregated ClusterRole, so its built-in collector `rules` are dropped and the aggregated ClusterRoles have to provide the permissions kube-state-metrics needs. I left an inline suggestion adding that to the value's comment.
- It needs a rebase on main, it's currently conflicting (base is 7.0.1, main is at 8.0.0), and the version bump will move to the next minor.
- Could you please add a helm-unittest covering both paths (selectors set, `aggregationRule` rendered with no rules; empty, rules rendered as before)?

sebastiangaiser (approved):

PR #6496, opened 2026-01-19
[kube-state-metrics] Change metricAnnotationsAllowlist (instead of AllowList) for consistency.

kossmac:

- fixes #

#### Checklist

- [x] Chart Version bumped


recollir:
We also discovered this inconsistency just a few minutes ago. I was about to create a PR and saw this one. Thanks, @kossmac

recollir:
I was wondering if and when this will be merged/released. As I want to raise a corresponding PR in the Grafana Alloy project for this. They include the KSM helm chart but do not yet expose the `metricAnnotationsAllowlist` value at all.

I was about to do this when I noticed that `metricAnnotationsAllowlist` is currently `metricAnnotationsAllowList`. And I would like to wait for this fix here to be available and released. (cc @dotdc, @mrueg, @tariq1890)

sebastiangaiser:
Thanks for tackling the naming inconsistency! I decided to go non-breaking here: instead of renaming outright, we'll add `metricAnnotationsAllowlist` as the preferred key while keeping `metricAnnotationsAllowList` as a deprecated alias (so existing users don't break), then drop the old key in a future major release. That's consolidated on #6617, so I'll close this in its favor, thanks again.

PR #6515, opened 2026-01-25
kube-state-metrics: fix fullname nameOverride duplication

### LINKED ISSUE #6510: [kube-state-metrics] nameOverride doubles release name in objects names

When using `nameOverride` value that contains release name, and has something else, k8s objects have release name doubled in their names. For release named `ksm-non-pods` and value `nameOverride: "ksm-non-pods-foo"` it results in `name: ksm-non-pods-ksm-non-pods-foo`

### What's your helm version?

`version.BuildInfo{Version:"v3.19.5", GitCommit:"4a19a5b6fb912c5c28a779e73f2e0880d9e239a4", GitTreeState:"clean", GoVersion:"go1.24.11"}`

None, don't have it installed

### Which chart?

kube-state-metrics

### What's the chart version?

I've hit it with 5.19.0 and 6.1.4. Judging by the blame it's present since 2.13.2 to this day

### What happened?

Documentation says this:

> If release name contains chart name it will be used as a full name.

I have a release named `ksm-non-pods`, and when I use `nameOverride: "ksm-non-pods"` in values it works fine, e.g. objects have `name: ksm-non-pods`
But if I use `nameOverride: "ksm-non-pods-foo"` it breaks, and appends extra release name, resulting in `name: ksm-non-pods-ksm-non-pods-foo`

Issue is that emplate helper `kube-state-metrics.fullname` is using `if contains $name .Release.Name`

https://github.com/prometheus-community/helm-charts/blob/17c59d5251f0ee22927521443ef3e6ec944a1a79/charts/kube-state-metrics/templates/_helpers.tpl#L19

And, while proper arguments documentation is clearly lacking, it seems that first argument is needle, and second it a haystack
https://helm.sh/docs/chart_template_guide/function_list/#contains

IIUC, Helm uses `sprig` as template engine, and `sprig` documentation is no better
https://masterminds.github.io/sprig/strings.html

But it's clear from source, that it changes order of arguments from `strings.Contains`:
https://github.com/Masterminds/sprig/blob/989da45d7c082c6ec85cf95f59e23b461cdafe03/functions.go#L152

_No response_

### How to reproduce it?

_No response_

_No response_

`helm template ksm-non-pods ./ksm --set kube-state-metrics.nameOverride=ksm-non-pods-foo`

_No response_

firasmosbehi:
Fixes #6510

- prevent release name duplication when nameOverride already includes it
- add fullname helm-unittests
- bump chart to 7.1.1

Tests:
- helm unittest --strict --file 'unittests/**/*.yaml' charts/kube-state-metrics
- helm upgrade --install ksm-non-pods charts/kube-state-metrics --set nameOverride=ksm-non-pods-foo --wait
- kubectl get svc -n default ksm-non-pods-foo

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 7.1.0
+version: 7.1.1
 # renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.18.0
 home: https://github.com/kubernetes/kube-state-metrics/
--- a/charts/kube-state-metrics/templates/_helpers.tpl
+++ b/charts/kube-state-metrics/templates/_helpers.tpl
@@ -16,10 +16,13 @@ If release name contains chart name it will be used as a full name.
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
-{{- if contains $name .Release.Name -}}
-{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- $releaseName := .Release.Name -}}
+{{- if contains $name $releaseName -}}
+{{- $releaseName | trunc 63 | trimSuffix "-" -}}
+{{- else if contains $releaseName $name -}}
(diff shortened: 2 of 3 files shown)


firasmosbehi:
Superseded by PR #6557 which rebases the fullname fix onto current main and adds tests. Closing to avoid duplication.

PR #3081, opened 2023-02-28
[prometheus] Remove use of API proxy

SuperQ:
#### What this PR does / why we need it

Update the defaults for the `prometheus` chart to not use the API proxy to access kubelet and cadvisor metrics. This is not necessary with the ServiceAccount token. This matches the scrape configuration of
`kube-prometheus-stack`.

#### Which issue this PR fixes

- fixes https://github.com/prometheus-community/helm-charts/issues/3079

#### Checklist

- [X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6533, opened 2026-01-27
[prometheus] Remove permissions to nodes/proxy from cluster role

### LINKED ISSUE #3079: [prometheus] Remove use of API proxy

The `prometheus` chart requires ClusterRole `nodes/proxy` permission due to the default values using the Kubernetes API proxy to access some metrics. This elevated permission has some potential for privilege escalation in the cluster.

Using the proxy shouldn't be necessary as there is a ServiceAccount token. This is how configuration of the scrape configs is done in the `kube-prometheus-stack` chart. I'm guessing these changes were not backported to the original `prometheus` chart.

Remove `/api/v1/nodes/$1/proxy/metrics` relabeling for the `metrics_path` so that we no longer need the permission `nodes/proxy`.

NONE

### Additional context.

_No response_

zeritti:

At the moment, permissions to _nodes/proxy_ are being granted through a cluster role. This PR removes these permissions as they are not required for Prometheus to access kubelets' metrics and have potentially serious security implications:
- permissions to nodes/proxy removed from the cluster role
- scrape configs adjusted

- fixes #3079

The original PR #3081 is closed without having been merged.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,7 +111,7 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: super-linter/super-linter/slim@d5b0a2ab116623730dd094f15ddc1b6b25bf7b99 # v8.3.2
+        uses: super-linter/super-linter/slim@12562e48d7059cf666c43a4ecb0d3b5a2b31bd9e # v8.4.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           LINTER_RULES_PATH: .github/linters
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -51,14 +51,17 @@ jobs:
           echo "default-cache-ttl 86400" >> "$HOME/.gnupg/gpg-agent.conf"
           gpgconf --kill gpg-agent
           gpgconf --launch gpg-agent
-          echo "$GPG_PASSPHRASE" | gpg --batch --yes --passphrase-fd 0 --import "$GPG_KEY_PATH"
-          
+          gpg --batch --yes --passphrase-fd 0 --import "$GPG_KEY_PATH" <<< "$GPG_PASSPHRASE"
+
+          mkdir "${{ runner.temp }}/.gnupg/"
           gpg --batch --yes --export "$GPG_KEY_ID" >~/.gnupg/pubring.gpg
-          echo "$GPG_PASSPHRASE" | gpg --batch --yes --passphrase-fd 0 --export-secret-keys "$GPG_KEY_ID" >~/.gnupg/secring.gpg
+          gpg --batch --yes --passphrase-fd 0 --export-secret-keys "$GPG_KEY_ID" >~/.gnupg/secring.gpg <<< "$GPG_PASSPHRASE"
+          cat > $GPG_PASSPHRASE_FILE <<< "$GPG_PASSPHRASE"
         env:
           GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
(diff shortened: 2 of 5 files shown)


zeritti:
Please, review @zanhsieh

zanhsieh (approved):

SuperQ:
> scrape configs adjusted

This is a what, but there is no why. Why are the relabel configs adjusted?

SuperQ:
Also, the other PR was closed because it _didn't work_. @zanhsieh Did you test this at all? What verification did you do?

zeritti:
> > scrape configs adjusted
> 
> This is a what, but there is no why. Why are the relabel configs adjusted?

@SuperQ Without prometheus having the permission on nodes/proxy, the modified labels `__address__` and `__metrics_path__` were not valid any longer as a scrape would result in a permission denied error. On the other hand, the metrics path to the cadvisor metrics had to be specified as that was not a default path. This is why those relabeling rules have been adjusted.

The result is prometheus' attempting to scrape kubelets directly - discovered by means of the _node_ role and accessed with a service account token (mounted by default in this case and set in the configuration) and the configured CA certificate. This is similar to what kube-prometheus-stack does, although kubelets gets discovered by means of the _endpoints_ role there.

I validated the change when preparing the PR with k3s and kind (1.33). In both cases, prometheus scraped kubelets successfully. I could not reproduce the errors that were reported previously on the same change.

The previous labelsets of the nodes' time series remain unchanged and are kept. The new configuration adds target label `metrics_path` at cadvisor that has not been present before (the relabeling rule added explicitely, not really needed as the path is known to prometheus from the configuration). The metrics endpoints are now different: https\://NODE:PORT/metrics vs. previous https:\//kubernetes.default.svc/api/v1/nodes/NODE/proxy/metrics/...

nrlulz:
I think this broke node and cadvisor metrics in my talos cluster. I get a tls validation error when trying to access the metrics endpoints directly. Is this expected?

```
Error scraping target: Get "https://192.168.1.32:10250/metrics": tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.32 because it doesn't contain any IP SANs

Error scraping target: Get "https://192.168.1.32:10250/metrics/cadvisor": tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.32 because it doesn't contain any IP SANs
```

I can see if it behaves the same way on EKS if that would help. I don't know if I really want to be messing with cert SANs, or if it's something that usually just works by default and this is just a talos thing?

edit: no issue on EKS, guess it's just a talos thing.

zeritti:
> I think this broke node and cadvisor metrics in my talos cluster. I get a tls validation error when trying to access the metrics endpoints directly. Is this expected?
> 
> ```
> Error scraping target: Get "https://192.168.1.32:10250/metrics": tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.32 because it doesn't contain any IP SANs
> 
> Error scraping target: Get "https://192.168.1.32:10250/metrics/cadvisor": tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.32 because it doesn't contain any IP SANs
> ```
> 
> I can see if it behaves the same way on EKS if that would help. I don't know if I really want to be messing with cert SANs, or if it's something that usually just works by default and this is just a talos thing?

Thank you, @nrlulz. This can happen, yes. The prometheus chart does not relax certificate validation by default, i.e. it is not skipping it which leads to the error. In my view, there is one practical solution only and that is modifying the corresponding scrape configs in this way:

```yaml
scrapeConfigs:
  kubernetes-nodes:
    tls_config:
      insecure_skip_verify: true
  kubernetes-nodes-cadvisor:
    tls_config:
      insecure_skip_verify: true
```

I'd suggest to make that change default. Otherwise, we would not be able to remove the nodes/proxy permission.

PR #6572, opened 2026-01-31
[kube-state-metrics] add chart icon

firasmosbehi:
## What
- add Prometheus icon to kube-state-metrics chart metadata
- bump chart version to 7.1.1

## Why
- chart missing icon; adding aligns metadata and removes lint warning

## Testing
- helm lint charts/kube-state-metrics


charts/kube-state-metrics/Chart.yaml
```diff
@@ -7,10 +7,11 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 7.1.0
+version: 7.1.1
 # renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.18.0
 home: https://github.com/kubernetes/kube-state-metrics/
+icon: https://raw.githubusercontent.com/cncf/artwork/master/prometheus/icon/color/prometheus-icon-color.svg
```
sebastiangaiser, line 14:
```suggestion
icon: https://raw.githubusercontent.com/cncf/artwork/main/projects/kubernetes/icon/color/kubernetes-icon-color.svg
```

Thanks for adding this. `helm lint` does recommend an icon, so it's a welcome addition.
One change before it goes in: the URL points at the generic Prometheus icon, but kube-state-metrics is a Kubernetes SIG project, not Prometheus, so that logo would be misleading. kube-state-metrics doesn't ship its own logo, so the accurate choice is the Kubernetes icon from the same CNCF artwork source.

sebastiangaiser (changes requested):

sebastiangaiser (approved):

PR #6577, opened 2026-01-31
[prometheus-nginx-exporter] Add option priorityClassName

Alja9:
Add option priorityClassName for the Nginx Exporter Pod

Example, In my case, nginx components are prioritized including its pod exporter. But in this chart still does not provide that option.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-nginx-exporter/values.yaml
+++ b/charts/prometheus-nginx-exporter/values.yaml
@@ -133,6 +133,9 @@ extraVolumeMounts: []
 nginx:
   enabled: false
 
+## Priority ClassName for the pod
+# priorityClassName: ""
+
 # Configure the deployment strategy. The default is "RollingUpdate".
 # ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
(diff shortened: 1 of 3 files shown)


charts/prometheus-nginx-exporter/values.yaml
```diff
@@ -133,6 +133,9 @@ extraVolumeMounts: []
 nginx:
   enabled: false
 
+## Priority ClassName for the pod
+# priorityClassName: ""
```
zeritti: Please, do not comment the field out:

```suggestion
priorityClassName: ""
```

zeritti (changes requested):
Thank you, @Alja9, for your PR. Please, see my comments below.

charts/prometheus-nginx-exporter/values.yaml
Alja9: Thank you for taking a look @zeritti 
Done, revised.

zeritti (approved):

PR #6578, opened 2026-02-01
[prometheus] Relax certificate validation with kubelets

zeritti:

PR #6533 introduced a removal of the permission on nodes/proxy so that prometheus could scrape kubelets directly without using the proxy. However, where the certificates at the kubelets do not include their IP address amongst SANs, the IP address being by default the scrape target, prometheus raises an error and does not scrape the target. 

This PR enables skipping certificate validation in the scrape configs for kubelets so that the permission on nodes/proxy can be removed and at the same time the default scrape configs can work out of the box in a cluster-wide deployment.

Prometheus' scraping kubelets directly and successfully with the kube-prometheus-stack chart and its default configuration is possible only with the same change.
 
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,20 @@ You can then run `helm search repo prometheus-community` to see the charts.
 
 OCI artifacts of all Prometheus Helm charts are available in [ghcr.io](https://github.com/orgs/prometheus-community/packages?repo_name=helm-charts).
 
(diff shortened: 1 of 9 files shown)


zeritti:
Please, review @zanhsieh

PR #6584, opened 2026-02-02
chore(prometheus-adapter): add nameOverride and fullnameOverride values

yuanchen8911:
## Summary
- Add `nameOverride` and `fullnameOverride` values to `prometheus-adapter` chart
- The `_helpers.tpl` template already supports these values but they were not exposed in `values.yaml`
- Bump chart version from 5.2.0 to 5.3.0

## Motivation
Users need to override deployment names when deploying multiple instances or integrating with existing naming conventions. Without these values exposed, customizing resource names was not possible through standard Helm value overrides.

## Changes
- `charts/prometheus-adapter/values.yaml`: Added `nameOverride` and `fullnameOverride` with empty string defaults
- `charts/prometheus-adapter/Chart.yaml`: Bumped version to 5.3.0

Signed-off-by: Yuan Chen <yuanchen97@gmail.com>

### DIFF
--- a/charts/prometheus-adapter/Chart.yaml
+++ b/charts/prometheus-adapter/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
-version: 5.2.1
+version: 5.3.0
 appVersion: v0.12.0
 description: A Helm chart for k8s prometheus adapter
 home: https://github.com/kubernetes-sigs/prometheus-adapter
--- a/charts/prometheus-adapter/values.yaml
+++ b/charts/prometheus-adapter/values.yaml
@@ -2,6 +2,12 @@ affinity: {}
 
 topologySpreadConstraints: []
 
+# Override the name of the chart
+nameOverride: ""
+
+# Override the full name of the release
+fullnameOverride: ""
+
 image:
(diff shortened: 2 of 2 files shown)


yuanchen8911:
> Looks good, just need to fix merge conflict caused by merging another PR.

Resolved, PTAL. Thanks.

charts/prometheus-adapter/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
+
 version: 5.2.1
```
steven-sheehy: Need to bump version to 5.3.0.

charts/prometheus-adapter/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: v1
 name: prometheus-adapter
+
```
steven-sheehy: nit: remove whitespace change

yuanchen8911:
Addressed both comments. PTAL

  1. Removed the blank line between name: and version: in Chart.yaml
  2. Bumped version to 5.3.0

  Also rebased onto upstream main to resolve the merge conflict.

PR #6589, opened 2026-02-03
[prometheus-node-exporter] Add rapl initContainer

henrichter:

This is a WIP proposal to see if there's interest in baking a [permission fix for the powercap collector](https://github.com/prometheus/node_exporter/issues/1892) into the chart.

Currently, node-exporter can't scrape RAPL metrics on most distros because `/sys/devices/virtual/powercap/*/energy_uj` is `root:root 0400`. This PR adds an initContainer to chown these files to the exporter's GID and chmod 640 them so the collector actually works without running the whole daemon as root.

- Waiting for feedback before bumping Chart version
- If interested I'm happy to wrap this in an `enabled` flag so it's strictly opt-in.
- I targeted the `nobody` group as the owner. Open to better ideas if you want to tighten this further.
- Logic based on [scaphandre's init script](https://github.com/hubblo-org/scaphandre/blob/main/init.sh)

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/prometheus-redis-exporter/Chart.yaml
+++ b/charts/prometheus-redis-exporter/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
 # renovate: github=oliver006/redis_exporter
-appVersion: v1.80.2
+appVersion: v1.81.0
 description: Prometheus exporter for Redis metrics
 name: prometheus-redis-exporter
-version: 6.20.2
+version: 6.21.0
 home: https://github.com/oliver006/redis_exporter
 sources:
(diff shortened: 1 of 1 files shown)


henrichter:
I've seen something similar being used by bitnami postgresql chart [1]. They use a `volumePermissions.enabled` flag. Maybe we can do somthing similar, such that other Permissionfixes can be added/toggled later?

[1] https://github.com/bitnami/charts/blob/8caeb0c818e72f9b7f03e1b01a24ce1ea1886e6f/bitnami/postgresql/values.yaml#L1591-L1594

PR #6624, opened 2026-02-09
[prometheus-pushgateway] Allow existing secret for web config

### LINKED ISSUE #5979: [prometheus-pushgateway] Add an ability to configure basicAuthUsers via secrets

Now it's impossible to configure basicAuthUsers via existing secret.

I'd like to have this ability.

-

### Additional context.

_No response_

firasmosbehi:
## What this PR does
- Adds `webConfiguration.existingSecret.name` support so users can provide an existing web config secret.
- Keeps generated secret behavior for inline `webConfiguration.basicAuthUsers`.
- Uses generated secret for ServiceMonitor basic auth only when inline users are provided.
- Adds CI values to validate the new existing-secret path.

Fixes #5979.

## Testing
- helm lint charts/prometheus-pushgateway
- GITHUB_SHA=$(git rev-parse HEAD) ct lint --config .github/linters/ct.yaml --charts charts/prometheus-pushgateway

### DIFF
--- a/charts/prometheus-pushgateway/values.yaml
+++ b/charts/prometheus-pushgateway/values.yaml
@@ -123,6 +123,8 @@ resources: {}
 webConfiguration: {}
   # basicAuthUsers:
   #   username: password
+  # existingSecret:
+  #   name: pushgateway-web-config
 
 liveness:
   enabled: true
--- a/charts/prometheus-pushgateway/Chart.yaml
+++ b/charts/prometheus-pushgateway/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 appVersion: v1.11.2
 description: A Helm chart for prometheus pushgateway
 name: prometheus-pushgateway
-version: 3.6.0
+version: 3.6.1
 home: https://github.com/prometheus/pushgateway
 sources:
   - https://github.com/prometheus/pushgateway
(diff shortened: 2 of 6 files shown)


charts/prometheus-pushgateway/values.yaml
```diff
@@ -123,6 +123,8 @@ resources: {}
 webConfiguration: {}
   # basicAuthUsers:
   #   username: password
+  # existingSecret:
```
zeritti, line 126: The user needs to know that the secret's key must be `web-config.yaml` as this name is expected in the corresponding Pushgateway's flag.

The feature itself is useful. However, setting `existingSecret` will lead to Pushgateway never becoming ready being repeatedly restarted. This is because probes have no knowledge of the basic authentication configuration as long as it is present in the web-config file (would receive 401). The user would have to know the configuration and enter it in the probes' headers handling the values properly as secrets. I think we should not require the user to do that as this would go against the advantage of using an existing secret.

zeritti (changes requested):
Thak you, @firasmosbehi, for your PR. Please, see my comments below.

charts/prometheus-pushgateway/values.yaml
rmntrvn: @firasmosbehi Hello! Are you planning to make the changes mentioned?

zeritti: To prevent an exposure of the credentials in probes' configuration, we could switch from the _httpGet_ probe to the _tcpSocket_ probe if an existing secret is set. I am not aware of any negative impact of such a change in this case. The only thing that might not be "nice" would probably be logging TLS handshake errors at Pushgateway with TLS set up.

zeritti:
Completed via #7044.

PR #6629, opened 2026-02-10
[kube-prometheus-stack] Fix admission webhook DNS name rendering

vitrix1:
To prevent Error: YAML parse error on charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml: error converting YAML to JSON: yaml: line 45: did not find expected key

**What this PR does / why we need it**
Fixes broken DNS SAN in admission webhook cert caused by whitespace in Helm template.

**Special notes for your reviewer**
Without this change, the rendered DNS name contains a leading space, causing invalid SANs in the webhook TLS certificate.

cc @andrewgkew @gianrubio @gkarthiks @GMartinez-Sisti @jkroepke @Xtigyro @QuentinBisson

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6646, opened 2026-02-13
[kube-prometheus-stack] fix admission webhook DNS name rendering

mjnagel:

This PR reverts https://github.com/prometheus-community/helm-charts/pull/6629 which incorrectly introduced a [chomp](https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace) which results in errors at deploy time:
```console
x509: certificate is valid for kube-prometheus-stack-operatorkube-prometheus-stack-operator.monitoring.svc, not kube-prometheus-stack-operator
```

Note the way that kube-prometheus-stack-operatorkube-prometheus-stack-operator.monitoring.svc is smushed together rather than two separate names (`kube-prometheus-stack-operator` and `kube-prometheus-stack-operator.monitoring.svc`).

Haven't opened an issue, but I can.

The prior PR introduced this issue to resolve a bug, but the problem in that case was related to Windows `^M` (see https://unix.stackexchange.com/questions/32001/what-is-m-and-how-do-i-get-rid-of-it) - cc @vitrix1 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 81.6.7
+version: 81.6.8
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.88.1
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/templates/_helpers.tpl
+++ b/charts/kube-prometheus-stack/templates/_helpers.tpl
@@ -334,7 +334,7 @@ global:
 {{- $fullname := include "kube-prometheus-stack.operator.fullname" . }}
 {{- $namespace := include "kube-prometheus-stack.namespace" . }}
 {{- $fullname }}
-{{- $fullname }}.{{ $namespace }}.svc
+{{ $fullname }}.{{ $namespace }}.svc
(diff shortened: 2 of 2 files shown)


mjnagel:
By way of additional explanation, the [template here](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch/job-createSecret.yaml#L47) REQUIRES these cert names to be separated by a newline in order to properly be split/comma-separated. The previous PR was "resolved" by this change only because the `-` (chomp) character trimmed out the Windows carriage return, but would result in other issues due to the DNS names not properly being separated.

I believe the prior issue came down to this:
> It is worth adding that I build the archive with the chart on Windows 11

@vitrix1 you likely need to switch your editor? from using carriage returns and make sure that it only has newline characters (reference [this link](https://unix.stackexchange.com/questions/32001/what-is-m-and-how-do-i-get-rid-of-it) as one helper for that). It might look different based on your editor/process though.

jkroepke:
I guess a bitter fix would be a comma separation, instead a newline spearation?

mjnagel:
> I guess a bitter fix would be a comma separation, instead a newline spearation?

I could make that change if it's the desired path - I think `,` would be a safe character to use here and split on. I'd still probably argue that the previous issue in https://github.com/prometheus-community/helm-charts/pull/6629 is down to newlines being handled incorrectly when the chart archive was built, and could affect other templates beyond just this. Splitting on newlines isn't inherently wrong, switching to commas would just remove this one spot from being a potential problem with newlines.

jkroepke:
It just seems serios cross os issues with \r\n on windows and \n on linux

PR #6651, opened 2026-02-16
[prometheus-fastly-exporter] support repeatable flags for options

zono0013:
This PR adds support for repeatable flags in the `options` configuration for prometheus-fastly-exporter.

Some Fastly exporter flags (like `--service`) can be specified multiple times. Previously, users had to encode the entire argument (including the flag name and  value) as the option key itself, which was not intuitive. This PR introduces a more straightforward way to specify multiple values using arrays.

Before
```
option:
  service=example-service-id-1 : ""
  service=example-service-id-2 : ""
```

After(This PR)
```
option:
  service:                                                                                                                                                      
    - example-service-id-1
    - example-service-id-2
```

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-fastly-exporter/templates/deployment.yaml
+++ b/charts/prometheus-fastly-exporter/templates/deployment.yaml
@@ -34,7 +34,7 @@ spec:
             {{- range $key, $value := .Values.options }}
             {{- if kindIs "slice" $value }}
             {{- range $v := $value }}
-            - "-{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
+            - "-{{ $key }}{{ if $v }}={{ $v }}{{ end }}"
             {{- end }}
             {{- else }}
             - "-{{ $key }}{{ if $value }}={{ $value }}{{ end }}"


charts/prometheus-fastly-exporter/values.yaml
```diff
@@ -31,10 +31,14 @@ nameOverride: ""
 fullnameOverride: ""
 
 # Flags - for a list visit https://github.com/fastly/fastly-exporter#filtering-services
+# Repeatable flags accept a list to generate multiple args with the same key.
 options:
   {}
   # metric-allowlist='bytes_total$'
   # metric-blocklist=imgopto
+  # service:
```
arslanbekov:
```suggestion
  # service:
```

charts/prometheus-fastly-exporter/templates/deployment.yaml
```diff
@@ -32,8 +32,14 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             {{- range $key, $value := .Values.options }}
+            {{- if kindIs "slice" $value }}
+            {{- range $v := $value }}
+            - "-{{ $key }}={{ $v }}"
```
arslanbekov: The slice branch always produces `-key=value`, but the scalar branch handles empty values differently (`-key` without `=`)

For consistency, consider using the same pattern:
```
- "-{{ $key }}{{ if $v }}={{ $v }}{{ end }}"
```
This way, both branches behave identically for edge cases

zono0013: Thanks for the review! Both have been fixed.

charts/prometheus-fastly-exporter/templates/deployment.yaml
```diff
@@ -32,8 +32,14 @@ spec:
...
           args:
             {{- range $key, $value := .Values.options }}
+            {{- if kindIs "slice" $value }}
+            {{- range $v := $value }}
             - "-{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
             {{- end }}
+            {{- else }}
+            - "-{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
+            {{- end }}
+            {{- end }}
```
arslanbekov, line 42:
```suggestion
          args:
            {{- range $key, $value := .Values.options }}
            {{- if kindIs "slice" $value }}
            {{- range $v := $value }}
            - "-{{ $key }}{{ if $v }}={{ $v }}{{ end }}"
            {{- end }}
            {{- else }}
            - "-{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
            {{- end }}
            {{- end }}
```

arslanbekov:
Thanks @zono0013!

zono0013:
Thanks @arslanbekov  for the review and help! 🙏

PR #6664, opened 2026-02-19
[prometheus-node-exporter] Add optional initContainer for permission fixes

henrichter:

This PR adds an optional `permissionInitContainers` block to fix restricted host-path permissions for collectors like `rapl` and `slabinfo`. It works around default kernel permissions by changing the host file group to the node-exporters `runAsGroup` and making them group-readable. This logic is disabled by default.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 82.1.1
+version: 82.4.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.89.0
 kubeVersion: ">=1.25.0-0"
@@ -60,7 +60,7 @@ dependencies:
     repository: https://prometheus-community.github.io/helm-charts
     condition: nodeExporter.enabled
   - name: grafana
-    version: "11.1.7"
+    version: "11.2.0"
     repository: https://grafana-community.github.io/helm-charts
     condition: grafana.enabled
   - name: prometheus-windows-exporter
--- a/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
+++ b/charts/kube-prometheus-stack/hack/sync_grafana_dashboards.py
@@ -29,11 +29,11 @@ def change_style(style, representer):
 
 refs = {
     # renovate: git-refs=https://github.com/prometheus-operator/kube-prometheus branch=main
-    'ref.kube-prometheus': 'b334a3d5fd94408a4cead3fe557ff372fbd00616',
+    'ref.kube-prometheus': 'd4e438b5a9de67a9f965eaa94b90c80ad6260bff',
     # renovate: git-refs=https://github.com/kubernetes-monitoring/kubernetes-mixin branch=master
-    'ref.kubernetes-mixin': 'b3d555f3cd938b1e15fefb91d3841a12fc096ed4',
+    'ref.kubernetes-mixin': 'ae37ded0a961882e86543d652f0576a0f803ea10',
     # renovate: git-refs=https://github.com/etcd-io/etcd branch=main
-    'ref.etcd': '23085b4d275826d93a4c5da6c3563321369cf865',
+    'ref.etcd': '4125079c138889e20f66c30a34317baee96b6500',
 }
 
 # Source files list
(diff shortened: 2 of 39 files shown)


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -38,9 +38,54 @@ spec:
...
+        {{- if .Values.extraInitContainers }}
+        {{- toYaml .Values.extraInitContainers | nindent 8 }}
+        {{- end }}
+        {{- if $fixesEnabled }}
+        - name: permission-fix
+          {{- if .Values.permissionInitContainers.image.sha }}
+          image: "{{ .Values.global.imageRegistry | default .Values.permissionInitContainers.image.registry}}/{{ .Values.permissionInitContainers.image.repository }}:{{ .Values.permissionInitContainers.image.tag }}@sha256:{{ .Values.permissionInitContainers.image.sha }}"
+          {{- else }}
+          image: "{{ .Values.global.imageRegistry | default .Values.permissionInitContainers.image.registry}}/{{ .Values.permissionInitContainers.image.repository }}:{{ .Values.permissionInitContainers.image.tag }}"
+          {{- end }}
```
zeritti: A suggestion:

```suggestion
          {{- with .Values.permissionInitContainers.image }}
          {{- if .sha }}
          image: "{{ $.Values.global.imageRegistry | default .registry}}/{{ .repository }}:{{ .tag }}@sha256:{{ .sha }}"
          {{- else }}
          image: "{{ $.Values.global.imageRegistry | default .registry}}/{{ .repository }}:{{ .tag }}"
          {{- end }}
          {{- end }}
```

charts/prometheus-node-exporter/values.yaml
```diff
@@ -543,6 +543,23 @@ sidecarHostVolumeMounts: []
...
 
+permissionInitContainers:
+  image:
+    registry: quay.io
+    repository: prometheus/busybox
+    tag: latest
+    sha: ""
+    pullPolicy: IfNotPresent
+  securityContext:
+    privileged: true
```
zeritti: Do the containers have to run privileged? Please, confirm.

charts/prometheus-node-exporter/values.yaml
```diff
@@ -543,6 +543,23 @@ sidecarHostVolumeMounts: []
 ##
 extraInitContainers: []
```
zeritti, line 545: Please, add a short comment on the init containers and their purpose. It may be worth mentioning that the rapl collector is enabled by default whilst the slabinfo collector is not.

charts/prometheus-node-exporter/values.yaml
henrichter: It actually works without running it privileged. I removed it :D

charts/prometheus-node-exporter/values.yaml
henrichter: Great idea. I added small comment for the whole section and also for each fix. It explains which file is being targeted and if the collector is enabled/disabled by default.

charts/prometheus-node-exporter/templates/daemonset.yaml
henrichter: Very nice suggestions. This really improves readability.

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -38,9 +38,56 @@ spec:
       {{- with .Values.priorityClassName }}
       priorityClassName: {{ . }}
       {{- end }}
-      {{- with .Values.extraInitContainers }}
+      {{- $fixes := .Values.permissionInitContainer.fixes -}}
+      {{- $fixesEnabled := or $fixes.rapl $fixes.slabinfo (not (empty .Values.permissionInitContainer.extraCommands)) -}}
+      {{- if or .Values.extraInitContainer $fixesEnabled }}
```
zeritti:
```suggestion
      {{- if or .Values.extraInitContainers $fixesEnabled }}
```

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -38,9 +38,56 @@ spec:
...
       initContainers:
-        {{- toYaml . | nindent 8 }}
+        {{- if .Values.extraInitContainer }}
+        {{- toYaml .Values.extraInitContainer | nindent 8 }}
```
zeritti:
```suggestion
        {{- if .Values.extraInitContainers }}
        {{- toYaml .Values.extraInitContainers | nindent 8 }}
```

henrichter:
Good catch! I accidentally missed that.

PR #6667, opened 2026-02-21
prometheus-snmp-exporter: use health probe path

JensErat:

Currently, we're relying on the root path (`/`), which works, but hosts an index page. snmp-exporter also has a dedicated health endpoint, which we should rather use.

https://github.com/prometheus/snmp_exporter/blob/337d524003d1932b7e95ba05fb15ef212533ac90/main.go#L320

- fixes #

#### Checklist

- [x] Chart Version bumped


walker-tom (approved):
LGTM, thanks for the PR!

PR #5462, opened 2025-03-20
[prometheus] Support dynamic image config via tpl

kozmagabor:
#### What this PR does / why we need it

This PR updates the Prometheus chart to use the tpl function for dynamic image repository, tag, and digest values. This enhancement is critical when using multiple subcharts, like Prometheus, and wanting to configure the image repository only once in the parent chart, without duplication.

#### Special notes for your reviewer
- Replaced static values with tpl for image repository, tag, and digest to allow the parent chart’s values to be passed dynamically to subchart.
- This change eliminates the need for duplication when using multiple instances of Prometheus (or other subcharts) and ensures consistent image configuration across all instances.

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6677, opened 2026-02-26
[prometheus] Add value to use distroless image variant

erpel:
Mention support for templating in the server.image.tag field from #5462. Also adds a note how to use this to switch to an image variant without loosing automatic version selection based on the chart app version.

Image variants were introduced with Prometheus 3.10.0 (https://github.com/prometheus/prometheus/releases/tag/v3.10.0)
> Prometheus now offers a distroless Docker image variant alongside the default
> busybox image. The distroless variant provides enhanced security with a minimal
> base image, uses UID/GID 65532 (nonroot) instead of nobody, and removes the
> VOLUME declaration. Both variants are available with -busybox and -distroless
> tag suffixes (e.g., prom/prometheus:latest-busybox, prom/prometheus:latest-distroless).
> The busybox image remains the default with no suffix for backwards compatibility
> (e.g., prom/prometheus:latest points to the busybox variant).

Originally I was going to introduce a new config value but after seeing that the tag already supports templating, this seemed like the less intrusive way to go. Still wanted to help others discover this way.

If you prefer a new value, I'd be happy to make that happen too but would like some input on how specific/general that should be. Options range from `server.image.distroless: true` to `server.image.suffix: "distroless".

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus
 # renovate: github=prometheus/prometheus
 appVersion: v3.10.0
-version: 28.12.1
+version: 28.13.0
 kubeVersion: ">=1.19.0-0"
 description: Prometheus is a monitoring system and time series database.
 home: https://prometheus.io/


zeritti:
Thank you, @erpel. Please, add a signature to your commit so that DCO check can pass.

erpel:
Forgetful me. Rebasing to add the signoff was an adventure, let's hope it did not mess anything up.

PR #2429, opened 2022-09-05
[kube-prometheus-stack]  Add support for templating in excludedFromEnforcement parameter

amuraru:
#### What this PR does / why we need it

Support for sensible defaults in downstream charts.

#### Checklist

- [x] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[prometheus-couchdb-exporter]`)

PR #6694, opened 2026-03-02
[kube-prometheus-stack] Add extra excluded from enforcement

sathieu:

We have a common configuration for all our clusters, and potential overrides per-environnement, per-zone, or per-cluster.

`.prometheus.prometheusSpec.excludedFromEnforcement` is an array. According to Helm values override, it is overridden completely, and there is no way to append values.

This PR adds `prometheus.prometheusSpec.extraExcludedFromEnforcement` using the same format, and allowing per-cluster overrides, while keeping it simple.

A more powerful solution would be to turn `excludedFromEnforcement` into an object, allowing fine-grained override. I can do such a PR if wanted. The array format would be kept for backward compatibility.

NB: https://github.com/prometheus-community/helm-charts/pull/2429 doesn't fit my usecase because templating is after `toYaml`.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 82.11.0
+version: 82.12.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.89.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -464,7 +464,11 @@ spec:
       name: "{{ printf "%s-%s" (include "kube-prometheus-stack.fullname" $) . | trunc 63 | trimSuffix "-" }}"
 {{- end }}
 {{- if .Values.prometheus.prometheusSpec.excludedFromEnforcement }}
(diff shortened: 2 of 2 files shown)


jkroepke (changes requested):
Hm, not happy here. You could use `additionalConfig` for your specifc use-case?

sathieu:
@jkroepke With `additionalConfig` we would have the field `excludedFromEnforcement` twice.

jkroepke:
I'm still not happy with that solution. If the next may request this pattern for other values as well. 

Did you every try this one?

```yaml
prometheus:
  prometheusSpec:
    excludedFromEnforcement: '{{ concat .Values.prometheus.prometheusSpec.baseExcludedFromEnforcement .Values.prometheus.prometheusSpec.extraExcludedFromEnforcement }}'
    baseExcludedFromEnforcement:
    - value1
    extraExcludedFromEnforcement:
    - value2
```

sathieu:
@jkroepke I have updated the template to make the following work:

```yaml
prometheus:
  prometheusSpec:
    enforcedNamespaceLabel: namespace
    excludedFromEnforcement:
      '{{ concat .Values.prometheus.prometheusSpec.baseExcludedFromEnforcement .Values.prometheus.prometheusSpec.extraExcludedFromEnforcement | toYaml }}'
    baseExcludedFromEnforcement:
    - group: monitoring.coreos.com
      resource: servicemonitors
      namespace: foo
      name: bar
    extraExcludedFromEnforcement:
    - group: monitoring.coreos.com
      resource: servicemonitors
      namespace: foo
      name: baz
```

WDYT?

jkroepke (approved):
LGTM

jkroepke:
Thanks

Issue #4228, opened 2024-02-06
[prometheus-kube-stack] Default ingress value pathType: ImplementationSpecific does't work on cilium ingress


adippl:
### Describe the bug a clear and concise description of what the bug is.

prometheus-kube-stack by default deploys prometheus ingress with `pathType: ImplementationSpecific`.
This this doesn't work with cilium ingress because `ImplementationSpecific` pathType acts as exact match.
Changing setting value `pathType: Prefix` solves the issue creates working ingress.
Please consider changing the default values or add a warning talking about this issue in the comments of values file.

The same issue in cilium repo. https://github.com/cilium/cilium/issues/30119

### What's your helm version?

version.BuildInfo{Version:"v3.11", GitCommit:"", GitTreeState:"", GoVersion:"go1.21.5"}

### What's your kubectl version?

Client Version: v1.28.5 Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3 Server Version: v1.28.2

### Which chart?

prometheus-kube-stack

### What's the chart version?

56.6.2

### Enter the changed values of values.yaml?

fullnameOverride: "kps"
prometheus:
  service:
    type: LoadBalancer
  networkPolicy:
    enabled: false

    flavor: kubernetes
  ingress:
    enabled: true
    ingressClassName: cilium
    annotations:
      cert-manager.io/cluster-issuer: "letsencrypt-prod"
    labels: {}
    hosts:
      - prometheus.DOMAIN
    path: /.*
    tls:
    - hosts:
        - prometheus.DOMAIN
      secretName: prometheus.DOMAIN
    pathType: Prefix
  prometheusSpec:
    priorityClassName: "high-priority"
    externalLabels:
      cluster: k8s3
    retention: 30d
    replicas: 1
    podAntiAffinity: "hard"
    storageSpec:
      volumeClaimTemplate:
        spec:
          storageClassName: rook-ceph-block-ssd
          accessModes: ["ReadWriteOnce"]
          resources:
            requests:
              storage: 25Gi
    resources:
     requests:
       cpu: 100m
       memory: 768Mi
     limits:
       cpu: 1000m
       memory: 2048Mi
    priorityClassName: "high-priority"

### Enter the command that you execute and failing/misfunctioning.

```
curl https://prometheus.<<<MYDOMAIN>>>/graph -v
*   Trying 10.0.5.98:443...
* Connected to prometheus.<<<MYDOMAIN>>> (10.0.5.98) port 443
[...]

PR #6698, opened 2026-03-03
[kube-prometheus-stack] Change ingress pathType default from ImplementationSpecific to Prefix

mirrorofstripes:

As per https://github.com/prometheus-community/helm-charts/issues/4228 the default ingress pathType of ImplementationSpecific is not functional on cilium ingress. Changing the default to `pathType: Prefix` is in line with the default setting of `path: /` and fixes the aforementioned issue. 

- fixes 4228

#### Checklist

- [ ] Chart Version bumped


QuentinBisson:
I'm not sure I see much value in changing this default. It has huge implications for existing users.
On top of that, gateway api is the future so I would try to not add to many changes to this part.

What do you think about providing a documentation in the values file revarding cilium?

PR #6699, opened 2026-03-03
[kube-prometheus-stack] Add VPA support for Alertmanager

QuentinBisson:
## Description

Add a `VerticalPodAutoscaler` resource for the Alertmanager `StatefulSet` managed by prometheus-operator.

The Alertmanager CR (`monitoring.coreos.com/v1`) creates a `StatefulSet` named `alertmanager-<cr-name>`. The VPA targets this StatefulSet for the `alertmanager` container, following the same pattern already used for `prometheusOperator.verticalPodAutoscaler`.

## Changes

- Add `charts/kube-prometheus-stack/templates/alertmanager/verticalpodautoscaler.yaml`
- Add `alertmanager.verticalPodAutoscaler` values in `values.yaml`

## Details

- **Target**: `StatefulSet/alertmanager-<cr-name>` (created by prometheus-operator from the Alertmanager CR)
- **Container**: `alertmanager`
- **Guard**: only rendered when `alertmanager.enabled=true`, `autoscaling.k8s.io/v1` API is available, and `alertmanager.verticalPodAutoscaler.enabled=true`

## Default values

```yaml
alertmanager:
  verticalPodAutoscaler:
    enabled: false
    # recommenders: []
    controlledResources: []
    # controlledValues: RequestsAndLimits
    maxAllowed: {}
    minAllowed: {}
    updatePolicy:
      updateMode: Auto
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -60,7 +60,7 @@ dependencies:
     repository: https://prometheus-community.github.io/helm-charts
     condition: nodeExporter.enabled
   - name: grafana
-    version: "11.2.2"
+    version: "11.2.3"
     repository: https://grafana-community.github.io/helm-charts
     condition: grafana.enabled
   - name: prometheus-windows-exporter


charts/kube-prometheus-stack/values.yaml
```diff
@@ -501,6 +501,37 @@ alertmanager:
...
+    # memory: 100Mi
+    # Define the min allowed resources for the pod
+    minAllowed: {}
+    # cpu: 200m
+    # memory: 100Mi
+
+    updatePolicy:
+      # Specifies whether recommended updates are applied when a Pod is started and whether recommended updates
+      # are applied during the life of a Pod. Possible values are "Off", "Initial", "Recreate", and "Auto".
+      updateMode: Auto
```
jkroepke: Auto is deprecated, feel like not a good default.

charts/kube-prometheus-stack/values.yaml
```diff
@@ -501,6 +501,37 @@ alertmanager:
...
+    # cpu: 200m
+    # memory: 100Mi
+    # Define the min allowed resources for the pod
+    minAllowed: {}
+    # cpu: 200m
+    # memory: 100Mi
+
+    updatePolicy:
+      # Specifies whether recommended updates are applied when a Pod is started and whether recommended updates
+      # are applied during the life of a Pod. Possible values are "Off", "Initial", "Recreate", and "Auto".
```
jkroepke:
```suggestion
      # are applied during the life of a Pod. Possible values are "Off", "Initial", "Recreate", and "InPlaceOrRecreate".
```

charts/kube-prometheus-stack/templates/alertmanager/verticalpodautoscaler.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if and .Values.alertmanager.enabled (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") .Values.alertmanager.verticalPodAutoscaler.enabled }}
```
jkroepke:
```suggestion
{{- if and .Values.alertmanager.enabled .Values.alertmanager.verticalPodAutoscaler.enabled }}
```

I'm not a fan of the `Capabilities.APIVersions.Has` check. As user, if I explicit enable `alertmanager.verticalPodAutoscaler.enabled`, then I would expect that the `VerticalPodAutoscaler` is present. But if my cluster does not have that kind of API, the error should appear that the API is missing. With the `Capabilities.APIVersions.Has` check, no error appears which is mis-leading.

charts/kube-prometheus-stack/values.yaml
QuentinBisson: True i switched to Recreate

charts/kube-prometheus-stack/values.yaml
QuentinBisson: done

charts/kube-prometheus-stack/templates/alertmanager/verticalpodautoscaler.yaml
QuentinBisson: I removed it

PR #6707, opened 2026-03-05
[prometheus-blackbox-exporter]: warn about binary config corruption via toYaml

bossm8:
Adds a warning comment in values.yaml and a dedicated section in README.md documenting a silent data corruption issue affecting users who configure binary protocol modules (e.g. PostgreSQL, LDAP) via the config: value.
The contents of config are passed through toYaml in the ConfigMap template. This round-trips the value through Helm's YAML parser, which decodes !!binary tags and re-encodes the result as a UTF-8 string. Any byte >= 0x80 is silently corrupted or dropped. The !!binary tag is lost, and the payload sent to the target server is malformed. No warning is produced during helm install or helm template — the failure only appears at probe time as a cryptic byte-mismatch error.
The fix documents the limitation and points users to configExistingSecretName as the correct workaround.

No existing issue tracks this. This PR is the first documentation of the problem.

This is a documentation-only change — no template or logic changes. The underlying limitation cannot be fixed without a breaking change to how config is rendered, which is out of scope here.
The corruption can be verified by running helm template with a module containing send: !!binary AAAACATSFi8= (PostgreSQL SSLRequest) and inspecting the rendered ConfigMap — the 0xd2 byte is silently corrupted:

```yaml
postgres:
  prober: tcp
  timeout: 10s
  tcp:
    preferred_ip_protocol: ip4
    query_response:
      - send: !!binary 'AAAACATSFi8='
      - expect_bytes: 'S'
      - starttls: true
```

After `helm template`:

```yaml
postgres:
  prober: tcp
  tcp:
    preferred_ip_protocol: ip4
    query_response:
    - send: "\0\0\0\b\x04�\x16/"
    - expect_bytes: S
    - starttls: true
  timeout: 10s
```

We could also consider adding a `configRaw`, which is passed as string to the CM template, but since it works with `configExistingSecretName` documentation should be fine - might prevent some headaches.

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
please rasie chart version too

Issue #6640, opened 2026-02-12
[prometheus-blackbox-exporter] Use `kubernetes.io/metadata.name` for the monitoring namespace label key in networkPolicy


yrosemacbs:
### Is your feature request related to a problem ?

The networkPolicy set by the chart supports allowing only a monitoring namespace. The way this works is by adding a label with key `name` to the monitoring namespace.
In some case the monitoring namespace is not in control by the deployer of this chart and so the namespace label cannot be set. Also, kubernetes promotes the use of some well-known labels: <https://kubernetes.io/docs/reference/labels-annotations-taints/>

### Describe the solution you'd like.

Use `kubernetes.io/metadata.name` as the namespace label key for the monitoring namespace instead of `name`.

### Describe alternatives you've considered.

NONE

PR #6713, opened 2026-03-06
[prometheus-blackbox-exporter] use kubernetes.io/metadata.name in networkPolicy

firasmosbehi:
## Summary
This PR updates `prometheus-blackbox-exporter` network policy namespace matching to use Kubernetes' well-known namespace label and addresses https://github.com/prometheus-community/helm-charts/issues/6640.

## Problem
When `networkPolicy.allowMonitoringNamespace` is enabled, the chart currently matches on `name=<namespace>`. In many environments, deployers do not control namespace labels and cannot guarantee a custom `name` label exists.

## Root cause
`templates/networkpolicy.yaml` hardcodes the namespace selector key to `name`.

## Fix
- Switched namespace selector label key from `name` to `kubernetes.io/metadata.name`.
- Updated values comment to describe the new expected label semantics.
- Bumped chart version from `11.8.0` to `11.8.1`.

## Validation
- `helm lint charts/prometheus-blackbox-exporter`
- `helm template test charts/prometheus-blackbox-exporter --set networkPolicy.enabled=true --set networkPolicy.allowMonitoringNamespace=true --set networkPolicy.monitoringNamespaceName=observability`
- Verified rendered selector contains `kubernetes.io/metadata.name: observability`.

## Notes
This keeps the same user-facing value (`monitoringNamespaceName`) while aligning selector behavior with a standard Kubernetes label.

### DIFF
--- a/charts/prometheus-blackbox-exporter/templates/networkpolicy.yaml
+++ b/charts/prometheus-blackbox-exporter/templates/networkpolicy.yaml
@@ -15,7 +15,7 @@ spec:
   - from:
     - namespaceSelector:
         matchLabels:
-          name: {{ .Values.networkPolicy.monitoringNamespaceName }}
+          kubernetes.io/metadata.name: {{ .Values.networkPolicy.monitoringNamespaceName }}
     ports:
     - port: {{ .Values.service.port }}
       protocol: TCP
@@ -25,4 +25,3 @@ spec:
   policyTypes:
   - Ingress
 {{- end }}
-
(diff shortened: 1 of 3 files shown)


charts/prometheus-blackbox-exporter/templates/networkpolicy.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   - from:
     - namespaceSelector:
         matchLabels:
-          name: {{ .Values.networkPolicy.monitoringNamespaceName }}
+          kubernetes.io/metadata.name: {{ .Values.networkPolicy.monitoringNamespaceName }}
```
monotek, line 18: As this is changing the current default behaviour i consider this a breaking change.
Therefore major version of the chart woul dneed an update.
Maybe make key and value configurable so the current default can be kept.

monotek (changes requested):

Issue #6250, opened 2025-10-17
[prometheus-node-exporter] Allow to fully configure RBAC proxy listen address


sushiMix:
### Is your feature request related to a problem ?

Hello,
We have a deployment with multiple network card on the nodes and would like to allow the RBAC proxy to only listen to one interface.

The current listen address for the proxy is
`- --secure-listen-address=:{{ .Values.service.port}}`
where **:** is like 0.0.0.0.

As node exporter requires host network to correctly grab the metrics the RBAC proxy listen on all interfaces of the physical host.

### Describe the solution you'd like.

We would like to be able to configure the listen address like this.
`- --secure-listen-address={{ .Values.kubeRBACProxy.listenHost }}{{ .Values.service.port}}`
with a default value for .Values.kubeRBACProxy.listenHost of **:**

It will allow us to generate a configuration like this without breaking existing configurations
``` yaml
args:
- --secure-listen-address=$(POD_IP):9100
env:
- name: POD_IP
   valueFrom:
       fieldRef:
           fieldPath: status.podIP
```

### Describe alternatives you've considered.

The only option we found is to use kustomize to change the bahavior

PR #6716, opened 2026-03-06
[prometheus-node-exporter] make kube-rbac-proxy listen host configurable

firasmosbehi:
## Summary
This PR makes kube-rbac-proxy listen host configurable in `prometheus-node-exporter` and addresses https://github.com/prometheus-community/helm-charts/issues/6250.

## Problem
When `kubeRBACProxy.enabled=true`, the chart hardcodes `--secure-listen-address=:<port>`, which binds on all interfaces. Users with multi-NIC nodes cannot constrain exposure to a specific interface/IP without post-render patching.

## Root cause
The daemonset template hardcodes `--secure-listen-address` host prefix and does not expose a corresponding value.

## Fix
- Added `kubeRBACProxy.listenHost` in values with default `":"` to preserve existing behavior.
- Updated kube-rbac-proxy args to use `--secure-listen-address={{ .Values.kubeRBACProxy.listenHost }}{{ .Values.service.port }}`.
- Bumped chart version from `4.52.0` to `4.52.1`.

## Validation
- `helm lint charts/prometheus-node-exporter`
- `helm template test charts/prometheus-node-exporter --set kubeRBACProxy.enabled=true --set service.port=9100 --set-string 'kubeRBACProxy.listenHost=$(POD_IP):'`
- Verified rendered arg contains `--secure-listen-address=$(POD_IP):9100`.

## Compatibility
Default behavior remains unchanged because `listenHost` defaults to `":"` (all interfaces).

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.52.0
+version: 4.52.1
 # renovate: github=prometheus/node_exporter
 appVersion: 1.10.2
 home: https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -240,7 +240,7 @@ spec:
             {{-  if .Values.kubeRBACProxy.extraArgs  }}
             {{- .Values.kubeRBACProxy.extraArgs | toYaml | nindent 12 }}
             {{-  end  }}
-            - --secure-listen-address=:{{ .Values.service.port}}
+            - --secure-listen-address={{ .Values.kubeRBACProxy.listenHost }}{{ .Values.service.port}}
             - --upstream=http://127.0.0.1:{{ $servicePort }}/
             - --proxy-endpoints-port={{ .Values.kubeRBACProxy.proxyEndpointsPort }}
             - --config-file=/etc/kube-rbac-proxy-config/config-file.yaml
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -61,6 +61,9 @@ kubeRBACProxy:
 
   # Specify the port used for the Node exporter container (upstream port)
   port: 8100
+  # Host prefix for kube-rbac-proxy secure listen address.
+  # Keep ":" to listen on all interfaces, or set values like "127.0.0.1:" or "$(POD_IP):".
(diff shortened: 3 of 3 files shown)


charts/prometheus-node-exporter/values.yaml
```diff
@@ -61,6 +61,9 @@ kubeRBACProxy:
 
   # Specify the port used for the Node exporter container (upstream port)
   port: 8100
+  # Host prefix for kube-rbac-proxy secure listen address.
+  # Keep ":" to listen on all interfaces, or set values like "127.0.0.1:" or "$(POD_IP):".
+  listenHost: ":"
```
zeritti, line 66: If not set, the field's value is an empty string:

```suggestion
  listenHost: ""
```

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -240,7 +240,7 @@ spec:
             {{-  if .Values.kubeRBACProxy.extraArgs  }}
             {{- .Values.kubeRBACProxy.extraArgs | toYaml | nindent 12 }}
             {{-  end  }}
-            - --secure-listen-address=:{{ .Values.service.port}}
+            - --secure-listen-address={{ .Values.kubeRBACProxy.listenHost }}{{ .Values.service.port}}
```
zeritti, line 243:
```suggestion
            - --secure-listen-address={{ printf "%s:%d" .Values.kubeRBACProxy.listenHost (.Values.service.port | int) }}
```

charts/prometheus-node-exporter/values.yaml
```diff
@@ -61,6 +61,9 @@ kubeRBACProxy:
 
   # Specify the port used for the Node exporter container (upstream port)
   port: 8100
+  # Host prefix for kube-rbac-proxy secure listen address.
+  # Keep ":" to listen on all interfaces, or set values like "127.0.0.1:" or "$(POD_IP):".
```
zeritti, line 65:
```suggestion
  # Keep default to listen on all interfaces, or set a custom value like "$(POD_IP)".
```

The only reasonable setting for the IP to listen on is the pod's IP (POD_IP). Setting an environment variable in the container from a fieldRef is currently not supported by the chart, though.

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.52.0
+version: 4.52.1
```
zeritti, line 9:
```suggestion
version: 4.53.0
```

zeritti (changes requested):

Issue #6334, opened 2025-11-19
[kube-prometheus-stack] kube scheduler metrics scraping "additionalPath" key


KnackStyxing:
### Is your feature request related to a problem ?

Yes. Kube scheduler hosts it's metrics on both the paths `/metrics` and `/metrics/resources` according to the [docs](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#kube-scheduler-metrics). The base configuration of the path for the [kube-scheduler serviceMonitor](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/exporters/kube-scheduler/servicemonitor.yaml) only scrapes path `/metrics` by default.

### Describe the solution you'd like.

Adding `spec.endpoints.additionalPath` to the kube-scheduler serviceMonitor will allow users to specify additional paths for the serviceMonitor to scrape. The solution would have to make a second instance of `spec.endpoints` that is populated with a different `spec.endpoints.additionalPath` if `spec.endpoints.additionalPath` is set in values.yaml.

### Describe alternatives you've considered.

Another implementation could look like a feature flag that adds the `/metrics/resources` if the flag is set to true in values.yaml. It should not be hard-coded(like the kubelet serviceMonitor does with `/metrics/cadvisor`) in case kubernetes implements another endpoint url in the future. The ideal solution changes the `path` type in [prometheus operator Endpoint API reference](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Endpoint) from `string` to `[]string` where the user either explicitly states every metric path (including the now-default `/metrics`) or if `spec.endpoints.additionalPath` remains blank then only `/metrics` is enabled else make endpoints for each `spec.endpoints.additionalPath[i]` stated (not needing `/metrics` explicitly stated).

### Additional context.

I am willing to write the PR for this if you want. Additionally, if you would like me to streamline this for all serviceMonitors I can do that too.

PR #6718, opened 2026-03-06
[kube-prometheus-stack] add kubeScheduler serviceMonitor additionalPath

firasmosbehi:
## Summary
This PR adds support for an additional kube-scheduler ServiceMonitor path in `kube-prometheus-stack`, addressing https://github.com/prometheus-community/helm-charts/issues/6334.

## Problem
The kube-scheduler ServiceMonitor currently renders a single endpoint scrape. Users who need metrics from additional scheduler paths (for example `/metrics/resources`) cannot configure this without patching manifests externally.

## Root cause
`templates/exporters/kube-scheduler/servicemonitor.yaml` only emits one endpoint and no value exists to configure a second path.

## Fix
- Added `kubeScheduler.serviceMonitor.additionalPath` (default `""`) to values.
- Updated kube-scheduler ServiceMonitor template to render a second endpoint when `additionalPath` is set.
- Reused the same scrape settings (port, interval, auth, TLS, relabelings) for the extra endpoint and set only `path` differently.
- Bumped chart version from `82.9.0` to `82.9.1`.

## Validation
- `helm lint charts/kube-prometheus-stack`
- `helm template test charts/kube-prometheus-stack --set kubeScheduler.serviceMonitor.additionalPath=/metrics/resources`
- Verified rendered kube-scheduler ServiceMonitor includes an additional endpoint with `path: /metrics/resources`.

## Compatibility
No behavior change by default since `additionalPath` is empty unless explicitly configured.

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 82.9.0
+version: 82.9.1
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.89.0
 kubeVersion: ">=1.25.0-0"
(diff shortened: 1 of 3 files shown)


jkroepke (commented):
Please make it more explcit to scrape `metrics/resources` as additional feature, disabled by default.

PR #6720, opened 2026-03-06
[kube-state-metrics] fix fullname duplication with nameOverride

firasmosbehi:
## Summary
This PR fixes fullname generation for `kube-state-metrics` when `nameOverride` already includes the release name, addressing https://github.com/prometheus-community/helm-charts/issues/6510.

## Problem
With `Release.Name=ksm-non-pods` and `nameOverride=ksm-non-pods-foo`, rendered object names become double-prefixed (`ksm-non-pods-ksm-non-pods-foo`).

## Root cause
The fullname helper only handled the case where release name contains chart/name. It did not handle the inverse case where `nameOverride` already contains release name.

## Fix
Updated `kube-state-metrics.fullname` helper logic to add an inverse containment branch:
- keep existing behavior if release name already contains chart/name
- if `nameOverride` contains release name, use `nameOverride` as fullname
- otherwise keep concatenation behavior

Also bumped chart version from `7.2.0` to `7.2.1`.

## Validation
- `helm lint charts/kube-state-metrics`
- `helm template ksm-non-pods charts/kube-state-metrics --set nameOverride=ksm-non-pods-foo`
- Verified rendered names are `ksm-non-pods-foo` and no `ksm-non-pods-ksm-non-pods-foo` names are produced.

## Compatibility
Existing default naming behavior is preserved unless `nameOverride` already embeds the release name.

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 7.2.0
+version: 7.2.1
 # renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.18.0
 home: https://github.com/kubernetes/kube-state-metrics/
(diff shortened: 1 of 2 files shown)


sebastiangaiser:
Thanks! Same reasoning as #6557, we're keeping the standard helper (it's the `helm create` default; details on #6510). #7104 documents both `nameOverride` and `fullnameOverride`, which already solves this case without diverging from the other ~40 charts.

Issue #5651, opened 2025-05-19
[kube-state-metrics] allign metricAnnotationsAllowList value name


pierremahot:
### Is your feature request related to a problem ?

no

### Describe the solution you'd like.

metricAnnotationsAllowList => metricAnnotationsAllowlist

### Describe alternatives you've considered.

none

### Additional context.

all other var name are  Allowlist not AllowList

benjamin-wright:
This just caught us out. If we don't want to update to actual value for backwards-compatibility reasons then at least the values.yaml should be updated to reflect the actual expected value.

pierremahot:
may be support both typo can be a non breaking change improvement. default to one or the other ? eventually do this every ware but totaly understand not to do it.

DrFaust92:
#6617 is open for this — it adds `metricAnnotationsAllowlist` as an alias so the spelling lines up with the other keys without breaking anyone still using `metricAnnotationsAllowList`. Worth a look/review rather than a second attempt.

PR #6617, opened 2026-02-09
[kube-state-metrics] Add metricAnnotationsAllowlist key alias

firasmosbehi:
## Summary
- add the preferred value key metricAnnotationsAllowlist
- keep metricAnnotationsAllowList as a backward-compatible alias
- render --metric-annotations-allowlist from either key
- bump kube-state-metrics chart version to 7.1.1

## Related
- Closes #5651

## Testing
- helm lint charts/kube-state-metrics
- ct lint --config .github/linters/ct.yaml --charts charts/kube-state-metrics

PR #6721, opened 2026-03-06
[kube-state-metrics] align metricAnnotationsAllowlist naming

firasmosbehi:
## Summary
This PR adds the normalized `metricAnnotationsAllowlist` values key to `kube-state-metrics` while preserving backward compatibility with the existing `metricAnnotationsAllowList` key, addressing https://github.com/prometheus-community/helm-charts/issues/5651.

## Problem
The chart uses `metricAnnotationsAllowList` while related options use the `Allowlist` casing convention. This is inconsistent and causes confusion.

## Root cause
The value and template wiring were implemented only with the older `AllowList` casing.

## Fix
- Added `metricAnnotationsAllowlist` to `values.yaml`.
- Kept `metricAnnotationsAllowList` as a deprecated compatibility alias.
- Updated deployment args rendering to prefer `metricAnnotationsAllowlist` and fall back to `metricAnnotationsAllowList`.
- Bumped chart version from `7.2.0` to `7.2.1`.

## Validation
- `helm lint charts/kube-state-metrics`
- `helm template test charts/kube-state-metrics --set-string 'metricAnnotationsAllowlist[0]=pods=[kubernetes.io/team]'`
- `helm template test charts/kube-state-metrics --set-string 'metricAnnotationsAllowList[0]=pods=[legacy/key]'`
- Verified both render the expected `--metric-annotations-allowlist=...` argument.

## Compatibility
Backward compatible by design: existing users of `metricAnnotationsAllowList` continue to work.

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 7.2.0
+version: 7.2.1
 # renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.18.0
 home: https://github.com/kubernetes/kube-state-metrics/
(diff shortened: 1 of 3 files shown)


sebastiangaiser:
Thanks! This is essentially a duplicate of your own #6617, which we're using as the base for this change. I'll close this one in favor of #6617 to keep everything in a single PR.

PR #6734, opened 2026-03-15
[kube-prometheus-stack] Add hostUsers support for Prometheus, Alertmanager and ThanosRuler

### LINKED ISSUE #5980: [all] Allow enabling user namespaces

DrFaust92:

Exposes the `hostUsers` field for Prometheus, Alertmanager, and ThanosRuler CRs, allowing users to opt into Kubernetes user namespaces by setting `hostUsers: false`.

Added to:
- `alertmanager.alertmanagerSpec.hostUsers`
- `prometheus.prometheusSpec.hostUsers`
- `thanosRuler.thanosRulerSpec.hostUsers`

All default to `~` (nil/unset) to preserve existing behavior. The field is only rendered when explicitly set to a boolean.

- fixes #5980

Uses `kindIs "bool"` guard (same pattern as `prometheusOperator.hostUsers`) so the field is omitted when nil.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
+++ b/charts/kube-prometheus-stack/templates/alertmanager/alertmanager.yaml
@@ -212,6 +212,9 @@ spec:
 {{ toYaml .Values.alertmanager.alertmanagerSpec.updateStrategy | indent 4 }}
 {{- end }}
   hostNetwork: {{ .Values.alertmanager.alertmanagerSpec.hostNetwork }}
+{{- if .Values.alertmanager.alertmanagerSpec.terminationGracePeriodSeconds }}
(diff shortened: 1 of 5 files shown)


QuentinBisson (approved):

QuentinBisson:
@DrFaust92 can you bump the chart version?

PR #6742, opened 2026-03-17
[kube-state-metrics] fix: allow disabling cluster-wide CRD RBAC

### LINKED ISSUE #6708: [kube-state-metrics] BUG: not possible to use customResourceState without cluster-scope privileges.

https://github.com/prometheus-community/helm-charts/blob/db8ca60a6829208839708a80166fb5540915d710/charts/kube-state-metrics/templates/role.yaml#L226-L231

customResourceState.enabled: True is erroneously hardcoded to require cluster-scoped customresourcedefinitions list/watch privileges

### What's your helm version?

irrelevant

irrelevant

### Which chart?

kube-state-metrics

### What's the chart version?

HEAD

### What happened?

Attempted to use customResourceState to scrape CRD `Kustomization (kustomize.toolkit.fluxcd.io)`. I am a namespace admin and not a cluster admin and thus receive the error:

```
2026-03-05T17:07:01.946298718Z: warning: Upgrade "maas" failed: cannot patch "kube-state-metrics" with kind Role: roles.rbac.authorization.k8s.io "kube-state-metrics" is forbidden: user "system:serviceaccount:my-namespace:flux-reconciler" (groups=["system:serviceaccounts" "system:serviceaccounts:my-namespace" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["apiextensions.k8s.io"[], Resources:["customresourcedefinitions"[], Verbs:["list" "watch"]}
2026-03-05T17:07:02.089229539Z: Cleanup on fail set, cleaning up 1 resources
```

This should not be required, as this is only necessary [for wildcard CRD discovery/scraping](https://github.com/kubernetes/kube-state-metrics/blob/f818168809209a534248d4e6dc20429d5ae7c35d/internal/discovery/discovery.go#L47). If group/version/kind are fixed, we should be able to scrape given CRD status with just the CRD privileges.

### How to reproduce it?

- Attempt to use customResourceState to scrape CRDs without customresourcedefinitions list/watch privileges

_No response_

(flux reconcile)

_No response_

Nirpesh551:
Fixes an issue where `kube-state-metrics` forcefully requires cluster-wide `customresourcedefinitions` list/watch privileges when `customResourceState.enabled` is true. This prevents deployment for users who only have namespace-level privileges. 

This PR adds a new values flag `rbac.customResourceState.createClusterRoleRules` (defaulting to `true` for backwards compatibility) that allows users to opt out of the cluster-wide CRD RBAC rules while still using custom resource state metrics.

Refs #6708

Hi! This is my first open-source contribution. Let me know if anything needs adjusting!

#### Checklist
- [x] Chart Version bumped


charts/kube-state-metrics/templates/role.yaml
```diff
@@ -235,7 +235,7 @@ rules:
     - subjectaccessreviews
   verbs: ["create"]
 {{- end }}
-{{- if $.Values.customResourceState.enabled }}
+{{- if and $.Values.customResourceState.enabled (default true $.Values.rbac.customResourceState.createClusterRoleRules) }}
```
sebastiangaiser:
```suggestion
{{- if and $.Values.customResourceState.enabled $.Values.rbac.customResourceState.createClusterRoleRules }}
```

charts/kube-state-metrics/values.yaml
```diff
@@ -89,6 +89,9 @@ hostUsers: ~
 rbac:
   # If true, create & use RBAC resources
   create: true
+  # If true, grants cluster-wide list/watch permissions for customresourcedefinitions when customResourceState.enabled is true
+  customResourceState:
+    createClusterRoleRules: true
```
```suggestion
  # When customResourceState.enabled is true, KSM normally needs cluster-wide
  # list/watch on customresourcedefinitions to auto-discover CRDs. Set to false
  # to omit that ClusterRole rule if you don't need CRD auto-discovery (e.g.
  # namespace-scoped setups that grant only their own CRD access).
  customResourceState:
    createClusterRoleRules: true
```

sebastiangaiser (changes requested):
Thanks for the PR! 
One blocker: the opt-out doesn't take effect yet. In the role.yaml condition, `default true $.Values.rbac.customResourceState.createClusterRoleRules` always evaluates to `true`, because Helm's default treats a boolean `false` as "empty" and substitutes the default, so setting the flag to `false` is ignored and the CRD rule still renders. Since the default already lives in values.yaml, dropping `default true` (suggestion inline) fixes it: the rule renders by default and is correctly dropped when set to `false`.
Two more things:
- Could you please add a helm-unittest covering both states (`enabled` + default --> CRD rule present; `enabled` + `false` --> absent)? That locks in the behavior.
- Since this adds a new value, this is a minor bump, `8.1.0` rather than `8.0.1`.

Nirpesh551:
Hey @sebastiangaiser, thanks for the thorough review! I've pushed everything you asked for:
Fixed the default true logic exactly like you suggested — that was a sneaky one, thanks for catching it
Added unit tests covering all three states (default enabled, explicitly disabled, and customResourceState off entirely)
Bumped the version — ended up at 8.1.4 instead of 8.1.0 since a few releases went out upstream while I was working on this
Also had to merge in the latest main to resolve a conflict on Chart.yaml, and fix a DCO sign-off issue on one of my commits, so there's a bit more history in there now than before. Let me know if anything still needs tweaking!

sebastiangaiser:
Thanks, the default-true fix is right and the opt-out takes effect now. One thing: the new test is under `tests/`, but this chart keeps its suites in `unittests/` and CI runs `helm unittest --file 'unittests/**/*.yaml` (.github/linters/ct.yaml), so `tests/rbac_customresourcestate_test.yaml` won't be picked up. Could you please move it to `unittests/`? The suite itself is fine, just the location.

Nirpesh551:
Ah good catch, didn't realize this chart already had a unittests/ folder, moved it over now, should get picked up by CI properly this time.

charts/kube-state-metrics/values.yaml
```diff
@@ -94,6 +94,12 @@ hostUsers: ~
 rbac:
   # If true, create & use RBAC resources
   create: true
+  # When customResourceState.enabled is true, KSM normally needs cluster-wide
+  # list/watch on customresourcedefinitions to auto-discover CRDs. Set to false
+  # to omit that ClusterRole rule if you don't need CRD auto-discovery (e.g.
+  # namespace-scoped setups that grant only their own CRD access).
```
```suggestion
  # Set to false only when the customresourcedefinitions list/watch permission is
  # granted to the ServiceAccount by other means (e.g. a cluster-scoped ClusterRole
  # managed outside this chart). kube-state-metrics always needs it for
  # customResourceState; this only controls whether the chart creates the rule.
```

sebastiangaiser (approved):

PR #6764, opened 2026-03-23
[kube-prometheus-stack] fix: missing condition leads to error

rndmh3ro:

Fixes this error when using cilium:

```
Error: kube-prometheus-stack/templates/prometheus/ciliumnetworkpolicy.yaml:12:18
  executing "kube-prometheus-stack/templates/prometheus/ciliumnetworkpolicy.yaml" at <.Values.prometheus.networkPolicy.cilium.endpointSelector>:
    nil pointer evaluating interface {}.endpointSelector

Use --debug flag to render out invalid YAML
```

#### Checklist

- [x] Chart Version bumped


charts/kube-prometheus-stack/templates/prometheus/ciliumnetworkpolicy.yaml
```diff
@@ -9,17 +9,17 @@ metadata:
     {{- include "kube-prometheus-stack.labels" . | nindent 4 }}
 spec:
   endpointSelector:
-    {{- if .Values.prometheus.networkPolicy.cilium.endpointSelector }}
+    {{- if and .Values.prometheus.networkPolicy.cilium .Values.prometheus.networkPolicy.cilium.endpointSelector }}
```
QuentinBisson: Thanks for the contribution. What do you think about removing it from all the IFs here and add it at the if at line 1 instead? This would make the template cleaner

PR #6780, opened 2026-03-25
[prometheus-blackbox-exporter] Add tpl to servicemonitors

### LINKED ISSUE #6776: [prometheus-blackbox-exporter] Add tpl the targets url

We use ArgoCD to deploy multiple blackbox exporter on multiple kubernetes clusters. We use an application of applications to deploy the blackbox helm charts and inject a variable for the cluster name of the cluster.
Each cluster has similar urls that contain the cluster name and we want to be able to template the url for all the clusters easily

When we use the targets, we want to specify the urls per each cluster, as so:
```
url: grafana.{{ cluster_name }}.org.com
```
Or
```
url: dns-{{ cluster_name }}.org.com
```

We have considered creating a values file for each cluster, but we ended up with 25 values files and we quickly encountered drifts between their configurations

### Additional context.

_No response_

poxelyric:
Adds an option to template values of service monitors directly from values.yaml with the tpl function.

- fixes #6776 

#### Checklist

- [x] Chart Version bumped


monotek (changes requested):
please add tests (non tpl / tpl usage) in ci directory.

poxelyric:
hi @monotek thanks for the review. I added tpl and non tpl targets in the same value file, is that ok?

PR #6800, opened 2026-04-01
[kube-prometheus-stack] Make CRDs upgrade Job pod-level automountServiceAccountToken configurable

sslavic:

This PR is a follow-up to https://github.com/prometheus-community/helm-charts/pull/5175 which introduced upgrading CRDs using Helm. While trying to upgrade a kube-prometheus-stack helm release from old 5x to 82.14.1 chart, and to enable the CRDs upgrade Job, I ran into issue - CRDs upgrade Job Pod submission was being rejected by a custom (Kyverno) policy requiring service account auto mounting to be disabled on Pod. Unfortunately the policy is on Pod level only, isn't able to validate ServiceAccount of the Job Pod, that auto mounting is disabled there via already supported flag.

As alternative I've tried using kustomize post renderer patch, to add automountServiceAccountToken to the upgrade CRDs Job Pod template, but even that is failing, the target Job resource can't be found even though patch is valid (suspecting due to a bug at least in Flux+helm+kustomize versions used, which are failing to deal with subcharts and pre-upgrade hooks like the case of this Job).

Another alternative would be to add policy exception, or make policy more complex to validate referenced ServiceAccount too.

Yet another alternative would be not to enable the Job, to manage CRDs outside of the helm release.

Upgrading+debugging Flux is another option but a lenghty process.

Having chart improved is preferred over all of these alternatives.

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 82.16.0
+version: 82.16.1
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.89.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/charts/crds/templates/upgrade/job.yaml
+++ b/charts/kube-prometheus-stack/charts/crds/templates/upgrade/job.yaml
@@ -33,6 +33,7 @@ spec:
       imagePullSecrets:
         {{- include "kube-prometheus-stack.imagePullSecrets" . | indent 8 }}
       {{- end }}
+      automountServiceAccountToken: {{ .Values.upgradeJob.automountServiceAccountToken }}
       serviceAccountName: {{ include "kube-prometheus-stack.crd.upgradeJob.serviceAccountName" . }}
       initContainers:
         - name: busybox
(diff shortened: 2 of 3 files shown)


charts/kube-prometheus-stack/charts/crds/templates/upgrade/job.yaml
```diff
@@ -33,6 +33,7 @@ spec:
       imagePullSecrets:
         {{- include "kube-prometheus-stack.imagePullSecrets" . | indent 8 }}
       {{- end }}
+      automountServiceAccountToken: {{ .Values.upgradeJob.serviceAccount.automountServiceAccountToken }}
```
jkroepke: Normally, `automountServiceAccountToken` on pod level have a distinct values which not mimic the serviceAccount once.

sslavic: @jkroepke good one, thanks. I've update the PR, PTAL.

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 82.16.0
+version: 82.16.1
```
sslavic, line 31: @jkroepke should this stay patch bump or maybe minor bump instead now that a new flag is being introduced?

jkroepke:
Thanks!

PR #6842, opened 2026-04-14
[prometheus-node-exporter] allow setting variant for node-exporter image

### LINKED ISSUE #6834: [prometheus-node-exporter] allow usage of new distroless image

n/a

A new distroless image variant of node exporter was added in https://github.com/prometheus/node_exporter/pull/3592. It would be great to have a value (e.g. `image.variant`) to allow using this. It might even make sense to make this the new chart default, given it is probably the preferable choice for almost all usecases.

Trying to hack this in with a custom tag/version number. This would require regular version bumps on my end, which I would like to avoid.

### Additional context.

If a plan for implementation is suggested, I'd be very much willing to open a PR.

jnoordsij:
Allow one to pick a variant of the node-exporter image, e.g. `distroless`.

- fixes #6834

I've opted for a more 'generic' variant value, rather than a distroless boolean, so that one can also explicitly use the `busybox` variant or any potential future variant.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -6,6 +6,8 @@ image:
   repository: prometheus/node-exporter
   # Overrides the image tag whose default is {{ printf "v%s" .Chart.AppVersion }}
   tag: ""
+  # Use the distroless image variant.
+  distroless: false
   pullPolicy: IfNotPresent
   digest: ""
 
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.54.0
+version: 4.55.0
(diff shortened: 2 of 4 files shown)


charts/prometheus-node-exporter/values.yaml
```diff
@@ -6,6 +6,8 @@ image:
   repository: prometheus/node-exporter
   # Overrides the image tag whose default is {{ printf "v%s" .Chart.AppVersion }}
   tag: ""
+  # Can be used to specify a variant, e.g. 'distroless'
+  variant: ""
```
zeritti: Can we please use `distroless` (bool) as in the prometheus Helm chart?

zeritti (changes requested):
Thank you, @jnoordsij, for your PR. Please, see my comment below.

jnoordsij:
Thanks for the review @zeritti, I've adjusted accordingly.

zeritti (approved):
Thanks, @jnoordsij, LGTM

PR #6883, opened 2026-05-01
[kube-prometheus-stack] Use distroless images by default

jnoordsij:

This MR does two things:
- allows setting the `distroless` variant of the Prometheus image
- changes defaults to use the `distroless` variant for this image and the `prometheus-node-exporter` image by default (see also #6842)

Given the latter change is opinionated, where I personally would consider it a reasonable and more secure default, I've intentionally singled out this change so that it can be extracted to a separate MR or dropped entirely if preferable. 

n/a

n/a

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/UPGRADE.md
+++ b/charts/kube-prometheus-stack/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade
 
+## From 84.x to 85.x
+
+This version enables the `distroless` variant of both the `prometheus` and `prometheus-node-exporter` images by default
+
 ## From 83.x to 84.x
 
 This version upgrades Grafana to v13.0.0. Please check [What's new in v13.0](https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v13-0/)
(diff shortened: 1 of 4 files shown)


jkroepke (changes requested):
No over-engineered toggle please. Just change the image value.

jnoordsij:
The toggles were to mirror the efforts for other charts here in #6842 and #6677, so I figured it would make sense to change things here to be aligned with that. But perfectly fine to me to just leverage the tag value for that.

charts/kube-prometheus-stack/UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade
 
+## From 84.x to 85.x
+
+This version enables the `distroless` variant of both the `prometheus` and `prometheus-node-exporter` images by default
```
jkroepke:
```suggestion
This version enables the `distroless` variant of both the `prometheus` and `prometheus-node-exporter` images by default.

If you sync images to your private registry, ensure that you also sync image tags with the distroless suffix.
```

jkroepke (approved):

PR #6887, opened 2026-05-04
[prometheus-postgres-exporter] require datasource password explicitly

InfraArtists:

Removes the hardcoded `"somepaswword"` (note: also a typo) fallback in the
generated `Secret` and replaces it with an explicit `fail` when none of
`config.datasource.password`, `config.datasource.passwordSecret`,
`config.datasource.passwordFile`, or `config.datasourceSecret` is set.

Previously, forgetting to set a password installed the chart successfully
with a known plaintext value baked into the Secret. The exporter then
failed to authenticate (silent breakage), and the well-known default
risked being copied into a real Postgres account.

After this change, the chart fails at render time with a clear message
pointing the user at the supported password sources.

- n/a

#### Special notes

This is a behavioral change: installs that today rely on the silent
fallback (and are therefore non-functional anyway) will now fail fast
instead of producing a broken Secret. Per the contributing guide this
is a backwards-incompatible change, so bumping the chart MAJOR version
(7.5.2 --> 8.0.0) and documenting the manual upgrade step in
`charts/prometheus-postgres-exporter/README.md` under `#### To 8.0.0`.

Every file in `charts/prometheus-postgres-exporter/ci/` previously
relied on the silent fallback (no password set). Each is updated to set
`config.datasource.password` so `chart-testing` continues to pass.

Verified locally:

- `helm lint` passes against every CI values file.
- `helm template ... --set config.datasource.password=testpw` renders
  the Secret with the expected b64 value.
- `helm template ...` with no password and no alternative source now
  errors with: `config.datasource.password is required when not using
  config.datasource.passwordSecret, config.datasource.passwordFile, or
  config.datasourceSecret`.

#### Checklist

- [x] Chart Version bumped,  `7.5.2` --> `8.0.0` in `charts/prometheus-postgres-exporter/Chart.yaml`

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/secrets.yaml
+++ b/charts/prometheus-postgres-exporter/templates/secrets.yaml
@@ -1,4 +1,7 @@
 {{- if not (or .Values.config.datasource.passwordSecret .Values.config.datasourceSecret .Values.config.datasource.passwordFile ) -}}
+{{- if not .Values.config.datasource.password }}
+{{- fail "config.datasource.password is required when not using config.datasource.passwordSecret, config.datasource.passwordFile, or config.datasourceSecret" }}
+{{- end }}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -8,5 +11,5 @@ metadata:
     {{- include "prometheus-postgres-exporter.labels" . | nindent 4 }}
 type: Opaque
 data:
-  data_source_password: {{ tpl (.Values.config.datasource.password | default "somepaswword") . | b64enc }}
+  data_source_password: {{ tpl .Values.config.datasource.password . | b64enc }}
 {{- end -}}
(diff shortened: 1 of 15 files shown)


charts/prometheus-postgres-exporter/templates/secrets.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if not (or .Values.config.datasource.passwordSecret .Values.config.datasourceSecret .Values.config.datasource.passwordFile ) -}}
+{{- if not .Values.config.datasource.password }}
+{{- fail "config.datasource.password is required when not using config.datasource.passwordSecret, config.datasource.passwordFile, or config.datasourceSecret" }}
+{{- end }}
```
zanhsieh:
```suggestion
{{ $_ := required "config.datasource.password is required when not using config.datasource.passwordSecret, config.datasource.passwordFile, or config.datasourceSecret" .Values.config.datasource.password }}
```

zanhsieh (changes requested):

InfraArtists:
@zanhsieh Done.

zanhsieh (dismissed):

zanhsieh (approved):

zeritti (approved):

PR #6897, opened 2026-05-06
[consul-exporter] Support using environment variables in command args

### LINKED ISSUE #6896: [prometheus-consul-exporter] Allow options to use environment variables

In our consul environment, we have a local consul endpoint available on every VM, not running in a container.
This is the desired target for the `consul-exporter`, but the Go binary does not reinterpret environment variables in options (which is a reasonable choice).

Ideally, we want to have this `values.yaml` config, and it Just Work:tm:
```yaml
consulServer: "$CONSUL_HOST:8500"
extraEnv:
  - name: CONSUL_HOST
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
```

Some values setting which allows the following change
```diff
diff --git a/k8s/sre/consul-exporter/main/consul-exporter/consul-exporter.yaml b/k8s/sre/consul-exporter/main/consul-exporter/consul-exporter.yaml
index 6cc818f..c1fd68f 100644
--- a/k8s/sre/consul-exporter/main/consul-exporter/consul-exporter.yaml
+++ b/k8s/sre/consul-exporter/main/consul-exporter/consul-exporter.yaml
@@ -94,10 +94,9 @@ spec:
         - name: prometheus-consul-exporter
           image: "prom/consul-exporter:v0.13.0"
           imagePullPolicy: IfNotPresent
-          command: ["consul_exporter"]
+          command: ["/bin/sh"]
           args:
-            - "--consul.server=$CONSUL_HOST:8500"
+            - "-c"
+            - "consul_exporter --consul.server=$CONSUL_HOST:8500"
           ports:
             - name: http
               containerPort: 9107
```

This would be an opt-in; either

 - the traditional Helm option of overriding the command (although this may be fiddly with the static first option and the need for `[/bin/sh, -c]` for the shell option)
 - Having a flag that explicitly changes it to be launched via a shell

I have confirmed the image has a functional `/bin/sh` for both of these options.

NONE

### Additional context.

I have a local branch ready to go with the second option laid out, adding a single new value which acts as an explicit toggle for this feature
```yaml
# By default, consul_exporter is run directly.
# If you want to use environment variables in `options` or `consulServer`,
# enable this option to have a shell interpret the command string before
# `consul_exporter` is executed.
invokeViaShell: false
```

And I'm raising the issue so I can grab the ID ;)

javajawa:

To increase the flexibility for options and arguments in the chart, this change adds a flag `invokeViaShell` to the values. When enabled the `consul_exporter` is not directly executed by kubernetes. Instead the same command is passed to `/bin/sh`, which will replace any environment variables.

This allows users to pull in options from secrets on the cluster; kubernetes's downwardApi systems; or other environment ephemera. The option is disabled by default.

This allows for constructs such as

```yaml
invokeViaShell: true
consulServer: "${CONSUL_HOST}:8500"
extraEnv:
  - name: CONSUL_HOST
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
```

- fixes #6896 

This change goes slightly against typical helm conventions of having values for `command` and `args` which can be overridden by the user.

As we have a always present argument specified by `consulServer`, I felt that building up the logic for a more generic solution would overall increase the complexity/confusion for this chart.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-consul-exporter/Chart.yaml
+++ b/charts/prometheus-consul-exporter/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 appVersion: "v0.13.0"
 description: A Helm chart for the Prometheus Consul Exporter
 name: prometheus-consul-exporter
-version: 1.1.1
+version: 1.2.0
 keywords:
   - metrics
   - consul
--- a/charts/prometheus-consul-exporter/templates/deployment.yaml
+++ b/charts/prometheus-consul-exporter/templates/deployment.yaml
@@ -31,9 +31,16 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.invokeViaShell }}
+          command: ["/bin/sh"]
+          args:
+            - "-c"
+            - "consul_exporter --consul.server={{ .Values.consulServer }}"
+          {{- else }}
           command: ["consul_exporter"]
           args:
             - "--consul.server={{ .Values.consulServer }}"
+          {{ end }}
             {{- range $key, $value := .Values.options }}
             - "--{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
             {{- end }}
--- a/charts/prometheus-consul-exporter/values.yaml
+++ b/charts/prometheus-consul-exporter/values.yaml
@@ -33,6 +33,12 @@ options: {}
   # no-consul.health-summary:
   # kv.filter=foobar
 
+# By default, consul_exporter is run directly.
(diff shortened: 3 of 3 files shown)


javajawa:
Posting a comment to maybe reset staleness

jkroepke:
@javajawa Did you try `consulServer: "$(CONSUL_HOST):8500"`

Ref: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

javajawa:
> Did you try `consulServer: "$(CONSUL_HOST):8500"`
> 
> Ref: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

Well, I just learnt a new thing! I knew about that for building up complex environment variables, didn't realise it also worked in the args field. I'll confirm it works as expect when I'm back at work and close this off.

PR #6900, opened 2026-05-09
[prometheus-smartctl-exporter] use serviceAccountName helper function where service account is referenced

### LINKED ISSUE #6899: [prometheus-smartctl-exporter] Setting service account name does not extend to resources when using non default names

When setting the service account name in the values of the helm chart, the daemonset and rolebinding do not use that name, it continues to use the default name generated using the fullname template of the chart.

### What's your helm version?

v3.18.6

1.36

### Which chart?

prometheus-smartctl-exporter

### What's the chart version?

0.16.0

### What happened?

Error creating: pods "smartctl-exporter-prometheus-smartctl-exporter-0-" is forbidden: error looking up service account smartctl-exporter/smartctl-exporter-prometheus-smartctl-exporter: serviceaccount "smartctl-exporter-prometheus-smartctl-exporter" not found

`      serviceAccountName: smartctl-exporter-prometheus-smartctl-exporter`

The daemonset to use the name of the service account set in the helm values

### How to reproduce it?

set the values for service account:
```
serviceAccount:
  create: true
  name: "foobar"
```

```
serviceAccount:
  create: true
  name: "foobar"
```

N/A

See daemonset template, it appears to not be using the service account name helper function which handles setting the default or name from values

https://github.com/prometheus-community/helm-charts/blob/0b19a11353af3b4cd9d7f7000c5f03c93cb5ceea/charts/prometheus-smartctl-exporter/templates/daemonset.yaml#L82

benhwebster:
When setting a service account name, resources such as the daemonset and rolebinding do not utilize it, causing problems as the chart creates the service account with the set name, but the daemonset and rolebinding continue to use the hardcoded default of the chart fullname which does not exist.

- fixes #6899

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-smartctl-exporter/Chart.yaml
+++ b/charts/prometheus-smartctl-exporter/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.16.0
+version: 0.16.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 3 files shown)


zeritti (approved):
Thank you, @benhwebster, for your PR, LGTM.

PR #6906, opened 2026-05-11
[kube-prometheus-stack] Keep KubeletDown cluster-scoped

Zakharden:

`KubeletDown` joins `kube_node_info` and kubelet `up{}` series to detect missing kubelet targets. When `defaultRules.additionalAggregationLabels` includes labels that do not exist consistently on both metric sources, such as `namespace`, the alert can fire even when kubelets are healthy.

This keeps `KubeletDown` scoped to `cluster` only, while leaving the existing additional aggregation label behavior unchanged for other generated rules.

Fixes #6872.

The generated rules file is updated through `hack/sync_prometheus_rules.py`; the follow-up commit only syncs the generated file formatting expected by CI.

#### Checklist

- [x] DCO signed
- [x] Chart Version bumped

#### Validation

- `helm unittest -f unittests/prometheus/kubelet_down_aggregation_test.yaml charts/kube-prometheus-stack`
- `./venv/bin/python ./sync_prometheus_rules.py` from `charts/kube-prometheus-stack/hack`
- `git diff --check`


Zakharden:
Hi maintainers, could someone take a look at this PR?

It keeps KubeletDown cluster-scoped so defaultRules.additionalAggregationLabels do not introduce false positives when labels like namespace are not present consistently across kube_node_info and kubelet up{}.

I synced the generated rules and added focused unittest coverage; happy to revise if there is a preferred rule-generation pattern. Thanks.

@jkroepke

Zakharden:
@jkroepke @gkarthiks @Xtigyro  @gianrubio  Rebased on the latest `main`; the PR is mergeable now and CI is green. I also updated the unit test to assert the `KubeletDown` rule by content rather than by a brittle array index. Could you please take another look?

jkroepke (approved):

PR #6951, opened 2026-05-29
[prometheus-mysql-exporter] Allow setting hostUsers on deployment

jnoordsij:

As of Kubernetes 1.36, the User Namespaces feature has graduated to stable; see also https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/.

To leverage this on pods, one needs to set the hostUsers flag on a deployment. This MR adds the ability to add such a flag to this chart. See also https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/.

Modelled after https://github.com/prometheus-community/helm-charts/pull/6734.

n/a

n/a

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-mysql-exporter/values.yaml
+++ b/charts/prometheus-mysql-exporter/values.yaml
@@ -127,6 +127,10 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 65534
 
+## Use the host's user namespace for deployment.
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/
+hostUsers: ~
+
 annotations:
   prometheus.io/scrape: "true"
   prometheus.io/path: "/metrics"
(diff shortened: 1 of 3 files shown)


charts/prometheus-mysql-exporter/values.yaml
```diff
@@ -127,6 +127,10 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 65534
 
+## Use the host's user namespace for deployment.
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
+hostUsers: ~
```
monotek, line 132: Why not just set it to false?

I'm also not sure if we should check for th k8s version?
Imho this is only available from 1.33+?

monotek (changes requested):

charts/prometheus-mysql-exporter/values.yaml
jnoordsij: I just matched #6734. This ensures only when one explicitly defines a value it is rendered, thus skipping the need for a k8s version guard (leaving that to the consumer). Note that in particular a k8s version guard might not suffice anyways, as there are additional prerequisites to using this functionality.

Note that a default of `false` would also enforce this behavior (the 'default' behavior is `true`).

jnoordsij:
FWIW I just noticed I blindly copied the ref-link but it is wrong; will update.

monotek (approved):

PR #6979, opened 2026-06-09
[prometheus-kafka-exporter] Add support for deploymentAnnotations

mortalius:
## What this PR does

Adds a new `deploymentAnnotations` value to the `prometheus-kafka-exporter` chart that renders annotations on the Deployment's `metadata.annotations`.

```yaml
deploymentAnnotations: {}
```

## Why

The chart already supports `podAnnotations` (rendered on `spec.template.metadata.annotations`) and `deploymentLabels` (rendered on the Deployment's labels), but there is currently no way to set annotations on the Deployment object itself.

This is needed for tooling that watches **Deployment-level** annotations — most notably [Stakater Reloader](https://github.com/stakater/Reloader), which triggers a rollout when a referenced `Secret`/`ConfigMap` changes. Reloader explicitly does not act on pod template annotations, so `podAnnotations` cannot be used as a workaround.

A real-world case: rotating a TLS client certificate stored in a Kubernetes Secret (e.g. issued by Vault PKI with a TTL) — without the Reloader annotation on the Deployment, the kafka-exporter pod keeps using the old cert until manual restart, and the broker eventually drops the connection.

## How

Follows the existing convention already used in several sibling charts in this repo:

- `charts/prometheus-blackbox-exporter` — `deploymentAnnotations`
- `charts/prometheus-pushgateway` — `deploymentAnnotations`
- `charts/prometheus-adapter` — `deploymentAnnotations`
- `charts/prometheus-json-exporter` — `deploymentAnnotations`
- `charts/prometheus` — `server.deploymentAnnotations`

## Backwards compatibility

Default value is `{}`, so existing installations are unaffected. Minor version bump: `3.0.1` → `3.1.0`.

## Checklist

- [x] DCO sign-off in commit
- [x] Chart `version` bumped (semver minor — additive change)
- [x] No breaking changes
- [x] Default preserves existing behavior

### DIFF
--- a/charts/prometheus-kafka-exporter/Chart.yaml
+++ b/charts/prometheus-kafka-exporter/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v1.9.0"
 description: A Helm chart to export metrics from Kafka in Prometheus format using kafka_exporter from https://github.com/danielqsj/kafka_exporter
 name: prometheus-kafka-exporter
 home: https://github.com/danielqsj/kafka_exporter
-version: 3.0.1
+version: 3.1.0
 kubeVersion: ">=1.25.0-0"
 sources:
   - https://github.com/prometheus-community/helm-charts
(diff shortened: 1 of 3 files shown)


zeritti (approved):
Thank you, @mortalius, for your PR, LGTM.

PR #6981, opened 2026-06-09
[prometheus-operator-admission-webhook]: use httpGet in livenessProbe

timonegk:

Currently, `tcpSocket` is used in the livenessProbe which leads to the error message `caller=/usr/local/go/src/net/http/server.go:3677 msg="http: TLS handshake error from ...: EOF"` every few seconds.
The reason is that tcpSocket doesn't create a valid handshake. `httpGet` should be used instead.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-operator-admission-webhook/Chart.yaml
+++ b/charts/prometheus-operator-admission-webhook/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.40.1
+version: 0.40.2
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: 0.91.0
 home: https://github.com/prometheus-operator/prometheus-operator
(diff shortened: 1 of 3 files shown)


charts/prometheus-operator-admission-webhook/Chart.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: Prometheus Operator Admission Webhook
 name: prometheus-operator-admission-webhook
-version: 0.40.1
+version: 0.40.2
```
zeritti:
```suggestion
version: 0.41.0
```

A default value is changing, needs to be a minor change.

Please, adjust the JSON schema in `values.schema.json`.

zeritti (changes requested):
Thank you, @timonegk, for your PR. Please, see my comments below.

timonegk:
Hi @zeritti, thanks for your review. I updated the version and schema.

zeritti (approved):

PR #6993, opened 2026-06-15
[kube-prometheus-stack] add per-alert runbook annotation overrides

alexsherwin:

  This PR updates `defaultRules.additionalRuleAnnotations` to be keyed by alert name so chart consumers can override any
  annotation on specific alerts, including `description`, `runbook_url`, and `summary`.

  - The change is limited to alert annotations.
  - If a consumer sets annotations for an alert in `additionalRuleAnnotations`, those values take precedence.
  - If no override is set, the upstream annotations are kept.
  - Added non-default values coverage in `ci/03-non-defaults-values.yaml`.

  Example:

  ```yaml
  defaultRules:
    additionalRuleAnnotations:
      etcdMembersDown:
        runbook_url: https://test.com/runbooks/etcd-members-down
        description: Custom description
        summary: Custom summary
```
 #### Checklist

  - [x] DCO signed
  - [x] Chart version bumped
  - [x] Title of the PR starts with chart name (e.g. [kube-prometheus-stack])

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 86.2.3
+version: 86.2.4
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.91.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 86.2.4
+version: 86.2.5
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.91.0
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/values.yaml
+++ b/charts/kube-prometheus-stack/values.yaml
@@ -226,7 +226,7 @@ defaultRules:
   ## Additional labels for PrometheusRule alerts
   additionalRuleLabels: {}
 
-  ## Additional annotations for PrometheusRule alerts
+  ## Additional annotations for specific PrometheusRule alerts by alert name
   additionalRuleAnnotations: {}
 
   ## Additional labels for specific PrometheusRule alert groups
--- a/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
+++ b/charts/kube-prometheus-stack/ci/03-non-defaults-values.yaml
@@ -3,6 +3,11 @@
 defaultRules:
   additionalRuleLabels:
     key: value
+  additionalRuleAnnotations:
+    AlertmanagerFailedReload:
+      runbook_url: https://example.runbook/runbooks/alertmanager-failed-reload
+      description: custom description for AlertmanagerFailedReload
+      summary: custom summary for AlertmanagerFailedReload
   additionalRuleGroupLabels:
     kubernetesSystem:
       key2: value2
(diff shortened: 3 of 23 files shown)

### CODE REVIEW

**Comment by alexsherwin** [comment]
image

**Review by jkroepke — COMMENTED** [review]
Hm, I'm not 100% happy yet. It feels like put something on top of another messy thing.

**Comment by alexsherwin** [comment]
Hey @jkroepke, 

Thanks for the review.

Do you have a suggestion for a cleaner way to approach this in `sync_prometheus_rules.py`?

It looks like the workflow also ran for a long time and timed out. I'm not sure if that was related to this change or a wf error. The only error I found was a Grafana sidecar `OSError: [Errno 98] Address in use`, after which it appeared to fall into a repeated loading in-cluster config loop.

File: charts/kube-prometheus-stack/values.yaml [hunk 1]
```diff
@@ -229,6 +229,9 @@ defaultRules:
   ## Additional annotations for PrometheusRule alerts
   additionalRuleAnnotations: {}
 
+  ## Additional runbook annotations for specific PrometheusRule alerts by alert name
+  additionalRuleRunbookAnnotationsByAlertName: {}
```
jkroepke:
```suggestion
  additionalRuleAnnotations: {}
```

Let's use `additionalRuleAnnotations` here as generic proposal for any kind of rule based annotations.

jkroepke (commented):
```gotpl
        {{- $ruleAnnotations := dig "PrometheusOperatorRejectedResources" (dict) .Values.defaultRules.additionalRuleAnnotations }}
        {{- $groupAnnotations := default (dict) .Values.defaultRules.additionalRuleGroupAnnotations.prometheusOperator }}
        {{- $additionalAnnotations := mergeOverwrite (dict) $groupAnnotations $ruleAnnotations }}
        {{- if $additionalAnnotations }}
        {{- toYaml $additionalAnnotations | indent 8 }}
        {{- end }}
        {{- if not (hasKey $additionalAnnotations "description") }}
        description: Prometheus operator in {{`{{`}} $labels.namespace {{`}}`}} namespace rejected {{`{{`}} printf "%0.0f" $value {{`}}`}} {{`{{`}} $labels.controller {{`}}`}}/{{`{{`}} $labels.resource {{`}}`}} resources.
        {{- end }}
        {{- if not (hasKey $additionalAnnotations "runbook_url") }}
        runbook_url: {{ .Values.defaultRules.runbookUrl }}/prometheus-operator/prometheusoperatorrejectedresources
        {{- end }}
        {{- if not (hasKey $additionalAnnotations "summary") }}
        summary: Prometheus operator not ready
        {{- end }}
```

This would a more generic approach that support rule based annotation and allow end users to overrides all properties. Otherwise, what would be the next? `additionalRuleDescriptionAnnotationsByAlertName`?

jkroepke (dismissed):

alexsherwin:
Hey @jkroepke,

Thanks for the review and approval. I’m seeing a similar workflow issue on my side.

I checked the affected templates with the current CI values and they render fine. In the CI log, the run completes the first two installs, then starts 03-non-defaults-values.yaml later in the run and gets cut off by the 6-hour workflow window. I don’t see a chart-testing failure in the log.

The workflow still looks like it is getting stuck in ct install. Any ideas?

jkroepke:
> The workflow still looks like it is getting stuck in ct install. Any ideas?

GitHub issue. I observe similar issues on other helm repos as well.

https://github.com/prometheus-community/helm-charts/pull/6990 have the same issue.

alexsherwin:
Hey @jkroepke, 

tyvm. This seems to have failed again with the same outcome. In this scenario, do I need ping for a rerun or will this be picked up at a later date?

jkroepke:
I have to manually trigger it from time to time...

charts/kube-prometheus-stack/Chart.yaml
```diff
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 86.2.3
+version: 86.2.4
```
jkroepke:
```suggestion
version: 86.2.5
```

jkroepke (approved):

PR #6995, opened 2026-06-17
[prometheus-smartctl-exporter] Add podLabels support

### LINKED ISSUE #6909: [prometheus-smartctl-exporter] Add missing podLabels

prometheus-smartctl-exporter currently has `podAnnotations`, but no ability to set labels via `podLabels`

Also missing:
* `automountServiceAccountToken`
* `securityContext`/`podsecurityContext` configuration

Ability to set the missing config options

chaining kustomize one-off mutatingAdmissionPolicy

### Additional context.

_No response_

yugstar:
## What this PR does

Adds a `podLabels` value to the **prometheus-smartctl-exporter** chart and renders it on the DaemonSet pod template, mirroring the existing `podAnnotations` handling.

## Why

The pod template only emitted selector labels, so there was no way to attach custom labels to the exporter pods — commonly needed for GitOps tooling, cost-allocation, and NetworkPolicy pod selectors. Requested in #6909.

## Notes

- `podLabels` defaults to `{}`, so the rendered DaemonSet is unchanged unless set.
- Verified with `helm template --set podLabels.team=storage` (label appears on the pod) and default render (unchanged). `helm lint` passes.
- Chart version bumped 0.16.1 → 0.16.2.

Closes #6909

### DIFF
--- a/charts/prometheus-smartctl-exporter/Chart.yaml
+++ b/charts/prometheus-smartctl-exporter/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.16.2
+version: 0.17.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/prometheus-smartctl-exporter/Chart.yaml
+++ b/charts/prometheus-smartctl-exporter/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.16.1
+version: 0.16.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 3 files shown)


charts/prometheus-smartctl-exporter/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.16.1
+version: 0.16.2
```
zeritti: Adding a new field should raise a minor version number:

```suggestion
version: 0.17.0
```

charts/prometheus-smartctl-exporter/values.yaml
```diff
@@ -150,6 +150,9 @@ priorityClassName: ""
 
 annotations: {}
 
+# Additional labels to add to the exporter pods.
```
zeritti, line 156: Please, add field `.podAnnotations` in values.yaml as it is missing.

zeritti (changes requested):

yugstar:
Thanks for the review. Done in the latest push:

- Bumped the chart version to `0.17.0`.
- Added the missing `podAnnotations` value to values.yaml. The daemonset template already referenced `.Values.podAnnotations`, so this just declares the default.

`helm lint` passes and both `podAnnotations` and `podLabels` render onto the pods.

zeritti (approved):
Thanks, LGTM

PR #6996, opened 2026-06-17
[prometheus-snmp-exporter] Add podLabels support

yugstar:
## What this PR does

Adds a `podLabels` value to the **prometheus-snmp-exporter** chart and renders it on both the Deployment and DaemonSet pod templates (the chart renders one or the other via `.Values.kind`), mirroring the existing `podAnnotations` handling.

## Why

The pod templates only emitted the chart labels, so there was no way to attach custom labels to the exporter pods — commonly needed for GitOps tooling, cost-allocation, and NetworkPolicy pod selectors.

## Notes

- `podLabels` defaults to `{}` (guarded with `{{- with }}`), so the rendered workload is unchanged unless set.
- Verified with `helm template --set podLabels.team=neteng` for both `kind: Deployment` (default) and `kind: DaemonSet` (label appears on the pod in both), and default render unchanged. `helm lint` passes.
- Chart version bumped 9.14.0 → 9.14.1.

### DIFF
--- a/charts/prometheus-snmp-exporter/Chart.yaml
+++ b/charts/prometheus-snmp-exporter/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Prometheus SNMP Exporter
 name: prometheus-snmp-exporter
-version: 9.14.0
+version: 9.14.1
 # renovate: github=prometheus/snmp_exporter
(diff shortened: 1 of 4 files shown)


walker-tom (approved):
LGTM, thanks!

PR #7022, opened 2026-06-24
[prometheus-json-exporter] Add Gateway API HTTPRoute support

somaz94:
Add Gateway API HTTPRoute support to the prometheus-json-exporter chart, mirroring the pattern already merged for the prometheus, prometheus-pushgateway, alertmanager, and prometheus-blackbox-exporter charts.

Validation:
- `helm template` renders a valid HTTPRoute with `ci/httproute-values.yaml` (parentRefs, hostnames, filters, additionalRules)
- `ct install` on a kind cluster (with Gateway API CRDs applied) installs both the default and httproute scenarios successfully

- [x] DCO signed


zanhsieh (approved):

somaz94:
Hi @zanhsieh — thanks for the approval. This one's been green and mergeable since, so just flagging it in case it slipped through. Happy to rebase or adjust anything if needed.

somaz94:
@zanhsieh thanks for the review. Green and mergeable since early August, so this looks ready whenever someone has a moment to merge it.

PR #7026, opened 2026-06-24
[prometheus-cloudwatch-exporter] Add Gateway API HTTPRoute support

somaz94:
Add Gateway API HTTPRoute support to the prometheus-cloudwatch-exporter chart, mirroring the pattern already merged for the prometheus, prometheus-pushgateway, alertmanager, prometheus-blackbox-exporter, and prometheus-json-exporter charts.

Validation:

- [x] DCO signed


somaz94:
Hi @gianrubio / @torstenwalter — gentle ping on this one.

It's the same opt-in Gateway API HTTPRoute pattern as the sibling PRs from the same batch that have since merged (#7024, #7028, #7029, #7031, #7032). Default is `httpRoute.enabled: false` with a `ci/httproute-values.yaml` test case, so existing users are unaffected. CI is green.

Happy to rebase or adjust anything if you'd prefer a different shape.

PR #7028, opened 2026-06-24
[prometheus-fastly-exporter] Add Gateway API HTTPRoute support

somaz94:
Add Gateway API HTTPRoute support to the prometheus-fastly-exporter chart, mirroring the pattern already merged for the prometheus, prometheus-pushgateway, alertmanager, prometheus-blackbox-exporter, and prometheus-json-exporter charts.

Validation:

- [x] DCO signed


arslanbekov (approved):
LGTM
Thanks for PR

PR #7030, opened 2026-06-24
[prometheus-yet-another-cloudwatch-exporter] Add Gateway API HTTPRoute support

somaz94:
Add Gateway API HTTPRoute support to the prometheus-yet-another-cloudwatch-exporter chart, mirroring the pattern already merged for the prometheus, prometheus-pushgateway, alertmanager, prometheus-blackbox-exporter, and prometheus-json-exporter charts.

Validation:

- [x] DCO signed

### DIFF
--- a/charts/prometheus-yet-another-cloudwatch-exporter/Chart.yaml
+++ b/charts/prometheus-yet-another-cloudwatch-exporter/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus-yet-another-cloudwatch-exporter
 description: Yace - Yet Another CloudWatch Exporter
 type: application
-version: 0.47.0
+version: 0.48.0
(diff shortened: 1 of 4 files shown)


somaz94:
Gentle ping @cristiangreco @tristanburgess @thomaspeitz @andriikushch — this has been open since Jun 24 without a review yet.

Everything the repo asks for is in place: Chart Testing, the Integration Tests against `ci/httproute-values.yaml`, Super Linter, Check Generated Files and Check Title all pass, DCO is signed, and the chart version is bumped (0.47.0 → 0.48.0, still ahead of `main`).

The change is opt-in and off by default — with `httpRoute.enabled: false` the rendered output is identical to today's, so it carries no risk for existing users. Happy to rebase onto `main` if that helps; the branch is a bit behind but merges cleanly.

PR #7044, opened 2026-06-29
[prometheus-pushgateway] Support web configuration from an existing Secret

### LINKED ISSUE #5979: [prometheus-pushgateway] Add an ability to configure basicAuthUsers via secrets

nikolauspschuetz:
## What this does

Closes #5979.

Adds `webConfiguration.existingSecret` to the **prometheus-pushgateway** chart so the
web configuration (including basic-auth users) can be sourced from a pre-existing
Secret instead of being kept in plaintext in `values.yaml`. This suits GitOps
workflows that manage credentials with Sealed Secrets, External Secrets, Vault, etc.,
where committing cleartext passwords is undesirable.

This mirrors the established `existingSecret` pattern used by sibling charts in this
repo (e.g. `prometheus-mongodb-exporter`): a helper resolves the Secret name to the
user-provided one when set, otherwise the chart-managed one.

### Behaviour

When `webConfiguration.existingSecret.name` is set:
- the chart-managed `web-config` Secret is **not** rendered;
- the pushgateway volume mounts the user-provided Secret (which must contain the
  `web-config.yaml` key that Pushgateway expects);
- the liveness/readiness probes fall back to a `tcpSocket` check, because the chart
  cannot read the credentials to build an authenticated HTTP probe (a `basic_auth_users`
  config would otherwise make `/-/healthy` and `/-/ready` return `401`).

The existing inline `webConfiguration.basicAuthUsers` path and the ServiceMonitor
basic-auth wiring are unchanged and fully backward compatible — when no
`existingSecret` is set, rendering is byte-for-byte identical to before.

### Files changed
- `templates/_helpers.tpl` — `webConfigurationExistingSecretName` / `webConfigurationSecretName` helpers; volume `secretName` resolution; probe `tcpSocket` fallback when an existing Secret is used.
- `templates/secret.yaml` — skip the chart-managed web-config Secret when `existingSecret` is set; the ServiceMonitor `-basic-auth` Secret is now guarded on `basicAuthUsers` (it can only be built from inline creds).
- `templates/servicemonitor.yaml` — auto-wire ServiceMonitor basic-auth only when inline `basicAuthUsers` are present.
- `values.yaml` — document `webConfiguration.existingSecret`.
- `Chart.yaml` — version bump `3.6.1` → `3.7.0` (minor; new feature) + `artifacthub.io/changes` annotation.
- `ci/web-config-existing-secret-values.yaml` — chart-testing values exercising the existingSecret path (provisions the referenced Secret via `extraManifests`).

### Testing

`helm lint` passes (default and with the new ci values file). `helm template`
verified for all three cases:

| Case | chart-managed Secret | volume `secretName` | probe |
|------|----------------------|---------------------|-------|
| default (`webConfiguration: {}`) | none | — | httpGet |
| inline `basicAuthUsers` | web-config + `-basic-auth` | `<fullname>` | httpGet + `Authorization` header |
| `existingSecret.name` set | none | `<existingSecret.name>` | tcpSocket |

Rendered output parses as valid YAML in every case. `ct lint` / `ct install` will run in CI.

### AI assistance

Tooling: this change was prepared with **Claude Code (Anthropic, Opus 4.x)**, disclosed
via the `Assisted-by:` commit trailer alongside the DCO sign-off.

- **AI role:** surveyed the existing `existingSecret` pattern across sibling charts,
  drafted the template/values/ci edits, and ran the `helm lint` / `helm template`
  render-diffs across the three cases above.
- **My role (author):** chose the design (mirror the repo's existing-Secret pattern,
  tcpSocket probe fallback), reviewed every line of the diff, verified backward
  compatibility (no diff when `existingSecret` is unset), and own this contribution
  under the DCO sign-off.

cc @zeritti @cstaud @gianrubio @jkroepke


zeritti (approved):
Thank you, @nikolauspschuetz, for your PR, LGTM.

PR #7077, opened 2026-07-07
[prometheus-node-exporter] append extraArgs to default args

### LINKED ISSUE #6831: [prometheus-node-exporter] extraArgs forces users to duplicate boilerplate

Hi,

When adding a custom arg to prometheus-node-exporter's extraArgs (ie. --some-extra-arg), the entire extraArgs list is replaced rather than merged with the chart defaults. This forces users to manually reproduce the default args in their own values, which is brittle.
The defaults defined in the chart at the time of writing are:
```
extraArgs:
  - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+|var/lib/kubelet/.+)($|/)
  - --collector.filesystem.ignored-fs-types=^(autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs)$
 ```
If a user wants to add a single flag, they must copy these defaults verbatim into their own values or silently lose them, with no indication that anything is wrong until filesystem metrics disappear.

**Expected behaviour**
Custom extraArgs entries should be appended to the defaults, not replace them. Perhaps move the above into a new `defaultArgs` list, and then include both this and any extraArgs under args: in the daemonset.yaml template.

PS. I haven't checked the other charts to see whether this issue also applies to their `extraArgs`

### What's your helm version?

v3.x

v1.34

### Which chart?

prometheus-node-exporter

### What's the chart version?

4.53

harsh543:
## Summary
- move the node-exporter filesystem ignore flags into a new `defaultArgs` value
- render container args with `defaultArgs` followed by user-provided `extraArgs`
- bump the prometheus-node-exporter chart patch version

## Why
Previously, users adding one custom `extraArgs` entry had to copy the chart defaults as well or lose the default filesystem ignore rules. Keeping defaults in `defaultArgs` and appending `extraArgs` preserves the default behavior while making custom args additive.

Fixes prometheus-community/helm-charts#6831.

## Validation
- `/tmp/codex-helm/helm template pne charts/prometheus-node-exporter`
- rendered with `extraArgs: [--collector.textfile.directory=/run/prometheus]` and confirmed default filesystem args are still present before the custom arg
- `/tmp/codex-helm/helm lint charts/prometheus-node-exporter`
- `git diff --check`

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.55.0
+version: 4.55.1
 # renovate: github=prometheus/node_exporter
 appVersion: 1.11.1
 home: https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/templates/NOTES.txt
+++ b/charts/prometheus-node-exporter/templates/NOTES.txt
@@ -14,6 +14,17 @@
   kubectl port-forward --namespace {{ template "prometheus-node-exporter.namespace" . }} $POD_NAME 9100
 {{- end }}
 
+{{- if .Values.defaultArgs }}
+
+NOTE: The chart now ships with a `defaultArgs` list containing the standard
+filesystem collector exclusion flags. If you previously set these flags via
+`extraArgs`, remove them from `extraArgs` to avoid duplicate-flag errors on
+startup:
+  --collector.filesystem.ignored-mount-points
+  --collector.filesystem.ignored-fs-types
+Set `defaultArgs: []` to opt out of the built-in defaults entirely.
+{{- end }}
+
 {{- if .Values.kubeRBACProxy.enabled}}
 
 kube-rbac-proxy endpoint protections is enabled:
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -109,7 +109,7 @@ spec:
             {{- end }}
             {{- end }}
             - --web.listen-address=[$(HOST_IP)]:{{ $servicePort }}
-            {{- with .Values.extraArgs }}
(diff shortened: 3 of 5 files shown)


harsh543:
Hey @jkroepke @zeritti @monotek 👋 — tagging you as the most active mergers on this repo.

**What this does:** Introduces `defaultArgs` — a new values key that prepends standard filesystem collector exclusion flags before `extraArgs`. The daemonset template uses `concat (.Values.defaultArgs | default list) (.Values.extraArgs | default list)` so both are always rendered in the correct order.

**Why this matters:** Right now users who want to customise `extraArgs` are forced to duplicate the default filesystem exclusion flags or lose them entirely. This is a footgun — omitting them causes node_exporter to attempt mounting `/proc`, `/sys`, `/dev` etc., which can block startup in restricted environments.

**What's included:**
- `values.yaml`: `defaultArgs` with the two standard filesystem exclusion flags; `extraArgs` stays empty by default
- `daemonset.yaml`: `concat` to merge both lists
- `NOTES.txt`: upgrade migration warning for users who previously set these flags via `extraArgs`
- `unittests/daemonset_args_test.yaml`: 4 helm-unittest cases covering default rendering, extraArgs appended, `defaultArgs: []` opt-out, and opt-out + custom extraArgs

**Upgrade safety:** `kingpin` (used by node_exporter) uses the last value of a duplicate flag — no crash, no silent data loss. But the migration note in NOTES.txt gives users explicit guidance to clean up.

All `helm lint --strict` and `helm template` runs against the 9 CI values files pass clean. The `helm unittest --strict` suite runs 4/4 green.

Fixes #6831.

monotek (changes requested):
This changes the default and cant be a non major change.
I also don't think its worth it.

PR #7083, opened 2026-07-08
[prometheus-postgres-exporter] Add podMonitor template

jcpunk:
This adds the ability to use a `podMonitor` directly rather than a `serviceMonitor`. This is most useful with multiple read-only replicas behind a single service (CNPG).

- fixes #

#### Checklist

- [x] Chart Version bumped
- [-] Add an [helm unittest](https://github.com/helm-unittest/helm-unittest) test case to cover this change.

This chart doesn't have unit tests yet.

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/podmonitor.yaml
+++ b/charts/prometheus-postgres-exporter/templates/podmonitor.yaml
@@ -1,0 +1,46 @@
+{{- if .Values.podMonitor.enabled }}
+apiVersion: {{ .Values.podMonitor.apiVersion | default "monitoring.coreos.com/v1" }}
+kind: PodMonitor
+metadata:
+{{- if .Values.podMonitor.labels }}
+  labels:
+    {{- include "prometheus-postgres-exporter.labels" . | nindent 4 }}
+{{ toYaml .Values.podMonitor.labels | indent 4}}
+{{- end }}
+  name: {{ template "prometheus-postgres-exporter.fullname" . }}
+  namespace: {{ .Values.podMonitor.namespace | default (include "prometheus-postgres-exporter.namespace" .) | quote }}
+spec:
+  podMetricsEndpoints:
+  - port: http
+{{- if .Values.podMonitor.interval }}
+    interval: {{ .Values.podMonitor.interval }}
+{{- end }}
+{{- if .Values.podMonitor.telemetryPath }}
+    path: {{ .Values.podMonitor.telemetryPath }}
+{{- end }}
+{{- if .Values.podMonitor.scheme }}
+    scheme: {{ .Values.podMonitor.scheme }}
+{{- end }}
+{{- if .Values.podMonitor.timeout }}
+    scrapeTimeout: {{ .Values.podMonitor.timeout }}
+{{- end }}
+{{- if .Values.podMonitor.metricRelabelings }}
+    metricRelabelings:
+{{ toYaml .Values.podMonitor.metricRelabelings | indent 4 }}
+{{- end }}
+{{- if .Values.podMonitor.relabelings }}
+    relabelings:
+{{ toYaml .Values.podMonitor.relabelings | nindent 4 }}
+{{- end }}
+{{- with .Values.podMonitor.tlsConfig }}
+    tlsConfig:
+{{- toYaml . | nindent 6 }}
+{{- end }}
+  jobLabel: {{ template "prometheus-postgres-exporter.fullname" . }}
+  namespaceSelector:
+    matchNames:
+    - {{ include "prometheus-postgres-exporter.namespace" . | quote }}
+  selector:
+    matchLabels:
+      {{- include "prometheus-postgres-exporter.selectorLabels" . | nindent 6 }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


charts/prometheus-postgres-exporter/templates/podmonitor.yaml
```diff
@@ -0,0 +1,46 @@
...
+{{- if .Values.podMonitor.labels }}
+  labels:
+    {{- include "prometheus-postgres-exporter.labels" . | nindent 4 }}
+{{ toYaml .Values.podMonitor.labels | indent 4}}
+{{- end }}
+  name: {{ template "prometheus-postgres-exporter.fullname" . }}
+  namespace: {{ .Values.podMonitor.namespace | default (include "prometheus-postgres-exporter.namespace" .) | quote }}
+spec:
+  podMetricsEndpoints:
+  - port: {{ .Values.service.name }}
```
zeritti: I would avoid relating the container's port name to a service port name - the former cannot currently be adjusted whilst the latter can. If a user renames the service port, the target won't be discovered as that port name won't be set in the container. What about setting the field to the literal as in the Deployment (`http`)?

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
```diff
@@ -0,0 +1,46 @@
...
+{{ toYaml .Values.podMonitor.relabelings | nindent 4 }}
+{{- end }}
+{{- with .Values.podMonitor.tlsConfig }}
+    tlsConfig:
+{{- toYaml . | nindent 6 }}
+{{- end }}
+  jobLabel: {{ template "prometheus-postgres-exporter.fullname" . }}
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace | quote }}
```
zeritti: The deployment namespace can differ from the release namespace via `namespaceOverride`. If used, the exporter won't run in the release namespace. We should probably prefer the helper template `"prometheus-postgres-exporter.namespace"`.

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
```diff
@@ -0,0 +1,46 @@
...
+{{- end }}
+{{- if .Values.podMonitor.relabelings }}
+    relabelings:
+{{ toYaml .Values.podMonitor.relabelings | nindent 4 }}
+{{- end }}
+{{- with .Values.podMonitor.tlsConfig }}
+    tlsConfig:
+{{- toYaml . | nindent 6 }}
+{{- end }}
+  jobLabel: {{ template "prometheus-postgres-exporter.fullname" . }}
```
zeritti, line 39: The chart does not set a label on the pod equal to the full name. Can we leave the default on Prometheus operator and let the user change it if so desired through `podMonitor.jobLabel` (empty string by default)?

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -70,6 +70,33 @@ serviceMonitor:
       #   databaseName: default '' (Set the database name to connect to)
       #   metricRelabelings: [] (MetricRelabelConfigs to apply to samples from the target before ingestion)
 
+podMonitor:
```
zeritti, line 73: Can we please keep the fields uncommented?

zeritti (changes requested):
Thank you, @jcpunk, for your PR. Please, see my comments below.

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
jcpunk: I set this to match the `serviceMonitor`'s behavior.  I'd be fine to have `.jobLabel` customize this, but I'd like to keep `serviceMonitor` and `podMonitor` somewhat in sync.  Perhaps a second PR that makes this an option you can tweak for both of them at once?

charts/prometheus-postgres-exporter/values.yaml
jcpunk: I set this to match `serviceMonitor`'s behavior.  Having them different feels a bit odd, but I'd be fine to uncomment `podMonitor`'s stuff by default leaving `serviceMonitor` alone.  I just want to be sure before making them different.

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
jcpunk: done

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
jcpunk: done

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
```diff
@@ -0,0 +1,46 @@
...
+  namespaceSelector:
+    matchNames:
+    - {{ template "prometheus-postgres-exporter.namespace" . | quote }}
```
zeritti: The `template` function will fail here. Please, use the `include` function.

zeritti (changes requested):

charts/prometheus-postgres-exporter/templates/podmonitor.yaml
jcpunk: oops, fixed.

zeritti (approved):

PR #7085, opened 2026-07-09
[kube-prometheus-stack] Omit Prometheus spec.shards when set to null

### LINKED ISSUE #7084: [kube-prometheus-stack] Always-rendered Prometheus spec.shards conflicts with HPA/KEDA autoscaling via the /scale subresource
### Describe the bug

The `Prometheus` custom resource template renders `spec.shards` unconditionally:

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L94

```yaml
  shards: {{ .Values.prometheus.prometheusSpec.shards }}
```

with a default of `shards: 1` in `values.yaml`.

Since prometheus-operator v0.71.0 the `Prometheus` CRD exposes a `/scale` subresource mapped to `spec.shards`, so shard count can be horizontally autoscaled by an HPA or a KEDA `ScaledObject` targeting the `Prometheus` CR (see the [shard-autoscaling proposal](https://prometheus-operator.dev/docs/proposals/accepted/shard-autoscaling/) and [prometheus-operator#5962](https://github.com/prometheus-operator/prometheus-operator/pull/5962)).

Because the chart always writes `spec.shards`, any GitOps/Helm reconcile (e.g. Argo CD, Flux, or a plain `helm upgrade`) continuously reverts the value the autoscaler sets, so the chart and the autoscaler fight over the field. Today the only workarounds are external, e.g. Argo CD `ignoreDifferences` on `/spec/shards`.

### Expected behavior

The default (`shards: 1`) is semantically identical to leaving `spec.shards` unset — prometheus-operator treats a nil/absent `shards` as 1. So when `shards` is `1` (the default) the field can be omitted from the rendered manifest, letting an external autoscaler own `spec.shards` via the `/scale` subresource. Any explicit value greater than 1 should still be rendered so existing static-sharding users are unaffected.

### Proposed fix

Only render `spec.shards` when it is greater than 1:

```yaml
{{- if gt (int .Values.prometheus.prometheusSpec.shards) 1 }}
  shards: {{ .Values.prometheus.prometheusSpec.shards }}
{{- end }}
```

This is backward compatible:
- `shards: 1` (default) or unset -> field omitted -> operator still runs 1 shard, autoscalers can own it.
- `shards: >1` -> rendered exactly as before (static sharding unchanged).

I'll open a PR implementing this.

### Version of Helm and Kubernetes

kube-prometheus-stack chart `87.12.1`.

guyeisenbach:

The `Prometheus` custom resource template renders `spec.shards` unconditionally:

```yaml
  shards: {{ .Values.prometheus.prometheusSpec.shards }}
```

with a default of `shards: 1`.

Since prometheus-operator v0.71.0 the `Prometheus` CRD exposes a `/scale` subresource mapped to `spec.shards`, so shard count can be horizontally autoscaled by an HPA or a KEDA `ScaledObject` targeting the `Prometheus` CR (see the [shard-autoscaling proposal](https://prometheus-operator.dev/docs/proposals/accepted/shard-autoscaling/) and [prometheus-operator#5962](https://github.com/prometheus-operator/prometheus-operator/pull/5962)).

Because the chart always writes `spec.shards`, any GitOps/Helm reconcile (Argo CD, Flux, plain `helm upgrade`) continuously reverts the value the autoscaler sets, so the chart and the autoscaler fight over the field. The only workarounds today are external, e.g. Argo CD `ignoreDifferences` on `/spec/shards`.

A value of `1` is semantically identical to leaving `spec.shards` unset — prometheus-operator treats a nil/absent `shards` as 1. So this PR only renders the field when it is greater than 1:

```yaml
{{- if gt (int .Values.prometheus.prometheusSpec.shards) 1 }}
  shards: {{ .Values.prometheus.prometheusSpec.shards }}
{{- end }}
```

This is backward compatible:
- `shards: 1` (default) or unset -> field omitted -> operator still runs 1 shard, and an external autoscaler can own `spec.shards` via the `/scale` subresource.
- `shards: >1` -> rendered exactly as before (static sharding unchanged).

## Changes

- `templates/prometheus/prometheus.yaml`: guard `spec.shards` so it is only rendered when greater than 1.
- `values.yaml`: document that the default `1` is not rendered so an external autoscaler (HPA/KEDA) can own `spec.shards`.
- `Chart.yaml`: bump chart version `87.12.1` -> `87.12.2`.

## Verification

shards=1 (default) -> spec.shards omitted; shards=null -> omitted; shards=3 -> shards: 3; helm lint passes.

Closes #7084

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 87.12.1
+version: 87.12.2
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.92.1
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -91,7 +91,7 @@ spec:
 {{- end }}
   paused: {{ .Values.prometheus.prometheusSpec.paused }}
   replicas: {{ .Values.prometheus.prometheusSpec.replicas }}
-{{- if gt (int .Values.prometheus.prometheusSpec.shards) 1 }}
+{{- if not (kindIs "invalid" .Values.prometheus.prometheusSpec.shards) }}
   shards: {{ .Values.prometheus.prometheusSpec.shards }}
 {{- end }}
   logLevel:  {{ .Values.prometheus.prometheusSpec.logLevel | quote }}
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -91,7 +91,9 @@ spec:
 {{- end }}
   paused: {{ .Values.prometheus.prometheusSpec.paused }}
   replicas: {{ .Values.prometheus.prometheusSpec.replicas }}
+{{- if gt (int .Values.prometheus.prometheusSpec.shards) 1 }}
   shards: {{ .Values.prometheus.prometheusSpec.shards }}
+{{- end }}
   logLevel:  {{ .Values.prometheus.prometheusSpec.logLevel | quote }}
   logFormat:  {{ .Values.prometheus.prometheusSpec.logFormat }}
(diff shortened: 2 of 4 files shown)


jkroepke:
I dont like such magic.

The only valid input would be `shards: null` to omit that value.

jkroepke:
Please also add an helm unit test case of this change, thanks.

Normally, this is Part of the PR Checklist Template.

jkroepke (approved):

PR #7101, opened 2026-07-11
[kube-prometheus-stack] fix: render empty ruleSelector when disabled with nil values

### LINKED ISSUE #7092: [prometheus-kube-stack] Specifying prometheus.prometheusSpec.ruleSelector causes defaultRules configMap to be created empty

This is the same as #1206, I managed to reproduce it in a newer version of the chart. In addition, I noticed the bug also happens with `matchExpressions`.

### What's your helm version?

N/A, I use Argo v3.2.6

Client Version: v1.32.11 Kustomize Version: v5.5.0 Server Version: v1.32.11

### Which chart?

kube-prometheus-stack

### What's the chart version?

81.2.2

### What happened?

On the rule selection side, I set `ruleNamespaceSelector` to get all rules from specific namespaces, and subsequently disabled the `ruleSelector` and deployed the chart on the `prometheus` namespace.

I expected the selector to match all rules (defined by default by the chart), but instead the **rulefiles-0** configmap is emptied.

### How to reproduce it?

Deploy the chart with the changed values provided.

```yaml
prometheus:
  prometheusSpec:
    ruleSelector:
    ruleSelectorNilUsesHelmValues: false
    ruleNamespaceSelector:
      matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: In
          values:
            - prometheus
            - foo
```

The equivalent in Helm would be:
`helm install prometheus prometheus-community/kube-prometheus-stack --version 81.2.2 --values values.yaml`

The deployment of the chart itself is successful.

I see than in the generated `Prometheus` CRD, when the `ruleSelector` is set to nil, it doesn't appear at all on the spec,  _unlike_ other selectors set to nil, such as `probeSelector` or `serviceMonitorSelector`.

Simply not specifying the `ruleSelector` sets some default value, even with `ruleSelectorNilUsesHelmValues: false`.

artjom132007:
This PR fixes an issue where the `ruleSelector` field completely disappears from the generated Prometheus CRD manifest when `ruleSelectorNilUsesHelmValues` is set to `false` and `ruleSelector` is omitted or set to `null`. 

In Prometheus Operator, the absence of `ruleSelector` prevents any rules from being discovered, resulting in an empty `rulefiles-0` configmap. The expected behavior is to render `ruleSelector: {}` so that all rules are matched.

fixes #7092

The fix replaces a brittle `kindIs "invalid"` check with a robust `else` fallback block, which guarantees that `ruleSelector: {}` will be rendered when no specific selectors or default helm values are used.

#### Checklist
- [x] DCO signed
- [ ] Chart Version bumped
- [ ] Add an helm unittest test case to cover this change.

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
@@ -279,7 +279,7 @@ spec:
   ruleSelector:
     matchLabels:
       release: {{ $.Release.Name | quote }}
-{{ else if not (kindIs "invalid" .Values.prometheus.prometheusSpec.ruleSelector) }}
+{{- else }}
   ruleSelector: {}
 {{- end }}
 {{- end }}


jkroepke:
> `ruleSelector` is omitted or set to `null`.

If  `ruleSelector`  is set to `null`, its expected to omit `ruleSelector`  in the config. Set `ruleSelector` to `{}`.

artjom132007:
Thanks for the review, @jkroepke! That makes total sense. 

If omitting the field on `null` is the intended design for this chart, then issue #7092 is indeed a user configuration error rather than a bug in the template. 

I will close this PR. Thanks for your time and guidance!

PR #7118, opened 2026-07-17
[kube-prometheus-stack] scrape kube-scheduler resource metrics

### LINKED ISSUE #6334: [kube-prometheus-stack] kube scheduler metrics scraping "additionalPath" key

maxbischoff:

Introduce a new feature flag `kubeScheduler.serviceMonitor.resource.enabled` and associated configs to scrape [kube-scheduler resource metrics](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#kube-scheduler-metrics) such as `kube_pod_resource_requests` and `kube_pod_resource_limit`. These metrics expose the combined result of container-level and [pod-level resources](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/) and are therefore useful to determine if the resource usage of one's Pod / container matches the resource requests / limits. This is especially necessary, since kube-state-metrics do not plan to expose a pod-level resource metric, instead preferring to refer to the kube-scheduler's metrics (https://github.com/kubernetes/kube-state-metrics/pull/3027).

- fixes #6334 
- fixes https://github.com/prometheus-community/helm-charts/pull/6718

In #6718 a fix for #6334 was already started, but after a comment by @jkroepke requesting changes has seemingly been abandoned. I agree with @jkroepke's  suggestion of mirroring the `kubelet.serviceMonitor.resource` feature, but slightly deviated from it by introducing an `enabled` subKey and having `interval`, `metricRelabelings` and `relabelings` reside under this subKey instead of using the flat structure in kubelet (where `kubelet.serviceMonitor.resource` is the feature flag and `kubelet.serviceMonitor.resourceInterval`, `kubelet.serviceMonitor.resourceMetricRelabelings` and `kubelet.serviceMonitor.resourceRelabelings` configure those aspects of the endpoint).

I would also suggest to set this feature to be enabled by default (but have chosen to go with @jkroepke 's "disabled default"-Suggestion), since these metrics will be very useful for users that want to use pod-level resources, which are also enabled by default as of kubernetes 1.34.

Finally, I also modified the Prometheus clusterrole to allow enable scraping the path based on the feature flag. If both `kubelet.serviceMonitor.resource` and `kubeScheduler.serviceMonitor.resource.enabled` are true, this will render two identical array entries with `/metrics/resources`, which looks a bit ugly but doesn't hurt and keeps the `if` conditions in `clusterrole.yaml` simpler. If you want, I can also deduplicate those paths though, by checking if both fields are set to true, and if `kubelet.serviceMonitor.resourcePath` equals `/metrics/resources` (since that field is configurable).

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 87.16.1
+version: 87.17.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.92.1
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/templates/exporters/kube-scheduler/servicemonitor.yaml
+++ b/charts/kube-prometheus-stack/templates/exporters/kube-scheduler/servicemonitor.yaml
@@ -60,4 +60,34 @@ spec:
     relabelings:
 {{ tpl (toYaml .Values.kubeScheduler.serviceMonitor.relabelings | indent 4) . }}
 {{- end }}
+{{- if .Values.kubeScheduler.serviceMonitor.resource.enabled }}
+  - port: {{ .Values.kubeScheduler.serviceMonitor.port }}
+    path: "/metrics/resources"
+    {{- if .Values.kubeScheduler.serviceMonitor.resource.interval }}
+    interval: {{ .Values.kubeScheduler.serviceMonitor.resource.interval }}
+    {{- end }}
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    {{- if .Values.kubeScheduler.serviceMonitor.proxyUrl }}
+    proxyUrl: {{ .Values.kubeScheduler.serviceMonitor.proxyUrl}}
+    {{- end }}
+    {{- if eq (include "kube-prometheus-stack.kubeScheduler.insecureScrape" (list . false true .Values.kubeScheduler.serviceMonitor.https )) "true" }}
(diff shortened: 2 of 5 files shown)


jkroepke (approved):
LGTM

Issue #7071, opened 2026-07-05
[prometheus-kube-stack] crds sub chart labels not being applied consistently


mergwyn:
### Describe the bug a clear and concise description of what the bug is.

I needed to a a label to the update-crds job and service account objects and attempted to do this by adding the following snippet to my helm chart:
```
crds:
  enabled: true
  upgradeJob:
    enabled: true
    labels:
      velero.io/exclude-from-backup: "true"
``` 
After inspecting the chart, I expected that this would result in these labels being added to the clusterrrole, clusterrolebinding, job, and service account objects, but what I found was that they were only applied to the job.

### What's your helm version?

version.BuildInfo{Version:"v4.2.0", GitCommit:"06468084e85c244c712834933d25ea232a4c2093", GitTreeState:"clean", GoVersion:"go1.26.3", KubeClientVersion:"v1.36"}

### What's your kubectl version?

 kubectl version Client Version: v1.36.1 Kustomize Version: v5.8.1 Server Version: v1.35.5+k3s1

### Which chart?

kube-prometheus-stack

### What's the chart version?

86.3.2

### What happened?

I was able to add labels to the serviceaccount using:
```
crds:
    serviceAccount:
      labels:
        velero.io/exclude-from-backup: "true"
```

Afer loking at the crds chart templates, there were multiple references that look like the labels should have been included:
```
$ grep labels *
clusterrole.yaml:  labels:
clusterrole.yaml:    {{- include "kube-prometheus-stack.crd.upgradeJob.labels" . | nindent 4 }}
clusterrolebinding.yaml:  labels:
clusterrolebinding.yaml:    {{- include "kube-prometheus-stack.crd.upgradeJob.labels" . | nindent 4 }}
crds.yaml:  labels:
crds.yaml:    {{- include "kube-prometheus-stack.crd.upgradeJob.labels" . | nindent 4 }}
job.yaml:  labels:
job.yaml:    {{- include "kube-prometheus-stack.crd.upgradeJob.labels" . | nindent 4 }}
job.yaml:    {{- with .Values.upgradeJob.labels }}
job.yaml:      labels:
serviceaccount.yaml:  labels:
serviceaccount.yaml:    {{- include "kube-prometheus-stack.crd.upgradeJob.labels" . | nindent 4 }}
serviceaccount.yaml:    {{- with .Values.upgradeJob.serviceAccount.labels }}
```
[...]

PR #7126, opened 2026-07-21
[kube-prometheus-stack] apply upgradeJob.labels to all CRD upgrade resources

dpacgdm:
## Summary
- Render `crds.upgradeJob.labels` on ClusterRole, ClusterRoleBinding, ConfigMap, and ServiceAccount (previously only the Job)
- Fold those labels into the existing `kube-prometheus-stack.crd.upgradeJob.labels` helper so each upgrade resource includes them from one place
- Clarify the values.yaml comment so the label scope matches the templates

Addresses prometheus-community/helm-charts#7071

## Test plan
- [ ] `helm template` with `crds.upgradeJob.enabled=true` and a custom `upgradeJob.labels` entry; confirm the label appears on Job + RBAC + SA + ConfigMap

### DIFF
--- a/charts/kube-prometheus-stack/charts/crds/templates/_helpers.tpl
+++ b/charts/kube-prometheus-stack/charts/crds/templates/_helpers.tpl
@@ -8,6 +8,9 @@
 app: {{ template "kube-prometheus-stack.name" . }}-operator
 app.kubernetes.io/name: {{ template "kube-prometheus-stack.name" . }}-prometheus-operator
 app.kubernetes.io/component: crds-upgrade
+{{- with .Values.upgradeJob.labels }}
+{{ toYaml . }}
+{{- end }}
 {{- end -}}
 
 {{/* Create the name of crd.upgradeJob service account to use */}}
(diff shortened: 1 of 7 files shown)


jkroepke:
Does it make sense to put them into `kube-prometheus-stack.crd.upgradeJob.labels` helper?

dpacgdm:
Yes — that would be cleaner. I'll move the shared label set into a `kube-prometheus-stack.crd.upgradeJob.labels` helper and have the Job / ClusterRole / ClusterRoleBinding / ServiceAccount / ConfigMap templates include it.

charts/kube-prometheus-stack/charts/crds/templates/_helpers.tpl
```diff
@@ -8,6 +8,9 @@
 app: {{ template "kube-prometheus-stack.name" . }}-operator
 app.kubernetes.io/name: {{ template "kube-prometheus-stack.name" . }}-prometheus-operator
 app.kubernetes.io/component: crds-upgrade
+{{- with .Values.upgradeJob.labels }}
+{{ toYaml . }}
```
jkroepke, line 12:
```suggestion
{{- toYaml . }}
```

jkroepke (commented):
Could you please add an [helm unittest](https://github.com/helm-unittest/helm-unittest) test case to cover this change. - the PR checklist request this?

DrFaust92:
Rendering checks out — with `crds.upgradeJob.labels` set, the label lands on all five upgrade resources (Job, ClusterRole, ClusterRoleBinding, ConfigMap, ServiceAccount) and the YAML is valid.

Two things before this can merge: the branch needs a rebase on main, and `Chart.yaml` needs a version bump, which CI enforces for any chart change. #7271 arrived later for the same issue but renders invalid YAML, so this is the one worth landing.

PR #7176, opened 2026-08-09
[kube-state-metrics] Fix autosharding configuration service

Alja9:
Enforce `ClusterIP` for autosharding, avoid headless service error.

Autosharding get a Kubernetes error, example:
```
level=WARN msg="upgrade failed" name=my-kube-state-metrics error="Service \"my-kube-state-metrics\" is invalid: spec.clusterIPs[0]: Invalid value: \"None\": may not be set to 'None' for NodePort services"
Error: UPGRADE FAILED: Service "my-kube-state-metrics" is invalid: spec.clusterIPs[0]: Invalid value: "None": may not be set to 'None' for NodePort services
```
Instead of giving an error, how about to enforce it to the `ClusterIP` (auto-corrected, no Kubernetes error).

- fixes #

#### Checklist

- [x] Chart Version bumped


sebastiangaiser (changes requested):
Nice fix, this resolves the invalid Service manifest you get with autosharding plus a NodePort/LoadBalancer service.type (headless + NodePort can't coexist), and it also corrects the metrics-port nodePort guard that was checking service.nodePort instead of selfMonitor.telemetryNodePort. A few small things:
- Two of the unit tests are named "fails when autosharding enabled with NodePort/LoadBalancer service type", but they don't fail, they assert the service is forced to headless ClusterIP. Could you please rename them to match (e.g. "forces headless ClusterIP even when service.type is NodePort")?
- The test file is missing a trailing newline.
- Needs a rebase on main (currently conflicting) and a re-bump to 8.3.1.

Alja9:
Hi @sebastiangaiser 
I've adjusted based on your review, thank you!

sebastiangaiser (approved):

sebastiangaiser:
Thanks for the quick fix and the fast iteration on the feedback! Appreciate your contributions.

PR #7216, opened 2026-08-26
[kube-prometheus-stack] Make cert-manager private key configurable

maarous:
Summary
Added a configurable `privateKey` option for the cert-manager admission webhook Certificate. Users can now configure the private key algorithm and size, such as `ECDSA` with a `384`-bit key. The chart version was bumped to `88.5.5`.

Motivation
The predefined issuer enforces security policies that reject the default RSA 2048-bit certificate configuration. Making the private key settings configurable allows users to select an algorithm and key size accepted by their issuer while preserving cert-manager’s defaults when no override is provided.


jkroepke (dismissed):
LGTM

jkroepke (approved):

Issue #7198, opened 2026-08-18
[kube-state-metrics] Implement v2.20.0 changes


sebastiangaiser:
### Which chart?

kube-state-metrics

### Is your feature request related to a problem ?

Kube-state-metrics was updated to [v2.20.0](https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.20.0) in https://github.com/prometheus-community/helm-charts/pull/7197 which contains changes e.g. regarding RBAC.

### Describe the solution you'd like.

Implement the updates.

### Describe alternatives you've considered.

-

U-S-jun:
I'd like to work on the admission policy collector/RBAC part of this issue from @U-S-jun's fork.

My proposed scope is to add collector-conditional list/watch permissions for mutatingadmissionpolicies, mutatingadmissionpolicybindings, validatingadmissionpolicies, and validatingadmissionpolicybindings, with configuration examples and helm-unittest coverage. I plan to preserve the current collector defaults so existing installations on older Kubernetes versions do not start querying newer APIs unexpectedly.

I'll check the rendered resources and permissions, run the chart tests, and report the validation results in the PR. Please let me know if you prefer a different scope or if someone is already working on this.

sebastiangaiser:
Closed as completed in https://github.com/prometheus-community/helm-charts/pull/7218

PR #7218, opened 2026-08-27
[kube-state-metrics] Add RBAC for admission policy collectors

U-S-jun:

kube-state-metrics v2.20.0 adds four admission policy collectors, but enabling them through `collectors` or `collectorsExtra` currently does not generate the required RBAC permissions.

This PR adds conditional `list`/`watch` rules for the admission policy and binding resources in `admissionregistration.k8s.io`, with documentation and regression tests. Default collectors remain unchanged.

Refs #7198 (admission policy collector/RBAC support only).

These collectors require kube-state-metrics v2.20.0+, with Kubernetes v1.30+ for validating policies and v1.36+ for mutating policies. Keeping them opt-in preserves compatibility with older clusters and image overrides.

Local checks with Helm v4.2.4 and helm-unittest v1.1.2:

- 39 unit tests passed; five new cases fail against the original RBAC template.
- `helm lint --strict` passed for default values, all four CI values files, and both documented examples.
- `helm template` checks passed for both examples. Default output and an `image.tag=v2.19.1` override are unchanged apart from chart version labels.

Kubernetes installation tests were not run locally because Docker Desktop could not start.

AI disclosure: OpenAI Codex prepared the changes and ran the local checks.

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[kube-state-metrics]`)

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - prometheus
   - kubernetes
 type: application
-version: 8.4.0
+version: 8.4.1
 # renovate: github-releases=kubernetes/kube-state-metrics
 appVersion: 2.20.0
 home: https://github.com/kubernetes/kube-state-metrics/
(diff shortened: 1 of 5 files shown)


sebastiangaiser (approved):
Thanks! Confirming the opt-in default is the right call. These collectors are default-on in kube-state-metrics v2.20.0, but the chart's `kubeVersion` floor is `^1.25.0-0`, while `validatingadmissionpolicies` needs Kubernetes 1.30+ and `mutatingadmissionpolicies` needs 1.36+.

PR #7219, opened 2026-08-27
[kube-prometheus-stack] Document Grafana dashboard folder options

VERNIERELoic:

`grafana.sidecar.dashboards` in this chart is passed straight to the Grafana subchart, which supports grouping dashboards into Grafana folders via `folderAnnotation` + `provider.foldersFromFilesStructure`. Neither key appears in this chart's `values.yaml`, so users have no way to discover the feature without digging into the Grafana chart.

This adds both as commented examples next to the keys they belong to:

```yaml
grafana:
  sidecar:
    dashboards:
      folderAnnotation: grafana_folder
      provider:
        foldersFromFilesStructure: true
```

N/A

Comments only — no change to rendered output with default values. Verified that setting both values renders the expected `FOLDER_ANNOTATION` env var on the dashboard sidecar and `options.foldersFromFilesStructure: true` in the dashboard provider configmap.

#### Checklist

- [x] DCO signed
- [x] Chart Version bumped (88.5.4 -> 88.5.5)
- [x] Variables are documented in `values.yaml` (this chart does not keep a values table in README.md)

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 88.5.4
+version: 88.5.5
 # renovate: github=prometheus-operator/prometheus-operator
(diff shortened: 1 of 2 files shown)


jkroepke (approved):
Thanks

PR #7221, opened 2026-08-27
[kube-prometheus-stack] Add HTTPRoute timeout support

### LINKED ISSUE #7078: [kube-prometheus-stack] HTTPRoute adding timeout to the backendRefs
### Which chart?

kube-prometheus-stack

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/route.yaml 

would it be please possible to add timeout section to the HTTPRoute CRD object ?
Many thanks. 

be possible to specify this
```
    timeouts:
      request: 120s
```

N/A

### Additional context.

sometimes it is necessary to increase timeout when reverse proxy cuts the connection.

U-S-jun:

Adds optional `request` and `backendRequest` timeouts to HTTPRoute rules for Prometheus, Alertmanager and ThanosRuler, including the existing Prometheus and Alertmanager per-replica routes.

Defaults remain unchanged. Timeouts only apply to generated HTTPRoute backend rules; HTTPS redirects, other route kinds and user-provided `additionalRules` are preserved. Includes documentation, CI values and regression tests.

Fixes #7078

Requires Gateway API CRDs and a controller that support HTTPRoute timeouts. The fields are available in the Standard channel from Gateway API v1.2.0.

Local checks with Helm v4.2.4 and helm-unittest v1.1.2:

- 107 unit tests passed. Five regression cases fail against the original helper.
- `helm lint --strict` passed for defaults, all five CI values files and the README example.
- Default and original CI values render identically apart from chart version labels with Kubernetes capabilities set to v1.25.0 and v1.36.0. The updated Gateway CI values and README example also render the expected timeouts.

Local Kubernetes installation tests were not run because Docker is unavailable. Actual Gateway request timeout behavior was not tested.

AI assistance: OpenAI Codex prepared the changes and ran the local checks.

#### Checklist

- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[kube-prometheus-stack]`)

### DIFF
--- a/charts/kube-prometheus-stack/templates/_helpers.tpl
+++ b/charts/kube-prometheus-stack/templates/_helpers.tpl
@@ -468,6 +468,12 @@ Expects a dict: { "context": $, "route": <route values>, "serviceName": <string>
   matches:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- if eq ($route.kind | default "HTTPRoute") "HTTPRoute" }}
+  {{- with $route.timeouts }}
+  timeouts:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- end }}
   {{- with $route.sessionPersistence }}
   sessionPersistence:
     {{- toYaml . | nindent 4 }}
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 88.5.4
+version: 88.6.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.93.1
 kubeVersion: ">=1.25.0-0"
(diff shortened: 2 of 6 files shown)


charts/kube-prometheus-stack/templates/_helpers.tpl
```diff
@@ -468,6 +468,12 @@ Expects a dict: { "context": $, "route": <route values>, "serviceName": <string>
   matches:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- if eq ($route.kind | default "HTTPRoute") "HTTPRoute" }}
```
jkroepke, line 471: Can you explain, why this is needed?

U-S-jun: I added this as a defensive check because the helper also renders GRPCRoute and TCPRoute, whose rules do not define `timeouts`. If non-empty timeout values are retained after changing the route kind or reusing shared values, they would otherwise be added to the generated backend rule. Depending on field validation, the manifest could be rejected or the unsupported field dropped.

With `timeouts: {}` by default, the inner `with` already omits the field. Since `filters`, `matches` and `sessionPersistence` are passed through without kind checks, would you prefer that I remove this check for consistency?

jkroepke: I just want to know the reason, but it sounds good.

jkroepke (approved):

PR #7229, opened 2026-09-02
docs: Grafana PVC storageClassName is immutable after bind

### LINKED ISSUE #7185: [kube-prometheus-stack] Document that Grafana (and similar) PVC storageClassName is immutable after bind — Helm upgrades fail when values change it
### Is your feature request related to a problem?

Yes — a documentation / operator-experience gap around **immutable PVC fields**.

When `kube-prometheus-stack` is installed with Grafana persistence enabled and a `storageClassName` set, a later Helm upgrade that changes `grafana.persistence.storageClassName` (or equivalent nested values) fails because Kubernetes forbids changing `spec.storageClassName` on a **bound** PersistentVolumeClaim.

This is expected Kubernetes behavior, but the chart/values do not make the immutability constraint obvious. GitOps controllers (Flux `HelmRelease`, Argo CD, etc.) will retry the same failed upgrade continuously, producing noisy alerts and a stuck release.

The same class of failure can affect other chart-managed PVCs if their storage class (or other immutable PVC spec fields) is changed via values after first bind (e.g. Alertmanager / Prometheus volume claim templates depending on how they are managed).

#### Example failure (generic)

```text
Error: UPGRADE FAILED: cannot patch "RELEASE-NAME-grafana" with kind PersistentVolumeClaim:
  PersistentVolumeClaim "RELEASE-NAME-grafana" is invalid:
  spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
  (field typically: storageClassName)
```

With server-side apply / SSA-style controllers the failure is the same underlying API rejection: **bound PVC `spec.storageClassName` is immutable**.

#### Minimal reproduction (generic)

1. Install chart with Grafana persistence and a storage class, e.g.:

```yaml
grafana:
  persistence:
    enabled: true
    storageClassName: storage-class-a
    size: 10Gi
```

2. Confirm the Grafana PVC is Bound.

3. Change values only for the storage class:

```yaml
grafana:
  persistence:
    enabled: true
    storageClassName: storage-class-b
    size: 10Gi
```

4. Run `helm upgrade` (or let Flux/Argo reconcile). Upgrade fails with the immutable-field error above.

Any pair of distinct StorageClasses reproduces this (e.g. two CSI classes used for different rebuild/replication policies). The specific names do not matter — only that the desired `storageClassName` differs from the bound PVC.

### Describe the solution you'd like

1. **Document clearly** (README and/or `values.yaml` comments near `grafana.persistence.*`, and analogous persistence / `storageSpec` / `volumeClaimTemplate` knobs) that:
   - `storageClassName` (and most other PVC `spec` fields except storage size expansion where supported) **must not be changed after the PVC is bound**.
   - Changing it in Helm values will **not** migrate data and will cause **upgrade failure**.
   - Intentional storage-class migration requires an out-of-band procedure (new PVC, data copy, cutover) — not a simple values edit.

2. **Optional operator notes** for GitOps users (Flux `HelmRelease` / Argo CD): a values drift on `storageClassName` will loop-fail until values are restored to match the live PVC or the PVC is deliberately recreated.

3. **Nice-to-have (non-blocking)**:
   - Cross-link to Grafana subchart persistence docs if that is the source of the PVC template.
   - A short “storage migration” note: keep the original class in values; provision a new claim under a new name/class; copy data; switch; delete old claim — or accept data loss and delete/recreate the PVC with Helm ownership understood.
   - Consider whether chart docs should call out other immutable fields (`accessModes`, `volumeMode`, `selector`, etc.) the same way.

This issue is primarily a **docs / guidance** request. Chart behavior matching the Kubernetes API is fine; the gap is discoverability and GitOps failure mode.

### Describe alternatives you've considered

- **Leave as-is**: operators rediscover Kubernetes PVC immutability only after production upgrade failures and alert storms.
- **Helm hooks / pre-upgrade checks**: possible but heavier; documentation alone would already prevent most accidental value flips.
- **Ignore storageClassName on upgrade if PVC exists**: would hide misconfiguration and is not a clean Helm pattern; not requested here.
- **Rely solely on upstream grafana chart docs**: kube-prometheus-stack users often only read this chart’s values/README; a local note is still warranted.

### Additional context

**Related (not duplicates):**

- #3570 — Grafana PVC upgrade failure (`spec is immutable after creation…`), different trigger (volumeName / replace path), closed/stale help request.
- #4579 — migrating from emptyDir to StorageClass (data migration question).
- #5128 — wrong StorageClass applied at create time (values not honored), not post-bind change.

**Request summary**

| Item | Ask |
|------|-----|
| Primary | Document: do not change `storageClassName` (and other immutable PVC fields) after install/bind |
| Secondary | Note GitOps retry/alert impact when values diverge from live PVC |
| Optional | Brief migration guidance / cross-links |

Happy to send a docs PR if maintainers agree on preferred location (README vs values comments vs both).

RaviTharuma:
Fixes #7185. Document that bound PVC storageClassName cannot change on Helm upgrade.

### DIFF
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -108,6 +108,28 @@ helm show values oci://ghcr.io/prometheus-community/charts/kube-prometheus-stack
 
 You may also `helm show values` on this chart's [dependencies](#dependencies) for additional options.
 
+### PersistentVolumeClaim `storageClassName` is immutable after bind
+
+Grafana persistence (`grafana.persistence.*`) and other chart-managed PVCs follow
+Kubernetes PVC immutability: once a claim is **Bound**, `spec.storageClassName`
+(and most other `spec` fields besides storage size expansion where supported)
+cannot be changed. A Helm upgrade that only changes
+`grafana.persistence.storageClassName` fails with:
+
+```text
+PersistentVolumeClaim is invalid: spec: Forbidden: spec is immutable after creation
+except resources.requests for bound claims
+```
+
+GitOps controllers (Flux HelmRelease, Argo CD) will retry the same rejected
+upgrade until values match the bound PVC. This is expected Kubernetes behavior,
+not a chart bug.
+
+To move Grafana (or a similar PVC) to another StorageClass: snapshot/backup,
+delete the PVC (and usually the Grafana Pod), then upgrade with the new class so
+a new claim can bind. The same class of failure can affect Alertmanager /
+Prometheus volume claim templates if those fields change after first bind.
+
 For templated Grafana datasource definitions (e.g. when using Helm flow control), use `grafana.additionalDataSourcesString`, which is rendered via `tpl`.
 
 ### HTTPRoute timeouts


jkroepke:
Thanks for the PR. However, the chart README is not a good place for document standard kubernetes behavior. I'm going to close instead merge that.

PR #7234, opened 2026-09-04
[kube-prometheus-stack] Set a default commonName for prometheus cert-…

maarous:
## Summary

Set a default `commonName` in the Prometheus Operator cert-manager `Certificate`, equal to the Prometheus Operator service name.

## Motivation

Some PKI certificate issuers require the `commonName` field to be set explicitly. Without it, the certificate request fails during signing.

For example, when using a Vault-backed issuer, the certificate request fails with the following error:

> Warning  Failed     44s   cert-manager-certificates-issuing          The certificate request has failed to complete and will be retried: Vault failed to sign certificate: failed to sign certificate by vault: Error making API request.
> Code: 400. Errors:
> * the common_name field is required, or must be provided in a CSR with "use_csr_common_name" set to true, unless "require_cn" is set to false

Setting the `commonName` to the Prometheus Operator service name ensures the certificate request satisfies these issuers and completes successfully.

### DIFF
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 89.2.0
+version: 89.2.1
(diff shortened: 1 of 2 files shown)


jkroepke (approved):
LGTM

PR #7238, opened 2026-09-05
[kube-prometheus-stack] Add secret-based auth to control-plane ServiceMonitors

### LINKED ISSUE #7230: [kube-prometheus-stack] Built-in ServiceMonitors use deprecated file-based credentials, breaking under deny-by-default collectors

The chart's built-in Kubernetes component ServiceMonitors authenticate with file-based credentials:
 
```yaml
bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
tlsConfig:
  caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
 
Both fields are rejected by consumers that block arbitrary filesystem access from ServiceMonitors - Prometheus objects with `spec.arbitraryFSAccessThroughSMs.deny: true`, and, since v1.19.0, Grafana Alloy's `prometheus.operator.servicemonitors` component by default. In both cases the whole ServiceMonitor is dropped, so all control-plane targets disappear and the `kubernetes-system-*` alerts start firing.
 
For four of the seven components - kubelet, kube-api-server, kube-controller-manager and kube-scheduler - the paths are hardcoded in the templates, so there is no way to change them without disabling the ServiceMonitor and re-creating it via `prometheus.additionalServiceMonitors`. core-dns, kube-etcd and kube-proxy do read `bearerTokenFile` from values (kube-etcd also `caFile`/`certFile`/`keyFile`), but blanking those only removes the field - it leaves the endpoint unauthenticated rather than switching it to a secret-based credential.

### What's your helm version?

version.BuildInfo{Version:"v4.2.4", GitCommit:"3900f434fd3ef2b84065dc04508df48f288dba00", GitTreeState:"clean", GoVersion:"go1.26.5", KubeClientVersion:"v1.36"}

Client Version: v1.37.0

### Which chart?

kube-prometheus-stack

### What's the chart version?

88.6.2

### What happened?

With `arbitraryFSAccessThroughSMs.deny: true`, the Prometheus Operator rejects the entire ServiceMonitor object on the first offending endpoint, so all control-plane targets are lost.
 
Grafana Alloy ≥ v1.19.0 behaves the same way in effect. It emits one warning per offending endpoint:
 
```
level=warn msg="serviceMonitor endpoint references an arbitrary file from Alloy's filesystem" component_path=xxx component_id=xxx kind=serviceMonitor namespace=prometheus name=prometheus-kubelet endpoint=2 field=bearerTokenFile mitigation="remove the file reference or set allow_arbitrary_file_access to true to opt out"
```
 
but `addServiceMonitor` then breaks out of the endpoint loop on the first generation error and returns without applying any of the endpoints it had already collected, so the whole ServiceMonitor is discarded too.
 
Affected ServiceMonitors, verified against 88.6.2 with default values:
 
| Component | `bearerTokenFile` | `caFile` | Source |
| --- | --- | --- | --- |
| kubelet | 3 | 3 | hardcoded in `kube-prometheus-stack.kubelet.authConfig` |
| kube-api-server | 1 | 1 | hardcoded |
| core-dns | 1 | 0 | values |
| kube-etcd | 1 | 0 | values |
| kube-controller-manager | 1 | 1 | hardcoded |
| kube-scheduler | 1 | 1 | hardcoded |
| kube-proxy | 1 | 0 | values |
 
kubelet accounts for three of each because `kubelet`, `cAdvisor` and `probes` all default to `true` and each pulls in the same auth helper (`resource` defaults to `false`). kube-controller-manager and kube-scheduler already emit `caFile` at defaults: both default to `https: null`, which the `insecureScrape` helper resolves to secure on Kubernetes ≥ 1.22 and ≥ 1.23 respectively. core-dns has no `tlsConfig` at all, kube-proxy defaults to `https: false`, and kube-etcd defaults to `scheme: http` with an empty `caFile` so those three trip the guard on `bearerTokenFile` alone unless etcd is switched to HTTPS with a CA path set.
 
(`kubeDns.serviceMonitor.bearerTokenFile` exists as well, but that component is disabled by default.)
 
Because Alloy v1.19.0 (2026-08-21) made deny-by-default the behaviour of `prometheus.operator.servicemonitors`, a user who scrapes this chart's ServiceMonitors via Alloy silently loses all control-plane targets on an Alloy minor upgrade, with no change on the chart side, unless they opt back out of the guard.

The credential source for these endpoints should be configurable, so secret-based auth can be used instead of file paths. For example, per-component:
 
```yaml
kubelet:
  serviceMonitor:
    authorization:
      type: Bearer
      credentials:
        name: prometheus-sa-token
        key: token
    tlsConfig:
      ca:
        configMap:
          name: kube-root-ca.crt
          key: ca.crt
```
 
When set, the template would emit `authorization` / `tlsConfig.ca` instead of `bearerTokenFile` / `tlsConfig.caFile`. Defaults would stay as they are today, so this is backwards compatible. These are also the replacements Alloy itself points at (`bearerTokenSecret or authorization`, `tlsConfig.ca`, `tlsConfig.cert`, `tlsConfig.keySecret`).
 
A global toggle (e.g. `kubernetesServiceMonitors.authorization`) applied to all built-in ServiceMonitors would also work and would avoid repeating the same block seven times.
 
Separately, `bearerTokenFile` on a ServiceMonitor endpoint is deprecated in the `monitoring.coreos.com/v1` API, the field is annotated `Deprecated: use authorization instead`, so these templates rely on a deprecated field regardless of the guard.

### How to reproduce it?

The file references can be confirmed from the rendered templates alone:
 
```
$ for t in kubelet kube-api-server core-dns kube-etcd \
         kube-controller-manager kube-scheduler kube-proxy; do
  printf '%-26s %s\n' "$t" \
    "$(helm template release-name prometheus-community/kube-prometheus-stack \
         --version 88.6.2 \
         --show-only "templates/exporters/$t/servicemonitor.yaml" 2>/dev/null \
       | grep -c bearerTokenFile)"
done
kubelet                    3
kube-api-server            1
core-dns                   1
kube-etcd                  1
kube-controller-manager    1
kube-scheduler             1
kube-proxy                 1
 
$ for t in kubelet kube-api-server core-dns kube-etcd \
         kube-controller-manager kube-scheduler kube-proxy; do
  printf '%-26s %s\n' "$t" \
    "$(helm template release-name prometheus-community/kube-prometheus-stack \
         --version 88.6.2 \
         --show-only "templates/exporters/$t/servicemonitor.yaml" 2>/dev/null \
       | grep -c caFile)"
done
kubelet                    3
kube-api-server            1
core-dns                   0
kube-etcd                  0
kube-controller-manager    1
kube-scheduler             1
kube-proxy                 0
```
 
To see the resulting target loss:
 
1. Install chart 88.6.2 with default values.
2. Have the ServiceMonitors scraped either by a Prometheus object with `spec.arbitraryFSAccessThroughSMs.deny: true`, or by Grafana Alloy ≥ v1.19.0 with `allow_arbitrary_file_access` unset.
3. Check the targets page: the control-plane targets are gone, and the `kubernetes-system-*` alerts fire. The Operator logs `skipping servicemonitor`; Alloy logs the per-endpoint warning above followed by `error generating scrapeconfig from serviceMonitor`.

NONE

helm template release-name prometheus-community/kube-prometheus-stack --version 88.6.2

This also matters outside of self-managed Prometheus instances: OpenShift's `openshift-user-workload-monitoring` Prometheus has `deny: true` hardcoded in the cluster-monitoring-operator asset that manages it, so users cannot change it.
 
Alternatives considered:
 
- Setting `kubernetesServiceMonitors.enabled: false` and re-declaring all control-plane ServiceMonitors under `prometheus.additionalServiceMonitors`. This works, but means maintaining a copy of seven templates, including the relabelings the chart's shipped dashboards and rules depend on, outside the chart.
- Blanking `bearerTokenFile` for core-dns, kube-etcd and kube-proxy via values. Possible today, but it removes authentication rather than replacing it, and does nothing for the four components with hardcoded paths.
- Leaving `arbitraryFSAccessThroughSMs.deny` unset, or setting `allow_arbitrary_file_access = true` in Alloy. Not viable where the guard is required by policy, or where the Prometheus object is managed by the platform.

Versions and references:
 
- Chart version: reproduced on 88.6.2
- Prometheus Operator API: `ArbitraryFSAccessThroughSMsConfig`, added in prometheus-operator v0.34.0 (2019-10-31)
- API reference: https://prometheus-operator.dev/docs/api-reference/api/
- Grafana Alloy v1.19.0 breaking change (deny-by-default): https://github.com/grafana/alloy/releases/tag/v1.19.0

filippobuletto:

The built-in ServiceMonitors for kubelet, kube-apiserver, kube-controller-manager, kube-scheduler, kube-etcd, kube-proxy, coredns and kube-dns authenticate exclusively through `bearerTokenFile` and `tlsConfig.caFile`. Both fields declare raw filesystem paths, which are rejected outright by Prometheus Operator when
`spec.arbitraryFSAccessThroughSMs.deny` is true, and by Grafana Alloy's `prometheus.operator.servicemonitors` component since v1.19.0, where `allow_arbitrary_file_access` defaults to false. In either environment the whole ServiceMonitor is dropped and the control-plane scrape targets are lost.

Add optional per-component values that reference a Kubernetes Secret instead:

  - `authorization` in place of `bearerTokenFile`, for every component
  - `caSecret` in place of `caFile`, for every component that scrapes over TLS
  - `certSecret` and `keySecret` in place of `certFile` and `keyFile`, for kube-etcd, the only component that scrapes with client certificates

Setting a field switches that endpoint to the secret-based equivalent and drops the corresponding file-based field, since the two are mutually exclusive in the ServiceMonitor spec. All new values default to `{}`, so charts that do not set them render exactly as before. Note that clearing a deny-by-default environment requires setting both `authorization` and the TLS secrets; either one left unset remains a rejection trigger.

`kubeApiServer` keeps its CA override at `kubeApiServer.tlsConfig.caSecret` rather than under `serviceMonitor`, alongside the pre-existing `tlsConfig.serverName` and `tlsConfig.insecureSkipVerify`, so that its TLS settings stay in one place. `coreDns` and `kubeDns` only gain `authorization`, as those endpoints have no TLS support.

Three shared helpers render the file-or-secret choice so the eight templates do not each repeat it: `serviceMonitor.authorization`, `serviceMonitor.tlsSecret` (parameterised over `ca` and `cert`) and `serviceMonitor.tlsKeySecret`. The key needs its own helper because `keySecret` is a bare SecretKeySelector, whereas `ca` and `cert` are SecretOrConfigMap and nest the selector under `secret`. Each helper takes its indent width and emits a newline-prefixed block, so a call site that renders nothing contributes no stray whitespace.

Verified against Prometheus Operator on minikube with `arbitraryFSAccessThroughSMs.deny: true`: the secret-based kube-etcd monitor is accepted and its scrape job generated with operator-projected credentials, while the file-based equivalent is rejected with "it accesses file system via tls config which Prometheus specification prohibits".

- fixes #7230

#### Checklist

- [ ] Chart Version bumped
- [ ] Add an [helm unittest](https://github.com/helm-unittest/helm-unittest) test case to cover this change.

### DIFF
--- a/charts/kube-prometheus-stack/templates/prometheus/serviceaccount-token-secret.yaml
+++ b/charts/kube-prometheus-stack/templates/prometheus/serviceaccount-token-secret.yaml
@@ -1,0 +1,15 @@
+{{- if and .Values.prometheus.enabled .Values.prometheus.serviceAccount.create .Values.prometheus.serviceAccount.createTokenSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "kube-prometheus-stack.prometheus.tokenSecretName" . }}
+  namespace: {{ template "kube-prometheus-stack.namespace" . }}
+  labels:
+    app: {{ template "kube-prometheus-stack.name" . }}-prometheus
+    app.kubernetes.io/name: {{ template "kube-prometheus-stack.name" . }}-prometheus
+    app.kubernetes.io/component: prometheus
+{{ include "kube-prometheus-stack.labels" . | indent 4 }}
+  annotations:
+    kubernetes.io/service-account.name: {{ template "kube-prometheus-stack.prometheus.serviceAccountName" . }}
+type: kubernetes.io/service-account-token
+{{- end }}
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 89.3.0
+version: 90.0.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.93.1
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/Chart.yaml
+++ b/charts/kube-prometheus-stack/Chart.yaml
@@ -28,7 +28,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 89.2.2
+version: 89.3.0
 # renovate: github=prometheus-operator/prometheus-operator
 appVersion: v0.93.1
 kubeVersion: ">=1.25.0-0"
--- a/charts/kube-prometheus-stack/README.md
+++ b/charts/kube-prometheus-stack/README.md
@@ -132,11 +132,11 @@ Timeouts are only rendered for `kind: HTTPRoute`. They do not apply to generated
 
 This requires Gateway API CRDs that support HTTPRoute timeouts (available in the Standard channel since v1.2.0) and a Gateway controller that supports the corresponding timeout features. When both values are set, `backendRequest` must not exceed `request`, unless `request` is `"0s"`. See the [Gateway API timeout documentation](https://gateway-api.sigs.k8s.io/reference/api-types/httproute/#timeouts-optional) for duration semantics and controller support requirements.
 
-### Secret-based authentication for control-plane ServiceMonitors
+### Authentication of the control-plane ServiceMonitors
 
-The built-in ServiceMonitors for kubelet, kube-apiserver, kube-controller-manager, kube-scheduler, kube-etcd, kube-proxy, and coredns default to the deprecated `bearerTokenFile` / `tlsConfig.caFile` fields. Environments that set `prometheus.prometheusSpec.arbitraryFSAccessThroughSMs.deny: true`, or that scrape with Grafana Alloy's `prometheus.operator.servicemonitors` component (which defaults `allow_arbitrary_file_access` to `false`), reject these ServiceMonitors outright and drop the targets.
+The built-in ServiceMonitors for kubelet, kube-apiserver, kube-controller-manager, kube-scheduler, kube-etcd, kube-proxy, coredns and kube-dns authenticate through Kubernetes objects rather than through files on the scraper's filesystem. The bearer token comes from a Secret referenced by `authorization`, and the CA comes from the `kube-root-ca.crt` ConfigMap that Kubernetes maintains in every namespace. This keeps the ServiceMonitors valid when `prometheus.prometheusSpec.arbitraryFSAccessThroughSMs.deny` is `true`, and when they are scraped by Grafana Alloy's `prometheus.operator.servicemonitors` component, which rejects filesystem references by default since v1.19.0.
 
-Each component's `serviceMonitor` accepts an `authorization` field (in place of `bearerTokenFile`) and a `caSecret` field (in place of `caFile`, for components that support TLS) to reference a Kubernetes Secret instead:
+The chart creates the credential itself: `prometheus.serviceAccount.createTokenSecret` (enabled by default) renders a `kubernetes.io/service-account-token` Secret named `<prometheus service account>-token`, which every component references by default:
 
 ```yaml
 kubelet:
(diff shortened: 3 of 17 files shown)


jkroepke:
Hi, I appreciate it that some you add a toggle and leave some backwards compatibility.

However, reference local file paths can be considers as anti-pattern. And I read its deprecated?

Then, I would use

```
kubelet:
  serviceMonitor:
    authorization:
      type: Bearer
      credentials:
        name: prometheus-sa-token
        key: token
    tlsConfig:
      ca:
        configMap:
          name: kube-root-ca.crt
          key: ca.crt
```

by default and fully drop the local file access. This should reduce the complexity inside the helm chart, may omit the helpers as well. Just remove the hard coded values and put every configuration into the values.yaml should be sufficient enough? 

Only etcd. needs some special handling.

filippobuletto:
Thanks @jkroepke. Makes sense to me, dropping the file-based paths and defaulting to secret/configMap auth removes the helper branching entirely.

Before I rework it:

1. This changes default behavior for existing installs: OK as a breaking change, or should we keep a deprecated fallback for one release?
2. Defaulting CA to `kube-root-ca.crt`: are we assuming that's reliably present across supported cluster setups?
3. etcd stays secret-only for client certs (`certSecret`/`keySecret`), no file fallback there either: matches what you had in mind?

If that's right I'll drop the helpers and hardcode secret-based config as default, with etcd as the exception.

jkroepke:
OK as a breaking change. It lowers the maintaince cost and the old way may vanish anyways.

But please bump the major version. And add a upgrading note.

And it may make sense to do this change on the dependencies on a separate PR first, node-exporter, kube-state-metrics, windows-exporter, grafana. I can assist in review here, too. (If needed, not sure)

> Defaulting CA to kube-root-ca.crt: are we assuming that's reliably present across supported cluster setups?

Yes. It's the documented way to trust TLS within a cluster :https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#trusting-tls-in-a-cluster . Exists since 1.21. It must be exists in each namespace, because its manatory for the sa token mount.

> etcd stays secret-only for client certs (certSecret/keySecret), no file fallback there either: matches what you had in mind?

Currently, no tls crts are defined by default and empty. I would do the bearer token change, too. + Give users the ability to configure keySecret via values.yaml, e.g. just expose more properties to values.yaml - no overenginered helpers. please.

charts/kube-prometheus-stack/templates/prometheus/serviceaccount-token-secret.yaml
jkroepke, line 1: Not happy to see this, but I do not see any alternatives right now. I created a issue: https://github.com/prometheus-operator/prometheus-operator/issues/8798 to discuss this upstream.

jkroepke (dismissed):
Tested locally.

jkroepke:
Thanks for your contribution!

filippobuletto:
Thanks to you! And thanks to your kindness and availability!

jkroepke (approved):

PR #7264, opened 2026-09-12
[prometheus-node-exporter] keep the chart version out of the pod labels

DrFaust92:

`prometheus-node-exporter.labels` includes `helm.sh/chart`, and the DaemonSet applies it to the pod template as well as to the object. That value changes on every chart version bump, so every node-exporter pod on every node is recreated on `helm upgrade` even when nothing about the workload changed — image, args and config identical.

This adds a `prometheus-node-exporter.podLabels` helper, the same label set without `helm.sh/chart`, and uses it in the pod template. The DaemonSet object keeps the full set, so the chart version is still visible where it belongs, and pods keep name, instance, component, part-of, managed-by and `app.kubernetes.io/version`.

Rendered 4.56.4 against a bumped version: the pod template is byte-identical, and the only differences are the object-level labels. All the values files under `ci/` still render and `helm lint` passes.

This is the first `unittests/` suite for this chart — three cases: the chart label is absent from the pod template but present on the object, the pod labels are unchanged under an overridden chart version, and user-supplied `podLabels` still render.

The same change was recently made in the argo-cd chart (argoproj/argo-helm#4044 covers the related checksum case) and is in flight for kargo (akuity/kargo#7192).

This PR was prepared with AI assistance (Claude), reviewed and tested by me before opening.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/prometheus-node-exporter/Chart.yaml
+++ b/charts/prometheus-node-exporter/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.56.3
+version: 4.56.4
 # renovate: github=prometheus/node_exporter
 appVersion: 1.12.1
 home: https://github.com/prometheus/node_exporter/
--- a/charts/prometheus-node-exporter/templates/_helpers.tpl
+++ b/charts/prometheus-node-exporter/templates/_helpers.tpl
@@ -54,6 +54,21 @@ release: {{ .Release.Name }}
 {{/*
 Selector labels
 */}}
+{{/*
+Pod labels: the common labels without helm.sh/chart. That label carries the chart version, so
+including it in the pod template restarts the DaemonSet on every chart upgrade, on every node,
+even when nothing about the workload changed.
+*/}}
+{{- define "prometheus-node-exporter.podLabels" -}}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: metrics
+app.kubernetes.io/part-of: {{ include "prometheus-node-exporter.name" . }}
+{{ include "prometheus-node-exporter.selectorLabels" . }}
+{{- with .Chart.AppVersion }}
+app.kubernetes.io/version: {{ . | quote }}
+{{- end }}
+{{- end }}
+
 {{- define "prometheus-node-exporter.selectorLabels" -}}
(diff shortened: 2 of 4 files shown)


charts/prometheus-node-exporter/templates/_helpers.tpl
```diff
@@ -54,6 +54,21 @@ release: {{ .Release.Name }}
 {{/*
 Selector labels
 */}}
+{{/*
+Pod labels: the common labels without helm.sh/chart. That label carries the chart version, so
```
zeritti, line 58: We can avoid repeating the labels in the new helper by leaving just the affected label out. What do you think?

```yaml
{{- define "prometheus-node-exporter.podLabels" }}
{{- $podLabels := (include "prometheus-node-exporter.labels" . | fromYaml) }}
{{- omit $podLabels "helm.sh/chart" | toYaml }}
{{- end }}
```

zeritti (commented):
Thank you, @DrFaust92, for your PR. Please, see my comment below.

charts/prometheus-node-exporter/templates/_helpers.tpl
DrFaust92: Applied, thanks — much better than re-listing them.

One thing I checked before switching: the fromYaml/toYaml round trip drops quotes, so I tested a numeric-looking appVersion (2.0) and the label still renders as `"2"` — goyaml quotes anything that would otherwise read as a scalar, so no risk of a non-string label value.

charts/prometheus-node-exporter/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
   - prometheus
   - exporter
 type: application
-version: 4.56.3
+version: 4.56.4
```
zeritti: Apologies for not having raised it earlier. I think this change should be minor.

```suggestion
version: 4.57.0
```

zeritti (changes requested):

charts/prometheus-node-exporter/Chart.yaml
DrFaust92: Done, 4.57.0.

zeritti (approved):
Thank you, @DrFaust92, LGTM.

PR #7266, opened 2026-09-12
[kube-state-metrics] keep the chart version out of the pod labels

DrFaust92:

`kube-state-metrics.labels` includes `helm.sh/chart`, and the Deployment applies it to the pod template as well as to the object. That value changes on every chart version bump, so the pod is recreated on `helm upgrade` even when nothing about the workload changed — image, args and config identical. It also rolls whenever kube-prometheus-stack bumps its dependency on this chart.

This adds a `kube-state-metrics.podLabels` helper that derives the pod labels from the common labels and omits the one key that churns, following the same approach agreed in #7264. The Deployment object keeps the full set, and pods keep name, instance, component, part-of, managed-by and `app.kubernetes.io/version`.

Rendered against a bumped chart version: the pod template is unchanged, the only differences are the object-level labels. All `ci/` values files render and `helm lint` passes. Minor version bump, since rendered pod metadata changes.

The call site moved from `indent 8` to `nindent 8` because the new helper does not emit a leading newline the way the `labels` helper does.

New `unittests/pod_labels_test.yaml` covers three cases: the chart label is absent from the pod template but present on the object, the pod labels are unchanged under an overridden chart version, and user supplied `podLabels` still render. Full suite is 42 tests passing.

This PR was prepared with AI assistance (Claude), reviewed and tested by me before opening.

#### Checklist

- [x] Chart Version bumped


sebastiangaiser (approved):
Thank you 🚀

Issue #7246, opened 2026-09-07
[prometheus] automountServiceAccountToken: false on kube-state-metrics fatally crashes the pod


Md-Arif-Hasan:
### Describe the bug a clear and concise description of what the bug is.

Setting `kube-state-metrics.automountServiceAccountToken: false`, a user-configurable Helm value, causes the kube-state-metrics pod to enter CrashLoopBackOff because the service account token is unavailable.

### What's your helm version?

version.BuildInfo{Version:"v3.21.0+ge0878d4", ...}

### What's your kubectl version?

version.BuildInfo{Version:"v3.21.0+ge0878d4", ...}

### Which chart?

prometheus

### What's the chart version?

29.27.0

### What happened?

Setting `kube-state-metrics.automountServiceAccountToken` to `false` causes the kube-state-metrics pod to enter `CrashLoopBackOff`.

The pod cannot create its in-cluster Kubernetes client because the service account token is unavailable:

```text
error creating inClusterConfig, falling back to default config: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory

### What you expected to happen?


`automountServiceAccountToken` is exposed as a user-configurable option in `charts/kube-state-metrics/values.yaml:61` and is applied to the pod specification in `templates/deployment.yaml:45`.

Setting this supported configuration to `false` should not cause kube-state-metrics to crash because the service account token is unavailable.

### How to reproduce it?

1. Install the Prometheus chart:

   ```bash
   helm upgrade --install prometheus prometheus-community/prometheus \
     --version 29.27.0 \
     --namespace prometheus-test \
     --create-namespace \
     --wait

2. Set automountServiceAccountToken to false on the kube-state-metrics Deployment:

kubectl -n prometheus-test patch deployment prometheus-kube-state-metrics \
  --type=merge \
  -p '{"spec":{"template":{"spec":{"automountServiceAccountToken":false}}}}'

3. Check the pod:
   kubectl -n prometheus-test get pods

4. Check the logs:
kubectl -n prometheus-test logs <pod>

The logs show that the service account token is unavailable and kube-state-metrics fails to start.







### Enter the changed values of values.yaml?

**Changed values of values.yaml**

```yaml
kube-state-metrics:
  automountServiceAccountToken: false

### Enter the command that you execute and failing/misfunctioning.

helm upgrade --install prometheus prometheus-community/prometheus \
  --version 29.27.0 \
  --namespace prometheus-test \
  --create-namespace \
[...]

PR #7274, opened 2026-09-13
[kube-state-metrics] Expose opt-in metrics, gzip and unscheduled pod tracking

DrFaust92:

Three kube-state-metrics flags that were only reachable through `extraArgs` become first-class values:

- `metricOptInList` → `--metric-opt-in-list`, next to the existing allow/deny lists. The chart lets you pick opt-in *collectors* but had no way to enable opt-in *metrics*.
- `enableGZIPEncoding` → `--enable-gzip-encoding`. It was the example in the `extraArgs` comment, which is a fair sign it gets used.
- `trackUnscheduledPods` → `--track-unscheduled-pods`.

All default off, so rendered args are unchanged unless set.

Also adds a note next to `automountServiceAccountToken` that turning it off means supplying credentials another way (`kubeconfig.enabled`, or a projected token volume) — that combination crashes the pod and has come up as a bug report (#7246).

Collides with #7266 on `Chart.yaml` only — both take 8.5.0, so whichever lands second needs a rebase. I'm an AI agent (Claude) working with @DrFaust92.

#### Checklist

- [x] Chart Version bumped


charts/kube-state-metrics/values.yaml
```diff
@@ -53,11 +53,16 @@ replicas: 1
...
 
 # List of additional cli arguments to configure kube-state-metrics
-# for example: --enable-gzip-encoding, --log-file, etc.
+# for example: --log-file, --auto-gomemlimit, etc.
 # all the possible args can be found here: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/developer/cli-arguments.md
 extraArgs: []
 
-# If false then the user will opt out of automounting API credentials.
+# Gzip responses to scrapes that accept the encoding.
+enableGZIPEncoding: false
```
sebastiangaiser, line 61:
```suggestion
enableGzipEncoding: false
```

charts/kube-state-metrics/templates/deployment.yaml
```diff
@@ -148,6 +151,12 @@ spec:
         {{- if .Values.authFilter.enabled }}
         - --auth-filter
         {{- end }}
+        {{- if .Values.enableGZIPEncoding }}
```
sebastiangaiser, line 154:
```suggestion
        {{- if .Values.enableGzipEncoding }}
```

charts/kube-state-metrics/unittests/args_test.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - kube_secret_annotations
+        - kube_node_annotations
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --metric-opt-in-list=kube_secret_annotations,kube_node_annotations
+
+  - it: should render gzip encoding and unscheduled pod tracking
+    set:
+      enableGZIPEncoding: true
```
sebastiangaiser, line 30:
```suggestion
      enableGzipEncoding: true
```

sebastiangaiser (changes requested):
Thanks, this is a useful addition, and the flags map correctly to kube-state-metrics v2.20.0 with defaults matching KSM's own (all opt-in). One naming change before merge though: `enableGZIPEncoding` should be `enableGzipEncoding`.

Helm's values best practices state that variable names ["should begin with a lowercase letter, and words should be separated with camelcase"](https://helm.sh/docs/chart_best_practices/values/#naming-conventions), i.e. an acronym is camelCased (`Gzip`) rather than kept as an all-caps run (`GZIP`). This chart already follows that convention: it uses `enableHttp2` (not `enableHTTP2`) in the `prometheus.monitor` blocks.

Since a values key becomes public API on release and can only be changed with a breaking rename afterwards, could you please rename it now? It appears in three spots:
- `values.yaml` (the key + its comment)
- `templates/deployment.yaml` (`{{- if .Values.enableGZIPEncoding }}`)
- `unittests/args_test.yaml`

`metricOptInList` and `trackUnscheduledPods` are fine as they are.

